PR #430, opened 2014-01-24
This update allows you to use auth_smtp with smtp_proxy.

boky8:
It basically provides a transparent SMTP proxy service to the back-end system. You can
then use one SMTP server (Haraka) to forward your emails to different
back-end servers based on your username. Especially useful if you are
using one IP-address and for example, two DavMail servers. This change
directly complements what Perdition (Mail retrieval proxy) does for POP
and IMAP.

### DIFF
--- a/plugins/auth/auth_proxy.js
+++ b/plugins/auth/auth_proxy.js
@@ -51,8 +51,12 @@ exports.try_auth_proxy = function (connection, hosts, user, passwd, cb) {
     var response = [];
 
     var hostport = host.split(/:/)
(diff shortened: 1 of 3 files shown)


dscho:
@smfreegard FWIW I tried to address a few issues you pointed out, but now the tests fail because `connection` is not known in `plugins/queue/smtp_forward.js`. Unfortunately, I ran out of time, but in case somebody wants to pick this up again: https://github.com/dscho/Haraka/compare/auth-proxy

duanefields:
Just the fixes to allow the smtp_forward to support authentication are available on my branch, https://github.com/glg/Haraka

PR #452, opened 2014-02-10
DMARC: working via Mail::DMARC

msimerson:
### Do Not Merge

A native DMARC validator is being written by @smfreegard 

Until then, this one works by communicating with the Mail::DMARC perl plugin via HTTP.  DKIM validation is via the `dkim_verify` plugin and also by checking the SpamAssassin results.


msimerson:
I intend to repackage this as `haraka-plugin-dmarc-perl`, making it fast and easy to load externally.

PR #738, opened 2014-11-12
geoip: deprecate geoip-lite for maxmind

msimerson:
- adds support for maxmind module
  - ASN lookups
  - IPv6 support
  - compiles/installs faster
  - runs faster
- maintains compatibility with geoip-lite
- removes geoip-lite from package.json, speeding Haraka installs and Travis tests
- doesn't register hooks unless module require works & MaxMind DBs are present
- loads connect.geoip.ini in a new function with a callback
- adds script for downloading GeoIP dbs


msimerson:
Yeesh, speaking of PRs that should be squashed....

PR #767, opened 2014-12-03
prevent replace \r\n

icepigeon:
If we pass a content which already has \r\n on every line – libs like Mailcomposer (https://github.com/andris9/mailcomposer) already set it – the email will end up in an invalid format.


msimerson:
Closing, this issue was addressed in #817. Please try that, and if it doesn't resolve the issue for you, reopen this PR.

PR #785, opened 2014-12-18
cfreader/[ini, json, yaml, binary, list] with tests!

msimerson:
The idea of this is to extract file type readers into their own libraries with test coverage for each. It's ready for comments only.


baudehlo:
Is this PR now out of date?

msimerson:
yes, it's out of date, but I intend to return to it. There's only 7 more Haraka items on my todo list before this one.

msimerson:
Closing. I'll re-open a new PR when I get around to revisiting this.

PR #811, opened 2015-01-12
Fix Txn.attachment_hooks to defer body creation like similar methods

chazomaticus:
Previously, if you called Transaction.attachment_hooks before Transaction.set_banner or .add_body_filter, your banner or filter wouldn't be applied because the body had already been created inside attachment_hooks.  This delays creating the body until it's actually going to be used, so all of these methods can be called in any order and still take effect.

I also snuck in a test for my last pull request (#739), which added Transaction.add_body_filter but no test on it.


msimerson:
:+1:

chazomaticus:
Thanks y'all!

PR #813, opened 2015-01-23
programmatically stop server

roosit:
I developed a project [SMTPeshka](https://github.com/abricos/smtpeshka) debugging messages in other programs. To pass the automatic test must programmatically stop Haraka server.

### DIFF
--- a/server.js
+++ b/server.js
@@ -201,6 +201,8 @@ Server.get_smtp_server = function (host, port, inactivity_timeout) {
     return server;
 };
 
+Server.smtp_server_listeners = [];
+
 Server.setup_smtp_listeners = function (plugins, type, inactivity_timeout) {
 
     var listeners = Server.get_listen_addrs(Server.cfg.main);
@@ -248,6 +250,12 @@ Server.setup_smtp_listeners = function (plugins, type, inactivity_timeout) {
         });
 
         server.listen(port, host);
+
+        Server.smtp_server_listeners.push({
+            host: host,
+            port: port,
+            server: server
+        });
     };
 
     async.each(listeners, setupListener, runInitHooks);
@@ -306,4 +314,21 @@ Server.listening = function () {
     }
 
     Server.ready = 1;
+};
+
+Server.close = function () {
+    var list = Server.smtp_server_listeners;
+    for (var i=0; i < list.length; i++) {
+        list[i].server.close();
+    }
+
+    if (!Server.cluster) {
+        out.stop_queue();
+        return;
+    }
+
+    for (var id in cluster.workers) {
+        cluster.workers[id].send({event: 'outbound.stop_queue'});
+    }
+    Server.smtp_server_listeners = [];
 };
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -307,3 +315,20 @@ Server.listening = function () {
...
+        list[i].server.close();
+    }
+
+    if (!Server.cluster) {
+        out.stop_queue();
+        return;
+    }
+
+    for (var id in cluster.workers) {
+        cluster.workers[id].send({event: 'outbound.stop_queue'});
```
baudehlo: In order for this to work you need to add it here too:

https://github.com/roosit/Haraka/blob/smtpeshka/outbound.js#L128

baudehlo:
Also note - to stop the outbound queue you'd be better off just doing: "echo 1 > config/outbound.disabled". That way it can be re-enabled by removing that file.

Also I'm not sure what the point is of this patch - can you elaborate on why you need to be able to stop Haraka, but have no facility to restart it again?

roosit:
Initially, I start and stop the Haraka server through child_process.spawn. But I would like to use a closer integration with the server in the future.
I wrote a [plugin for Haraka](https://github.com/abricos/smtpeshka/blob/master/haraka-def/plugins/json.js), which does not send messages and stores them in a JSON File. This file is then used in order to be able to obtain information about the e-mail to a third party application.
[SMTPeshka](https://github.com/abricos/smtpeshka) works as follows:
1. launches web-server to view e-mail messages sent
2.  launches Haraka server to receive e-mail messages sent and save them to the database (currently JSON files)
3. The client program sends a message and Haraka server saves it in a JSON file.
The user can view your messages on the Web site at http://localhost:2580. You can also use the REST API to get sent e-mail messages
[](http://abricos.org/filemanager/i/741049da/home.png)
[](http://abricos.org/filemanager/i/46e3351b/email.png)

How do I use SMTPeshka?
I am developing a Abricos Platform (CMS). When I run the test REST API functions, mocha launches SMTPeshka. 
During the test, Abricos sends emails and test checks them.
If you do not closing Haraka-server programmatically, the mocha fails timeout

In the future, I plan to develop the ability to send e-mails stored in a database (Mysql, PostgreSQL and etc.)
SMTPeshka should be a powerful tool for debugging emails

At the moment, stop the server Haraka need to test finished. Restart it is not necessary.
Thanks

baudehlo:
Sounds like terminating Haraka with child.kill() would be MUCH more appropriate for the use case you have, no?

BTW your project sounds a lot like mailcatcher, and I'm glad to see a Node version of it springing up :)

roosit:
if you use Haraka as a child process, it loses all meaning in the philosophy of Node.js :)
With direct access to the functions Haraka appears more opportunities for the future.

baudehlo:
We had a very old issue for Haraka about using it as a library (can't find it now) which I closed because it really didn't seem sensible. I'm going to close this ticket as running as a child process is the right way to handle this.

PR #815, opened 2015-01-28
Prevent duplication of multi-part preamble text

chazomaticus:
Currently, if you add Body filters (e.g. with Transaction.add_body_filter or indirectly with .set_banner), messages that contain a MIME multi-part preamble (often "This is a multi-part message in MIME format.") will be enqueued with that preamble repeated at the very end of the message, after the terminating boundary line.  This adds a test to catch the error and fixes it by not capturing the body text for multi-part containers.  It doesn't seem useful to capture that text anyway--it's not a text part for one, and we already don't capture the "text" of attachments.

The reason this fixes the issue is that when Body.parse_end is called, the body text for each body part is added to the returned "line" before filters are applied.  With no body text for containers, there's nothing to erroneously add to the returned line.  In (I think) every other case, we don't emit the body text until after filters are applied, so there's no duplication.

As far as I can tell, this duplication might actually be harmless (I learned from reading RFC 2046 that preambles *and* "epilogues" are valid), but Haraka shouldn't be modifying messages that way when it enqueues them.

### DIFF
--- a/tests/mailbody.js
+++ b/tests/mailbody.js
@@ -69,6 +69,7 @@ exports.filters = {
         test.ok(/This is some HTML/.test(parts[1]));
         test.done();
     },
+
     'search/replace': function (test) {
         test.expect(2);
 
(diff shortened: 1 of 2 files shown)


chazomaticus:
Hrm, just realized these commits don't have my full name in the author field.  Let's find out what happens when I amend them!

tests/mailbody.js
```diff
@@ -87,4 +88,44 @@ exports.filters = {
...
+        lines = lines.filter(function (l) {
+            return l.trim();
+        });
+
+        var dupe = false;
+        var line;
+        while (line = lines.pop()) {
+            lines.forEach(function (l) {
+                dupe = dupe || line === l;
+            });
```
msimerson:
```js
dupe = dupe || line === l;
```
That's very odd looking, and I'm not sure what that's supposed to be doing. Care to enlighten me?

chazomaticus:
It flips dupe to true if a line is ever repeated from body.parse_*, but doesn't let it flip back to false.  I suppose that could be shortened to use lines.some() if it's bugging you.

msimerson:
Ah, I see. The `dupe = dupe` yields a false (the product of the set operation, vs the 'success' of having set it. I learned something new about JS today, thanks.

chazomaticus:
:+1:

msimerson:
LGTM

chazomaticus:
Thanks!

PR #823, opened 2015-02-03
plugins: improved timeout handling

msimerson:
* adds test coverage to timeout handling
* disables config cache when WITHOUT_CONFIG_CACHE is set so it's possible to test that the timeout is read properly when config files are and aren't present)

### DIFF
--- a/plugins.js
+++ b/plugins.js
@@ -15,13 +15,7 @@ var states      = require('./connection').states;
 function Plugin(name) {
     this.name = name;
     this.base = {};
-    this.timeout = config.get(name + '.timeout');
-    if (this.timeout === null) {
-        this.timeout = config.get('plugin_timeout') || 30;
-    }
-    else {
-        logger.logdebug("plugin " + name + " set timeout to: " + this.timeout + "s");
-    }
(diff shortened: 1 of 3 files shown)


plugins.js
```diff
@@ -12,33 +12,34 @@ var util        = require('util');
...
     this.name = name;
     this.base = {};
-    this.timeout = config.get(name + '.timeout');
-    if (this.timeout === null) {
-        this.timeout = config.get('plugin_timeout') || 30;
-    }
-    else {
-        logger.logdebug("plugin " + name + " set timeout to: " + this.timeout + "s");
+    this.timeout = parseFloat((config.get(name + '.timeout')));
+    if (!this.timeout) {
```
baudehlo: Doesn't this prevent a zero timeout (i.e. infinite) working?

msimerson: Hmmm, maybe. IIRC, I put that there after ending up with a timeout that was NaN. So maybe this would be better?
```js
if (this.timeout === NaN)
```

msimerson:
New and improved PR with only substantive change introducing test coverage to plugin.timeout.

PR #827, opened 2015-02-06
fix adding timestamps to log messages

damoxc:
The current method of adding a log timestamp results in the startup time of
Haraka being added to every log message, which isn't ideal. Use a proxy function
which uses a new Date for every log message written to the console.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -147,7 +147,13 @@ logger._init_timestamps = function () {
     });
     
     if (_timestamps) {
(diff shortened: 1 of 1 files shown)


msimerson:
I have confirmed that if one enables timestamps by:
````
echo 1 > config/log_timestamps
````
Then log messages for the rest of time are indeed prefixed with the daemons startup time, just as @damoxc describes.

PR #831, opened 2015-02-08
connection: shorten lines

msimerson:


connection.js
```diff
@@ -169,8 +172,10 @@ function Connection(client, server) {
     this.transaction = null;
     this.tran_count = 0;
     this.capabilities = null;
-    this.early_talker_delay = config.get('early_talker.pause') || config.get('early_talker_delay') || 1000;
-    this.banner_includes_uuid = config.get('banner_includes_uuid') ? true : false;
+    this.early_talker_delay = config.get('early_talker.pause') ||
+                              config.get('early_talker_delay') || 1000;
+    this.banner_includes_uuid =
+        config.get('banner_includes_uuid') ? true : false;
```
baudehlo: Side conversation: Do we have any way of telling flat config files that they are boolean options like we do with ini files? Would be nice to support the same 1/0/true/false/yes/no syntax.

msimerson: I believe what you're asking for is implemented [right here](https://github.com/baudehlo/Haraka/blob/master/configfile.js#L217).

baudehlo: Sweet. So I guess we need to specify that these things are boolean in the
get() call.

PR #832, opened 2015-02-08
configfile: shed some excess indention

msimerson:
* replace cfreader.config_path with cp (5x)
* early exit of functions permit removing of extra indention

### DIFF
--- a/configfile.js
+++ b/configfile.js
@@ -59,28 +59,26 @@ cfreader.on_watch_event = function (name, type, options, cb) {
 
 cfreader.watch_dir = function () {
     // NOTE: This only works on Linux and Windows
-    if (cfreader._watchers[cfreader.config_path]) return;
+    var cp = cfreader.config_path;
+    if (cfreader._watchers[cp]) return;
+    var watcher = function (fse, filename) {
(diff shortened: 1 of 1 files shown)


baudehlo:
That's a rather large diff for a straight pull-merge while we're supposed to be in code freeze... Do we have enough test coverage of that area of code to know it's working the same in all conditions?

msimerson:
The diff size is primarily the result of the "left shift" of those two blocks by 4 spaces. I wonder if, in cases like this, if wouldn't be better (for the purposes of review at least) to make the 2-3 line changes that permit the left shift in one PR, and then follow it with the left shifted block in another. I'm betting that github's whitespace highlighting would then work and it wouldn't **appear** be a substantive change.

msimerson:
Actually, using two PRs wouldn't be necessary. Two commits in one PR, with a note explaining that the code change is in the first, and the second is **only** a left shift that happens to balloon the diff in rude proportion ought to work...

baudehlo:
Yeah I only commented because there's a bit of a logic change (or reversal)
to make the indent change, and I was too tired to be able to take in the
entire thing.

PR #833, opened 2015-02-10
utils: added .extend(), copy props onto objects

msimerson:
* and wrap some long lines


baudehlo:
Can this wait until 2.6? I see no reason to review/merge it yet.

msimerson:
Sure, it's a function I'll be using in my `watch` plugin to simplify that code a bit. (Use case is merging the transaction.results with connection.results).

PR #834, opened 2015-02-12
populate config/plugins with disabled plugins

msimerson:
* makes it easier to enable features
* provides information about available plugins
* makes it likely that they'll be enabled in the right order
* (en | dis)able plugins with sed

### DIFF
--- a/config/plugins
+++ b/config/plugins
@@ -1,33 +1,58 @@
 # default list of plugins
 
+#process_title
 # Log to syslog (disabled by default, see docs)
 # log.syslog
 
+# CONNECT
+#toobusy
+#relay
 # control which IPs, rDNS hostnames, HELO hostnames, MAIL FROM addresses, and
 # RCPT TO address you accept mail from. See 'haraka -h access'.
 access
-
+# connect.p0f
+# connect.geoip
+# connect.asn
+# connect.fcrdns
 # block mails from known bad hosts (see config/dnsbl.zones for the DNS zones queried)
 dnsbl
 
-# Check mail headers are valid
-data.headers
-
-# block mail from some known bad HELOs - see config/helo.checks.ini for configuration
+HELO
+#early_talker
+# see config/helo.checks.ini for configuration
 helo.checks
-
 # see 'haraka -h tls' for config instructions before enabling!
 # tls
+# auth: auth_base   auth_ldap   auth_proxy  flat_file
 
+# MAIL FROM
 # Only accept mail where the MAIL FROM domain is resolvable to an MX record
 mail_from.is_resolvable
+#spf
+
+# RCPT TO
+# NOTE: A rcpt_to plugin is REQUIRED for inbound email.
+# Edit config/host_list to set what domains you accept mail for.
+rcpt_to.in_host_list
+#rcpt_to.qmail_deliverable
+#rcpt_to.ldap
+#rcpt_to.routes
+
+# DATA
+#bounce
+# Check mail headers are valid
+data.headers
+#data.uribl
+#attachment
+#clamd
+#spamassassin
+#dkim_sign
+#karma
+
+# QUEUE
+# queues: discard  qmail-queue  quarantine  smtp_forward  smtp_proxy
+# Queue mail via smtp - see config/smtp_forward.ini for where your mail goes
+queue/smtp_forward
 
 # Disconnect client if they spew bad SMTP commands at us
 max_unrecognized_commands
-
-# Only accept mail for your personal list of hosts. Edit config/host_list
-# NOTE: THIS IS REQUIRED for inbound email.
-rcpt_to.in_host_list
-
-# Queue mail via smtp - see config/smtp_forward.ini for where your mail goes
-queue/smtp_forward


msimerson:
Any objections to merging this?

baudehlo:
Just seems kinda messy to me. I don't really see the point.

msimerson:
There's been at least a half dozen times in the past few months where we had to explain to users that they need to add the plugin name into the config/plugins files to enable it. There's also been a few where they added it in config/plugins incorrectly. There's been a couple where they added in the wrong place and so it wasn't working for them.

This change isn't for Haraka developers, it's for Haraka users.

baudehlo:
I think you make a good point about the ordering of plugins, so +1 from me.

smfreegard:
My only concern is that pre-populating config/plugins doesn't actually teach newcomers what they need to know in order to understand how it actually works:

e.g. hooks run in SMTP state order, then each plugin that hooks that state in the order that they are listed on config/plugins  e.g. lookup_rdns, connect, ehlo/helo, esmtp:starttls, esmtp:auth, mail, rcpt, rcpt_ok, data, [attachment], data_post, queue or queue_outbound, queue_ok, [next transaction], disconnect

And in the case of outbound:

queue_outbound, [deliver or defer (multiple) or bounce], [bounce or deliver]

And then we have the special cases of deny, unrecognized_command etc.

What we really need to do is document this process better and put some pretty graphics in showing it and formalize the plugin docs a bit (e.g. list any dependencies, list which hooks are used etc.) as some plugins will rely on the results of others, so must run in a specific order (which we can't currently enforce or describe in code yet) and others are more of an ideological choice (e.g. data.uribl before clamd/spamassassin) and things like dkim_verify (which is missing from this BTW) need to be run early (as other data_post plugins might modify headers) and dkim_sign needs to be run late so it signs the final version of the message before it is queued etc.

baudehlo:
Yeah it's a good point about some things being preferences of installation.
If we specify the entire file it's like it should never change... Urgh I'm
really in two minds about this one. Should we unmerge it so we can think
about it a bit more?

msimerson:
> If we specify the entire file it's like it should never change

How do you arrive at that conclusion?

We don't specify **every** plugin, but an assortment of commonly used ones. As much as anything, we're providing a **template** for how to include additional plugins. We're also enhancing plugin discoverability by listing them there. The contents of the plugins dir and its structure is currently *not* obvious to newcomers, nor is figuring out the plugin names so they can run 'haraka -h $plugin_name'.

The other thing added was the section headings that outline the SMTP state order.  They aren't entirely obvious unless you regularly test SMTP with netcat, but they do provide a helpful scaffolding.

I'm all for better documentation and pretty graphics, but until that glorious day arrives, incremental improvements should be welcome. I believe that this change will prove helpful to Haraka newcomers and trying this is the only way to proof the theory.

> dkim_sign needs to be run late so it signs the final version of the message

We recently added a new hook to Qpsmtpd, (data_post_headers) which is called by plugins that insert new headers or alter existing headers like Authentication-Results. data_post_headers is called immediately before data_post, so that plugins that run on data_post have access to headers like Received-SPF and Authentication-Results. That remains true regardless of plugin ordering in config/plugins. It was done because SpamAssassin benefits from having those headers available and A-R isn't "complete" until after DKIM and DMARC have run.

It's imperfect because SA, dspam, and DKIM all have to run on or after data_post and they too add their own headers to the message. It's almost like we need something like [rcorder](https://github.com/kil/rcorder) that allows each plugin to declare what it provides and what it depends on and can thus be sorted deterministically.

baudehlo:
On Mon, Feb 23, 2015 at 3:01 PM, Matt Simerson <notifications@github.com>
wrote:

> If we specify the entire file it's like it should never change
>
> How do you arrive at that conclusion?
>
> I meant "it's like you should never change the order". Ordering is
critical to how you want the system to behave. For example I want HELO
checks before DNSBLs, but others might not.

msimerson:
> For example I want HELO checks before DNSBLs

Is that just a poor example? Per my understanding, you can list HELO plugins first in config/plugins but the DNSBL plugins will still "check" first because they run on hook `connect`. Right?

msimerson:
Perhaps another helpful inclusion to config/plugins is instructions for running:
```sh
haraka -o -c /path/to/haraka/config
```
As well as, "plugin ordering matters, run $CMD to see how your changes will run," or something like that.

baudehlo:
Sorry yes probably a bad example (although I do wish my HELO checks ran
before DNSBL).

baudehlo:
Yeah +1 to both things. Though I guess nobody reads comments anyway. *sigh*

smfreegard:
@baudehlo - easy to do; hook_deny is great for that as I use it for similar stuff here.  (see the delay_deny plugin).

@msimerson - I added the -o option and I thought I did add some information to docs/Plugins for it.   I'll add something shortly and see if I can improve it.

smfreegard:
If anything config/plugins should just point to the documentation.   We can update the documentation whereas config/plugins we can't as it's essentially 'fixed' once they're run 'haraka -i' for the first time.   So what we're essentially talking about here only affects new users from the next released version.

Personally I'd rather ship an essentially empty plugins file and no make any presumptions about how someone uses Haraka and then include some boilerplate config files in the docs/config folder for common use cases.

soundslocke:
Outsider/lurker perspective:

When I first began working with Haraka it was tricky figuring out the plugins config file and in what order things were/should be happening. Partly to do with little experience with SMTP of course. I think what @msimerson is suggesting would have helped. Better documentation sounds great as well. The more the better.

Sort of related: I really like the inline documentation that Redis contains with [its config file](https://github.com/antirez/redis/blob/unstable/redis.conf).

PR #838, opened 2015-02-13
add Windows build status badge

msimerson:
* add appveyor.yml
* remove ./ from run_tests
* make plugins test Windows compatible.
* permit build failures on node 0.8


msimerson:
There is currently a build failure in node 0.8 due to, you'll never believe this, an iconv dependency that fails. Other than that, this adds a couple small changes to the test suite to get it passing when running on Windows.

PR #847, opened 2015-02-16
Fix the default port to match the docs and default .ini

baudehlo:
@msimerson can you just check this PR out? It broke my upgrade because I didn't have the config file.


msimerson:
I'm in Canada on a ski trip. It couldn't look better.

baudehlo:
Lucky you! Thanks for the merge.

PR #852, opened 2015-02-20
Handle IPv6 in is_rfc1918 - fixes #850

smfreegard:

PR #854, opened 2015-02-22
rename is_rfc1918 -> is_private_ip, add IPv6 capability

### LINKED ISSUE #850: Error: Outbound email with "auth/flat_file"
=== Connected to localhost.
<-  220 li940-16 ESMTP Haraka 2.5.0 ready
 -> EHLO li940-16.members.linode.com
<-  250-li940-16 Hello ip6-localhost [::1], Haraka is at your service.
<-  250-PIPELINING
<-  250-8BITMIME
<-  250 SIZE 500000
*** Host did not advertise authentication
 -> QUIT
<-  221 li940-16 closing connection. Have a jolly good day.

Please any one help me on this...

Thanks,
Varma

msimerson:
this is @smfreegard 's PR #852, squashed, rebased, and with an extra commit that renames is_rfc1918 to ip_private_ip. 

Closes #852
Closes #850

### DIFF
--- a/net_utils.js
+++ b/net_utils.js
@@ -191,7 +191,7 @@ exports.is_ip_in_str = function(ip, str) {
     return false;
 };
 
-exports.is_rfc1918 = function (ip) {
+exports.is_private_ip = function (ip) {
     if (net.isIPv4(ip)) {
         return re_private_ipv4.test(ip);
     }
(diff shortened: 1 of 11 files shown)


baudehlo:
We need to keep the original name around for people who have written their own plugins.

msimerson:
Good point. I just pushed with this change, I think that'll work, right?

```js
exports.is_rfc1918 = exports.is_private_ip;
```

baudehlo:
Yeah that will be fine. I haven't reviewed the rest of the changes yet.

smfreegard:
Yes - this function was poorly named; but I simply wanted a quick fix for an edge case before 2.6 hit, but it appears the changes never made it in anyway.

I'd actually started to re-write this completely as using regexps for this isn't the 'right' way to do it but ultimately realised that it was going to too big of a change, so I made the minimum amount of changes necessary.

The 'right' way for this function to work is using the ipaddr.js module and it's ranges and maybe creating a bitmask value so the plugin author can decide what is included or excluded, so I think both my original PR and this one might as well be closed and I'll work on a more substantial re-write of it.

PR #858, opened 2015-02-22
net_utils: refactoring for IPv6 compatibility, simplicity and maintainability

### LINKED ISSUE #850: Error: Outbound email with "auth/flat_file"

msimerson:
* rename net_utils.is_rfc1918 -> is_private_ip
* closes #850
* added net_utils.is_local_ipv4, is_private_ipv4, & is_local_ipv6.
    * updated is_private_ip to use them
    * adds IPv6 link local to is_private_ip
    * adds additional test coverage
    * considerably easier to maintain than one regex for ipv4 and one for ipv6
    * no significant performance impact (0.05 second difference on 2 million iterations)
    * side note: using ?: in /regex/.test(str) seems to make no performance different (less than rounding error in testing).

### DIFF
--- a/net_utils.js
+++ b/net_utils.js
@@ -5,9 +5,6 @@ var config = require('./config');
 var net    = require('net');
 var punycode = require('punycode');
 
-// Regexp to match private IPv4 ranges
-var re_private_ipv4 = /^(?:10|127|169\.254|172\.(?:1[6-9]|2[0-9]|3[01])|192\.168)\..*/;
-
 var public_suffix_list = {};
 load_public_suffix_list();
 
@@ -34,8 +31,8 @@ exports.is_public_suffix = function (host) {
 };
 
 exports.get_organizational_domain = function (host) {
-    // the domain that was registered with a domain name registrar
-    // See https://datatracker.ietf.org/doc/draft-kucherawy-dmarc-base/?include_text=1
+    // the domain that was registered with a domain name registrar. See
+    // https://datatracker.ietf.org/doc/draft-kucherawy-dmarc-base/?include_text=1
     //   section 3.2
 
     if (!host) return null;
@@ -121,13 +118,10 @@ exports.hex_to_dec = function (h) {
 };
 
 exports.ip_to_long = function (ip) {
-    if (!net.isIPv4(ip)) {
-        return false;
-    }
-    else {
-        var d = ip.split('.');
-        return ((((((+d[0])*256)+(+d[1]))*256)+(+d[2]))*256)+(+d[3]);
-    }
+    if (!net.isIPv4(ip)) { return false; }
+
+    var d = ip.split('.');
+    return ((((((+d[0])*256)+(+d[1]))*256)+(+d[2]))*256)+(+d[3]);
 };
 
 exports.octets_in_string = function (str, oct1, oct2) {
@@ -156,7 +150,6 @@ exports.octets_in_string = function (str, oct1, oct2) {
 };
 
 exports.is_ip_in_str = function(ip, str) {
-    // Only IPv4 for now
     if (!str) { return false; }
     if (!ip) { return false; }
     if (!net.isIPv4(ip)) {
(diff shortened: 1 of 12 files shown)


msimerson:
@smfreegard , per your comments in #854, I'm wondering if this might be a stepping stone on the way there. To make it easier to extend / maintain in the future, I've broken up is_public_ip into discreet functions. Now the big regex is a series of discreet ones with test coverage (and to my surprise, it's not slower!). Each regex can now be replaced with ipaddr functions, the test coverage doesn't need much of any changes, and net_utils would be a fair bit cleaner for it.

I'd also be curious to see what difference using ipaddr.js makes. It'd be great if it were measurably faster, but maybe staying with "less right" regexes is faster enough to be worth while.

net_utils.js
```diff
@@ -121,79 +118,121 @@ exports.hex_to_dec = function (h) {
 };
 
 exports.ip_to_long = function (ip) {
-    if (!net.isIPv4(ip)) {
-        return false;
-    }
-    else {
```
smfreegard: Can't see the point of keeping the nesting here.

smfreegard:
The is_ip_in_str() function needs tests to prove that your refactored version returns the same results before this is merged.

I still don't like the is_private_ip etc. changes as it means were going to have to do multiple function calls depending on what we want and it doesn't have full coverage of all the relevant things that a plugin author might want.

My idea was something like this:

`````
var ipaddr = require('ipaddr.js');

var ranges = {
    IPV4_UNSPECIFIED: 1,
    IPV4_BROADCAST:   2,
    IPV4_MULTICAST:   4,
    IPV4_LINKLOCAL:   8,
    IPV4_LOOPBACK:    16,
    IPV4_PRIVATE:     32,
    IPV4_RESERVED:    64,
    IPV6_UNSPECIFIED: 128,
    IPV6_LINKLOCAL:   256,
    IPV6_MULTICAST:   512,
    IPV6_LOOPBACK:    1024,
    IPV6_UNIQUELOCAL: 2048,
    IPV6_IPV4MAPPED:  4096,
    IPV6_RFC6145:     8192,
    IPV6_RFC6052:     16384,
    IPV6_6TO4:        32768,
    IPV6_TEREDO:      65536,
    IPV6_RESERVED:    131072,
    ANY_PRIVATE:      (8 + 16 + 32 + 256 + 1024 + 2048),
};


exports.ip_to_range = function (ip) {
    var parsed = ipaddr.parse(ip);
    var kind = parsed.kind();
    var range = parsed.range();
    if (kind === 'ipv4') {
        switch (range) {
            // From source code of ipaddr.js
            case 'unspecified':
                return ranges.IPV4_UNSPECIFIED;
            case 'broadcast':
                return ranges.IPV4_BROADCAST;
            case 'multicast':
                return ranges.IPV4_MULTICAST;
            case 'linkLocal':
                return ranges.IPV4_LINKLOCAL;
            case 'loopback':
                return ranges.IPV4_LOOPBACK;
            case 'private':
                return ranges.IPV4_PRIVATE;
            case 'reserved':
                return ranges.IPV4_RESERVED;
            default:
        }
    }
    else if (kind === 'ipv6') {
        switch (range) {
            // TODO
        }
    }

    // Anything else; return 0
    return 0;
}


// test
var r = this.ip_to_range('169.254.1.1');
// See if the address is APIPA, RFC1918 or Loopback
if (r & (ranges.IPV4_LINKLOCAL | ranges.IPV4_PRIVATE | ranges.IPV4_LOOPBACK)) {
    console.log(true);
}
if (r & ranges.ANY_LOCAL) console.log(true);
`````

This would then require just a single function call .vs. a bunch of them like you have right now and allows maximum flexibility to the plugin author about what should be included or excluded.

However - I don't have the time to implement this or benchmark it against a regexp approach right now, however I think I'd prefer accuracy, simplicity and flexibility over any potential speed increase when using a quick'n'dirty regexp in this case.

So unless you have a good argument for doing it another way; I'd prefer you either wait until I have time to do this, write tests and docs etc. or leave is_rfc1918 as it is for now.

**Comment by msimerson** [comment]
Okay @smfreegard , please have another look. I have added documentation to Net_Utils.md. This PR adds IPv6 link local detection to is_private_ip, a much improved function name. When you get around to implementing `ipaddr.js`, you need only update is_private_ip and nuke the other 3. Until then, the new functions are easy to read, easy to extend, covered by tests, and just as performant as shoving 'em all into two long regex.

File: net_utils.js [hunk 2]
```diff
@@ -187,10 +180,59 @@ exports.is_ip_in_str = function(ip, str) {
...
+    if (/^192\.168\./.test(ip)) return true;    // 192.168/16
+    // 172.16/12  ->   172.16/16 .. 172.31/16
+    if (/^172\.(1[6-9]|2[0-9]|3[01])\./.test(ip)) return true;
+
+    return false;
+};
+
+exports.is_local_ipv4 = function (ip) {
+    // 127/8 (loopback) and 0/8 (this network)  # RFC 1122
+    if(/^(127|0)\./.test(ip)) return true;
```
smfreegard: This is wrong IMO.   0.0.0.0/8 shouldn't be included here.

RFC 1122 states:
`````
Specified host on this network. It MUST NOT be sent, except as a source address as part of an initialization procedure by which the host learns its full IP address.
`````

So you'll never see hosts with 0.0.0.0 addresses connecting to Haraka and if you do; they certainly shouldn't be treated specially, so I don't think it should be present.

File: net_utils.js [hunk 3]
```diff
@@ -187,10 +180,59 @@ exports.is_ip_in_str = function(ip, str) {
...
+
+    // link local: 169.254/16) RFC 3927
+    if (/^169\.254\./.test(ip)) return true;
+
+    return false;
+};
+
+exports.is_local_ipv6 = function (ip) {
```
smfreegard: This function is missing uniqueLocal (fc00::/7)

smfreegard:
Ok - I've benchmarked the ipaddr.js methods that I was proposing and they're much slower because the ranges are compiled on each call to parse(), so the regexps are definitely the way to go.

net_utils.js
```diff
@@ -187,10 +180,59 @@ exports.is_ip_in_str = function(ip, str) {
     return false;
 };
 
-exports.is_rfc1918 = function (ip) {
-    return (net.isIPv4(ip) && re_private_ipv4.test(ip));
+exports.is_private_ipv4 = function (ip) {
+
+    // RFC 1918, reserved as "private" IP space
+    if (/^10\./.test(ip)) return true;          // 10/8
```
smfreegard: I just benchmarked - and this:

````
var re_1 = /^10\./;
var re_2 = /^192\.168\./;
var re_3 = /^172\.(1[6-9]|2[0-9]|3[01])\./;

exports.is_private_ipv4 = function (ip) {
    if (re_1.test(ip) || re_2.test(ip) || re_3.test(ip)) return true;
    return false;
}
````

And as you would expect - this Is always faster than making V8 compile the regexps every call as you're doing here.  Which is why the regexps were compiled on the previous version.

msimerson: How did you benchmark?

I benchmarked by calling the is_private_ip function with 5 different IPs (2 hits, 3 misses), which in turn called the is_local_ipv4 and is_private_ipv4 functions (as existing in this PR). My understanding is that node compiles functions (including the regexes) in them, which would explain why I saw no performance difference.

smfreegard: I benchmarked just the is_private_ipv4() function above against your stock version through 1,000,000 iterations and I found the pre-compiled regexps to be on average 1 second faster.   I tested a static value that was always matched in both cases.

`````
var result;
for (var i=0; i<10000000; i++) {
    result = exports.is_private_ipv4('192.168.1.1');
}
`````

msimerson:
Further updated. Removed 0/8, added IPv6 unique local, moved REs out of functions to save 1s of processing on 1 million iterations on Steve's computer. The difference is < 0 .1 seconds on my 3.5GHz Core i5 (Retina iMac).

smfreegard:
1s of processing was a 25% improvement here.   And it could easily be explained by the difference in Node version (0.10.22 here) which might have different versions of V8.

PR #863, opened 2015-02-23
updated Public Suffix List

msimerson:
yeah, should have done that *before* release


baudehlo:
Easy to do 2.6.1 with that if we need to.

PR #865, opened 2015-02-23
refactor octets_in_string from is_ip_in_str

msimerson:
* is_ip_in_str, invert IPv4 detection and exit early
* test/net_utils: lint updates
* removes a duplicated is_public_suffix test

### DIFF
--- a/net_utils.js
+++ b/net_utils.js
@@ -1,4 +1,4 @@
-"use strict";
+'use strict';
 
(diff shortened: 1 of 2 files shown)


msimerson:
There are already tests for is_ip_in_str that are untouched. This extracts the duplicated block of code from is_ip_in_str into a function and adds tests covering the new function.

PR #866, opened 2015-02-23
add additional documentation in config/plugins

msimerson:


config/plugins
```diff
@@ -1,7 +1,14 @@
-# default list of plugins
+# This file lists plugins that Haraka will run
+#
+# Plugin ordering often matters, run 'haraka -o -c /path/to/haraka/config'
+# to see the order plugins (and their hooks) will run in.
```
smfreegard: `to see the order plugins will run in each hook.`

haraka -o shows the ordering by hook and not by plugin

Issue #862, opened 2015-02-23
config file watching needs a sedative


msimerson:
I'm thinking that after a change is detected, we should set a timer that resets every time we receive the event, and after we haven't received a notice for > 1 second, **then** reload the file.

Failing to do that, we're parsing config files while they're being written. That means there's likely a period where Haraka is running with a config file that is incomplete. This seems very likely to lead to random and very difficult to track down crashes. (I've seen it happen once).

````
Feb 22 17:09:19 node haraka[9339]: [INFO] [-] [core] Detected change, reloading /usr/local/haraka/config/smtp.ini
Feb 22 17:09:19 node haraka[9340]: [INFO] [-] [core] Detected change, reloading /usr/local/haraka/config/smtp.ini
Feb 22 17:09:19 node haraka[9338]: [INFO] [-] [core] Detected change, reloading /usr/local/haraka/config/smtp.ini
Feb 22 17:09:19 node haraka[9340]: [INFO] [-] [core] Detected change, reloading /usr/local/haraka/config/smtp.ini
Feb 22 17:09:19 node haraka[9338]: [INFO] [-] [core] Detected change, reloading /usr/local/haraka/config/smtp.ini
Feb 22 17:09:19 node haraka[9339]: [INFO] [-] [core] Detected change, reloading /usr/local/haraka/config/smtp.ini
....
Feb 22 17:11:32 node haraka[9626]: [INFO] [-] [core] Detected change, reloading /usr/local/haraka/config/public-suffix-list
Feb 22 17:11:32 node haraka[9627]: [INFO] [-] [core] Detected change, reloading /usr/local/haraka/config/public-suffix-list
Feb 22 17:11:32 node haraka[9625]: [INFO] [-] [core] Detected change, reloading /usr/local/haraka/config/public-suffix-list
````

godsflaw:
I see no other way to solve the problem.  I might even make it 2 seconds as
that is an unheard of amount of time for the writer to not have had its
buffer flushed to disk, and I can think of no real problem for us having a
2s delay on reading a config. +1

Note: I have seen installs where the disks write buffer flushes every
second to improve write performance.

msimerson:
> writer to not have had its buffer flushed to disk

Does it matter whether the file change is flushed/committed to disk? I'm under the impression that at least since the 1980s, the OS solved that for us and accesses are always going to get the version the OS has in its buffers (or disk, in the case of cache misses) at **that** moment.

[...]

PR #868, opened 2015-02-24
Potential fix for #862

smfreegard:


msimerson:
I finally got around to testing this. I *think* the reason I see this message run 3 times is because I run in cluster mode with 2 children + 1 master = 3. You can see that each message has a separate PID.
````
Mar 23 12:56:50 node haraka[79427]: [INFO] [-] [core] Reloading file: /usr/local/haraka/config/log.elasticsearch.ini
Mar 23 12:56:50 node haraka[79426]: [INFO] [-] [core] Reloading file: /usr/local/haraka/config/log.elasticsearch.ini
Mar 23 12:56:50 node haraka[79420]: [INFO] [-] [core] Reloading file: /usr/local/haraka/config/log.elasticsearch.ini
````

PR #878, opened 2015-03-05
Fixed bug with reading of config file for smtp_forward plugin

gziskind:
When reading the smtp_forward.ini configuration, the get_config method returns plugin.cfg.main, but in parsing the ini file, data under the [auth] header, existing under plugin.cfg.auth, not plugin.cfg.main.auth.

Added plugin.cfg.auth to plugin.cfg.main.auth so it can be used by the plugin.

Otherwise, the [auth] configuration are not found.

### DIFF
--- a/plugins/queue/smtp_forward.js
+++ b/plugins/queue/smtp_forward.js
@@ -22,6 +22,8 @@ exports.register = function () {
 exports.get_config = function (connection) {
     var plugin = this;
 
+    plugin.cfg.main.auth = plugin.cfg.auth;
+
     if (!connection.transaction) return plugin.cfg.main;
     if (!connection.transaction.rcpt_to[0]) return plugin.cfg.main;
     var dom = connection.transaction.rcpt_to[0].host;


smfreegard:
I already fixed this a few days ago albeit in a slightly different manner.  https://github.com/baudehlo/Haraka/pull/872

smfreegard:
BTW - your way would prevent having a different AUTH for different domains which is why I didn't do it that way.

salvop:
Yes, we need to have authentication peer domain.
Will be nice to have it working and have also  a sample in the doc.

Config should be like:

host=1.2.3.4
user= user
pass= pass

[example1.com]
host=1.2.3.5
user= user1
pass= pass1

[example2.com]
host=1.2.3.5
user= user2
pass= pass2

[example3.com]
host=1.2.3.6
salvo

smfreegard:
See https://github.com/baudehlo/Haraka/blob/master/docs/plugins/queue/smtp_forward.md

salvop:
Oh yes Steve! I missed the last updates! Great  today i'll test it!

gziskind:
yeah mine was just a quick fix to get it working, which is why i didnt do a merge request in the first place. only did it cause someone asked. thanks smfreegard

PR #879, opened 2015-03-06
allow domains with underscore

mjpearson:
fixes 'Invalid domain in address' error parsing domains which contain underscores.  Underscores in domain name do not break spec and are often used.


baudehlo:
Merged. Could you also patch it in https://github.com/haraka/node-address-rfc2821 and add a test?

mjpearson:
Thanks, yes can do

PR #884, opened 2015-03-19
Add AUTH support to outbound

smfreegard:
Requires the use of the get_mx hook to return an MX object that supplies `auth_user` and `auth_pass` properties.

Also adds an auth=[Y|N] to the outbound delivery line to show if AUTH was used:

````
[NOTICE] [5E910B8A-C52B-46E9-8EB0-8B88FCE139FF.1.1] [outbound] delivered file=1426782311656_0_25646_3451.mail1-ec2.fsl.com domain="tnpi.net" host="mail.theartfarm.com" ip=66.128.51.165 port=25 mode=SMTP tls=Y auth=Y response="ok 1426782313 qp 36423 (995AD5EE-DC67-4E4A-88BB-A2CE2075960F.1)" delay=2.205 fails=0 rcpts=1/0/0
````

### DIFF
--- a/docs/Outbound.md
+++ b/docs/Outbound.md
@@ -99,10 +99,16 @@ The MX record is sent via next(OK, mx) and can be one of:
     * hostname:port
     * ipaddress
     * ipaddress:port
-* An MX object of the form: `{priority: 0, exchange: hostname}` and optionally
-a `port` value to specify an alternate port, and a `bind` value to specify an
-outbound IP address to bind to and a `using_lmtp` boolean to specify that
-delivery should be attempted using LMTP instead of SMTP.
+* An MX object of the form: `{priority: 0, exchange: hostname}` with the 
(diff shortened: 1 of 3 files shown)


msimerson:
I helped @smfreegard test this against my mail server, so I can attest it works. Besides not introducing any new tests in t/outbound.js (which is already quite bare), it LGTM.

smfreegard:
To confirm; I have this running in production and I have tested AUTH PLAIN, AUTH LOGIN and AUTH CRAM-MD5 (the only methods supported) successfully and I have tested all the failure modes e.g. AUTH not offered, no matching methods or invalid method requested via `auth_type`.

msimerson:
Assuming no objections arise in the next couple days, I'll happily merge.

baudehlo:
Looks OK to me.

Issue #726, opened 2014-10-29
base class for Redis


msimerson:
Haraka might benefit from a base class for Redis. Rather than having each plugin do common things like require('redis'), redis_connect(), redis_ping(), have those implemented in a base class that can be inherited.

celesteking:
DB indexes? Sounds good. How about allowing a plugin have a separate redis config?

for ex.: redis.ini:
```ini
[default]
host = 1.com
port = 1234
db = 0

[karma]
host = localhost
db = 1
password = xxx

[queue/random_delivery]
port = 12354
```

smfreegard:
Indeed; you might want one plugin to use in-memory and another to use Redis too.

Equivalent functionality can easily be achieved by adding a Redis plugin and exporting it via server.notes to plugins that want to use it (which is what I do here).

msimerson:
considering this done with the merge of plugins/redis.

PR #895, opened 2015-03-31
nosql: ram, SSC, or redis based key-values

msimerson:
with a single API

see also #726 (base class for Redis)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ node_js:
 #   - "0.6"   # no longer supported by async
     - iojs
 
+services:
+    - redis-server
+
 matrix:
     allow_failures:
         - node_js: "0.8"
--- a/config/nosql.ini
+++ b/config/nosql.ini
@@ -1,0 +1,16 @@
+; nosql.ini
+
+[store]
+; default: volatile RAM, with Stong Store Cluster (if clustered)
+; backend=redis
+
+
+[cluster]
+; expire: how many minutes entries persist, default: 10
+expire=10
+
+
+[redis]
+; host=localhost
+; port=6379
+; db=0
--- a/docs/Nosql.md
+++ b/docs/Nosql.md
@@ -1,0 +1,140 @@
+# nosql.js
+
+Store stuff in memory-backed objects, smartly.
+
+## Usage
+
+    var nosql = require('./nosql');
+
+Includes the following functions:
+
+## set
+
+    nosql.set('foo', 'bar', function (err, result) {
+        if (err) {
+            // error handling code
+            return;
+        }
+        // do fun stuff with result
+    });
+
+You're confident it'll work out?  Skip the callback, it's optional. For all of these methods.
+
+    nosql.set('foo', 'bar');  // going commando
+
+## get
+
+    nosql.get('foo', function (err, result) {
+        if (err) { return; }
+        // result == 'bar'  (because that's what we set it to)
+    });
+
+## del
+
+    nosql.del('foo', function (err, result) {
+        if (err) { return; }
+        // result == 1   (# of objects keys deleted)
+    });
+
+## incrby
+
+    nosql.incrby('my_counter', 1, function (err, result) {
+        if (err) { return; }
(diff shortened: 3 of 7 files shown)


smfreegard:
Personally - I'm -1 on this for the following reasons:

- Redis provides so much more functionality over what is exposed by this e.g. transactions, hashes, sets, sorted sets, hyper log log's, key expiry etc. which are essentially unavailable if you use this API.
- The in-memory code is very basic and no better than what some of the plugins already do themselves, without the need for any namespacing etc.   There's no way to share the keys between processes; which could be achieved via the Cluster API (e.g. have the master store the keys and then have the children get/set them via the API) and it's pretty similar to the result_store code which is already available.
- There's no way to select different Redis databases or have connections to different hosts or mix in-memory for some plugins and Redis for others (which you'd probably want to do for concurrency limits).
- Could be implemented by base class plugins (like dns_list_base) e.g. redis_base, kvm_base, nosql_base that would then be inherited as desired by the plugins which I think would be a bit more useful to plugin authors (and more flexible).

msimerson:
> Redis provides so much more 

I'm not trying to re-invent Redis. For applications that want / need Redis, it's *still* there. This PR provides a way for plugins (some committed, some not) and base libs (auth_base, tls_socket) to benefit from basic NoSQL functionality **and** do so **without** a redis dependency. When Redis is available, switching a single config setting provides many of the benefits of Redis with no code changes.

This is not *instead* of a redis library, this is written to be easily extended to support any other key-value store without having to modify the plugins or base libraries that depend on it.

If Redis is available, why would you ever prefer direct RAM storage?

msimerson:
# TODO
* add support for [Strong Store for Cluster](http://docs.strongloop.com/display/SLC/Strong+Store+for+Cluster) when in cluster mode.

msimerson:
In the tests/nosql.js file, there are three functions, exports. lifecycle_ram, exports. lifecycle_ssc, and exports.lifecycle_redis.  All three are identical, which is exactly what I want (test each "backend" the same), but I can't figure out how to get nodeunit to run them unless they're triplicated. Is there a way get this more DRY?

msimerson:
This module wants a cheap reliable way to determine what plugin is calling the functions it exports. This would seemingly be a straight forward problem to solve with `module.parent.filename`. In testing, nosql._get_caller_name() worked exactly as expected and I thought I had that problem licked, until I saw the name of the plugin in the MONITOR output of redis-cli. Instead of being `limit`, as I expected, it was instead `plugins`.

If I move the `var caller = _get_caller_name()` out of the exported functions and to the **limit** module scope, it works as I expect and all further `nosql.[set|get|*]()` operations get the value `limit` for the plugin name. However, subsequent modules that load nosql also get the value `limit` for their plugin name because of the node module cache.

The working here invalidates the cache, causing future loads of nosql by other plugins to load their own instance of `nosql`, working around the singleton issue. 

Short of further changes to plugins.js, I can't see any other way around this.

smfreegard:
Ok - thought about this some more and this is the wrong way go.   I think you should remove the changes to plugins.js.   They're simply too hacky and fragile.

I think the correct way forward for this module is to run it into a 'proper' class with a constructor that you can use to set the options on a plugin-by-plugin basis e.g.

`````
function NoSQL (collection, options) {
    this.collection = collection;
    // other options passed in in object, like TTL    
}

NoSQL.prototype.get = function () { ... };
NoSQL.prototype.set = function () { ... };
NoSQL.prototype.incrby = function () { ... };
NoSQL.prototype.reset = function () { ... };

module.exports = NoSQL;
`````

Then in the plugins you'd load nosql.js and use it like so:

`````
var NoSQL = require('nosql');
var nosql = new NoSQL(plugin.name, { ttl: null });

exports.hook_lookup_rdns = function (next, connection) {
    nosql.incrby(...);
    return next();
};
`````

baudehlo:
Could be a good place to start moving things to npm too. Create a
haraka-nosql package there (happy to create the repo on github.com/haraka)
and add it as a dependency.

msimerson:
https://github.com/haraka/haraka-nosql

Issue #754, opened 2014-11-25
make rate_limit plugin smarter


msimerson:
# bounces

Increment counters for every user that records the number of invalid recipients they've mailed to, by listening on the bounce hook.
# auth requests

Increment counters every time an IP attempts to authenticate. Basically, make the current back-off mechanism persist across connections. ~~(Once done, remove this feature from `karma` plugin.)~~

Dexus:
@msimerson @smfreegard 
can this be closed or is this a open task?

msimerson:
It's still open. The limit plugin does some of that "smarter" work but the rate_limit plugin also needs additional work.

PR #896, opened 2015-03-31
new plugin: limit

msimerson:
### limits for
* [x] connection concurrency
* [x] recipients
* [x] unrecognized commands
* [x] errors

### future plans:
- [x] remove recipient limits from karma
- [ ] add auth attempt limits to rate_limit #754
- [x] remove connection concurrency from karma
- [ ] remove connection concurrency from rate_limit
- [ ] deprecate max_unrecognized_commands plugin
- [ ] deprecate rcpt_to.max_count

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ node_js:
 #   - "0.6"   # no longer supported by async
     - iojs
 
+services:
+    - redis-server
+
 matrix:
     allow_failures:
         - node_js: "0.8"
--- a/config/limit.ini
+++ b/config/limit.ini
@@ -1,0 +1,45 @@
+; limits imposed on connection(s)
+
+[storage]
+; backends supported: ram, ssc (strong storage cluster), redis
+; default: ram if not clustered, ssc if clustered.
+; backend=ram
+;
+; expire: minutes concurrency information is cached in RAM, default: 10
+; expire=10
+
+; there are no defaults unless shown. Limits are disabled until a max value is set.
+; help: haraka -h limits
+
+[concurrency]
+; max=3
+; disconnect_delay=5
+; reset=10   // default: minutes
+
+; History: when enabled, the value must be a plugin which stores IP history
+;   results. The result store must have a positive value for good connections
+;   and negative integers for poor / undesirable connections. At present,
+;   karma is the only such plugin.
+; history=karma
+; history_bad=1
+; history_none=3
+; history_good=10
+
+
+[recipients]
+; max= 20
+; max_relaying= 100
+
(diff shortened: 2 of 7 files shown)


smfreegard:
Ok - there's a few issues with this:

- The lookup_rdns, and disconnect hooks need to run before any other plugins (as other plugins might return OK and prevent this from running), so this will been to be listed at the very top of the `config/plugins` file *BUT* you want the unrecognized_command hook to run *last* (e.g. at the bottom of `config/plugins`) as you want other plugins that implement commands via the hook to avoid this by returning OK otherwise you'll end up always counting valid STARTTLS or AUTH commands as unrecognized incorrectly.   Basically you're going to have to split this into two plugins and document it as such for this to work correctly.

- The concurrency code has numerous issues if you were to run it under Redis and even more if you had multiple Haraka instances sharing a Redis instance.   If you stop Haraka, then any concurrent connections that were running at the time will not be decremented and you'll quickly end up blocking valid hosts indefinitely.   Additionally when running in-memory, you'll want to make sure that the counter can't go negative (I've seen this happen in my own code, haven't been able to trace it as yet).

msimerson:
> this will need to be listed at the very top of the config/plugins

I disagree. I plan to put limit towards the bottom of config/plugins, as I do with karma.

1. the only plugin that calls next(OK) on the lookup_rdns hook is `connect.rdns_access`, which has been deprecated by `access`.
2. IMO, the *lookup_rdns* and *disconnect* hooks shouldn't ever skip plugin processing. Plugin consumers of disconnect: graph, karma, limit, process_title, profile, rate_limit, log.elasticsearch.
3. No plugins call next(OK) on disconnect. Any that do deserve the pain they cause. (I say this, knowing I'm likely to want to do this some fine day, but hoping we find a way to polish the sharp edges of plugin ordering before then).
4. The concurrency issue is worked around by clearing the concurrency keys every 10 minutes (by default).

msimerson:
> Additionally when running in-memory, you'll want to make sure that the counter can't go negative (I've seen this happen in my own code, haven't been able to trace it as yet).

While testing this on a live server, I got to witness the negative counter issue. I put in a little code to [repair and bark about it](https://github.com/baudehlo/Haraka/pull/896/files#diff-096ff9f2e86fba8fa0ab4e6ebb896427R135), and then figured out exactly what the issue was: I had a conditional in the incrementing code that sometimes prevented the lookup_rdns function from incrementing. Of course, the disconnect always fires just fine. I refactored the code so that increment always happens first, and *then* other stuff with error handling is off in it's own function. Haven't seen this issue since.

PR #903, opened 2015-04-02
connection: return 450/550 for plugin DENY*

msimerson:
instead of 452/552 'insufficient system storage'

looks like a copy/paste error, as there are two cases where a 452 in connection remains, for disk space errors.


msimerson:
Updates to RFC 5321  (vs RFC 2821)

PR #904, opened 2015-04-02
access: provide a debug level msg for empty re

msimerson:
For @celesteking


celesteking:
thanks

celesteking:
Now, after this has been merged, I'm wondering why the heck it outputs:
[DEBUG] [-] [access] empty file: helo.checks.regexps

while 
```
$ cat config/helo.checks.regexps 
ylmf\-pc
```

PR #908, opened 2015-04-04
karma: add rule tests for plugins with / in name

msimerson:
(such as auth/auth_*)

for @celesteking

This rule should have worked for you:
````
results.auth/auth_base.fail      = -4  if length gt 0
````

However, Haraka's custom ini parser didn't permit it to work because of the forward slash:
````
[ERROR] [-] [core] Invalid line in config file 'karma.ini': results.auth/auth_base.fail = -4  if length gt 0
````

So attached is also a commit which adds the '/' char to the permitted list.

### DIFF
--- a/configfile.js
+++ b/configfile.js
@@ -8,7 +8,7 @@ var yaml = require('js-yaml');
 // for "ini" type files
 var regex = exports.regex = {
     section:        /^\s*\[\s*([^\]]*?)\s*\]\s*$/,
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by celesteking** [comment]
What can I say. It works:
```
 [karma] applied results.auth/auth_anti_brute.fail:-10
```

thanks sweeties.

smfreegard:
Looks OK to me.   @baudehlo - I don't think allowing slashes in the key name will be a problem?

baudehlo:
Yeah I don't see why it would be a problem.

Issue #909, opened 2015-04-04
Allow specifying plugin name instead of plugin object in results_store.add() call


celesteking:
connection.results.add() should accept a String instead of current Object. 
```{name: 'blah'}``` works, but garbles the output.

msimerson:
It [can't](https://github.com/baudehlo/Haraka/blob/master/result_store.js#L83). If you change the [API of logger](https://github.com/baudehlo/Haraka/blob/master/logger.js#L166) so I can pass in an initial string instead of a plugin name, I'd happily update result_store.

celesteking:
```log_if_level``` is essentially using implicit ```plugin.toString()``` call, so that shouldn't be a problem.

msimerson:
Are you [sure](https://github.com/baudehlo/Haraka/blob/master/logger.js#L189) about that? I'd be happy to be wrong.

PR #910, opened 2015-04-04
results: permit a string or object as 1st arg

msimerson:
In #909 @celesteking volunteered to test.

### DIFF
--- a/docs/Results.md
+++ b/docs/Results.md
@@ -161,12 +161,12 @@ Same thing with **has** (retrieve a string match):
         // some_test passed (1x)
     }
 
-So the syntax for using **has** is a little more pleasant.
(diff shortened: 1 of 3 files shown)


celesteking:
give me some time (a week)

celesteking:
It logs:
```[INFO] [896A7AD8-B0AA-46E7-9EB1-F1EF93486548] [core] { name: 'auth' }fail:mismatch```
instead of ```[auth] fail:mismatch```

msimerson:
@celesteking, correct, because as I pointed out in #909, the logger expects a plugin as the first argument. If you want ResultStore to work this way, then logger.js needs changing. If you're inclined to do that, I'll leave this PR open.

PR #914, opened 2015-04-07
add HTTP support

### LINKED ISSUE #433: web servers and ports
Node.js makes it really easy to have a web interface, and I'm the second, and surely not the last guy to add one to Haraka. There's different ways to crack this nut:

1. make each one listen on their own port. I'm not particularly fond of that idea.

2. have a 'default' web server that plugins can register URL space in. The default web server could have:

    - an index with a list of plugin providers
    - a dispatch, so /watch/ would go to my watch.js, /graph/ to graph.js, etc...
    - uses the same TLS/SSL certs as the SMTP servers
    - can auth against SMTP-AUTH provider(s)

I really option 2 and would like feedback on what type of implementation might be desirable.

msimerson:
* server: add http child listeners
* server: only load http if configured
* http.ini: new config file
* updated plugins/graph for new HTTP server
* warn if http enabled but can't load express or ws
* add config/bower.json, Gruntfile, package.json

### Pros
* allows multiple plugins to serve on the same HTTP(S) port
* provides an index/landing page
* closes #433 
* the basis for facilities to manage the outbound queue (@baudehlo in #443)

### Known Issues
* ~~the html/* directory is not included. Yet.~~ (installed with bower)
* ~~how to install web dependencies and avoid test failures re: them? (optionalDeps isn't ideal)~~ (bower)
* ~~docroot handling for haraka and per-plugin~~ (each plugin registers a route)

### Future Ideas
* populate plugin tabs dynamically (based on plugins that register routes)
* auth via any SMTP-AUTH provider
* provide a consistent UI via standardised views/templates (@baudehlo in #443)

Reopens #734

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 .coveralls.yml
 Haraka
 *~
+bower_components
--- a/TODO
+++ b/TODO
@@ -52,9 +52,6 @@ Move the following plugins:
  - test_queue   -> queue/test_queue
 
 Built-in HTTP server
-- instead of each plugin bringing their own, and getting relegated to some
-  god-forsaken port number:
-- a dispatch mechanism, so each plugin gets their own URL space
 - uses the same TLS/SSL certs as smtpd
 - auth against SMTP-AUTH provider
 
--- a/config/graph.ini
+++ b/config/graph.ini
@@ -3,10 +3,6 @@
 ;db_file=:memory:
 db_file=graphlog.db
 
-; The port to listen on for http. Default: `8080`.
-http_addr=127.0.0.1
-http_port=8080
-
 ; Regular expression to match plugins to ignore for logging.
 ; Default: `queue|graph|relay`
 ignore_re=`queue|graph|relay`
--- a/config/http.ini
+++ b/config/http.ini
@@ -1,0 +1,7 @@
+
+; listen: the HTTP address:port(s) to listen on
+; default: [::]:80 (port 80 on all IPv4 and IPv6 addresses)
+; listen=[::]:80
+
+; docroot: the directory where web content is served from
+;docroot=/usr/local/haraka/html
--- a/http/Gruntfile.js
+++ b/http/Gruntfile.js
@@ -1,0 +1,72 @@
+/* jshint camelcase: false */
+module.exports = function(grunt) {
+
+    grunt.loadNpmTasks('grunt-contrib-clean');
+    grunt.loadNpmTasks('grunt-bower-install-simple');
+    grunt.loadNpmTasks('grunt-bower');
+
+    var bpSrc = [], bpDest = [];
+    [
+        "404.html",
+        "apple-touch-icon.png",
+        "crossdomain.xml",
+        "browserconfig.xml",
+        "favicon.ico",
+        "robots.txt"
+    ].forEach(function (file) {
+        bpSrc.push("dist/" + file);
+        bpDest.push("html/" + file);
+    });
+
+    // Project configuration.
+    grunt.initConfig({
+        pkg: grunt.file.readJSON('../package.json'),
+        'bower-install-simple': {
+            options: {
+                color: true,
+                directory: 'bower_components'
+            },
+            'prod': {
+                options: {
+                    production: true
+                }
+            },
+        },
+        bower: {
+            prod: {
+                dest: 'html',
+                js_dest: 'html/js/vendor',
+                css_dest: 'html/css/vendor',
+                fonts_dest: 'html/fonts/vendor',
+                options: {
+                    expand: false,
+                    keepExpandedHierarchy: false,
+                    packageSpecific: {
+                        bootstrap: {
+                            files: [
+                                "dist/css/bootstrap.min.css",
+                                "dist/css/bootstrap-theme.min.css",
+                                "dist/js/bootstrap.min.js"
+                            ]
+                        },
+                        jquery: {
+                            files: [
+                                "dist/jquery.min.js"
+                            ]
+                        },
+                        "html5-boilerplate": {
+                            files: bpSrc
+                        }
+                    }
+                },
+            }
+        },
+        clean: {
+            bower: [ 'html/**/vendor' ],
+            boilerplate: bpDest,
+            // dist: ['node_modules', 'bower_components']
+        },
+    });
+
+    grunt.registerTask('default', ['bower-install-simple', 'bower']);
+};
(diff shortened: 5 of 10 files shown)


smfreegard:
Ok - we might as well start the discussion here.

Pros:
* Single point of configuration
* Provides a single HTTP port for all plugins
* Completely optional; doesn't add dependencies to packages.json
* Plugins can register mount points via hooks

Cons:
* Undocumented
* HTTP service is running on the same event loop as the SMTP listener
  - Security implications; there is no separation of concerns.
  - HTTP listener will have the same user/group as the SMTP listener and will be able to access all files created by the Haraka and the plugins and there will be no way to prevent it from doing so.
  - Express bugs could crash Haraka
  - Additional resource usage, Express etc. will be loaded by master and children.  Memory, CPU cost etc.

msimerson:
# Pros

* HTTP listener has the same user/group as the SMTP listener and is able to access all files created by the Haraka
    * permits a config plugin to modify / update config files and settings via HTTP interface
    * permits a plugin like `watch` to collect real-time SMTP info and display via HTTP

> Express bugs could crash Haraka
Sure, but that's little different than any other library Haraka uses. In the case of Express, it's extremely heavily tested, very minimal (since 4.x) and focused and in widespread use. I'd be more concerned with bugs in http serving plugins than with Express itself.

> Additional resource usage, Express etc. will be loaded by master and children. Memory, CPU cost etc.
This isn't a con so much as an effect. Loading Express in the Haraka process is going to [normally] use fewer resources than spinning up another node process just for HTTP.  Express is quite minimal.

smfreegard:
> * HTTP listener has the same user/group as the SMTP listener and is able to access all files created by the Haraka
>  - permits a config plugin to modify / update config files and settings via HTTP interface
>  - permits a plugin like watch to collect real-time SMTP info and display via HTTP

So you're going to list a potential security issue as a feature?   We might feel differently when the first CVE is published for Haraka.

This isn't the UNIX way to do things: http://en.wikipedia.org/wiki/Privilege_separation

> - permits a config plugin to modify / update config files and settings via HTTP interface

This can be done without having the HTTP server in Haraka.   I'm *already* doing this here and my web app is a totally separate entity.   The changes I submitted for JSON, YAML loading and overrides allowed that.

> Express bugs could crash Haraka
> Sure, but that's little different than any other library Haraka uses. In the case of Express, it's extremely heavily tested, very minimal (since 4.x) and focused and in widespread use. I'd be more concerned with bugs in http serving plugins than with Express itself.

Sure - I'm simply pointing out the failure mode.  Other libraries that we require don't typically listen on public TCP ports though.

msimerson:
> So you're going to list a potential security issue as a feature?

Listening on a network port is a potential security issue, and it's also an opportunity to add a lot of useful features. Having a HTTP interface opens up a lot of future features, including:
    * config management interface
    * queue management
    * setup wizard, with help links

baudehlo:
My major thought on this is what do we lose by having this as an app outside of Haraka? Configs are reloaded automatically so they don't need in-memory access. Queue management can be done with child process calls etc. The only benefit is registering routes via hooks.

I sort of regret opening Pandora's box with this via the graph plugin and still think a separate http server might be better.

The flip side of course is you look at commercial email servers and they all have gui configuration so maybe we should embrace it. I'm very on the fence.

msimerson:
I'm not opposed to **also** having a completely separate haraka-httpd, and have been contemplating asking you to create that repo in the haraka org. That would be *better* for many purposes such as my plugin that shows you why I rejected your message, that needs to access to the rest of Haraka.

That would not work for my `watch` plugin. When we want to add SMTP AUTH enabled features in `haraka-httpd`, having some basic HTTP support baked-in to Haraka so we can add a couple API calls removed the need to duplicate functionality in `haraka-httpd`.

msimerson:
> maybe we should embrace it

I'm not even asking for an embrace, just tolerance. I wrote my watch plugin over a year ago. We talked about it again when #433 was still young and we all agreed that bundling a web server in a plugin was less than ideal. I waited many more months for the "blessed" implementation to arrive before I put in the time and effort to write this one.

We discussed. I listened to critiques and made changes. I implemented it off by default, acknowledging those who don't want a web server with their smtp server. This adds no dependencies to Haraka and any impacts affect only those who **choose** it. I waited more months.

I've added docs/Web.md.

smfreegard:
> The flip side of course is you look at commercial email servers and they all have gui configuration so maybe we should embrace it. I'm very on the fence.

Sure - but I'll bet my life savings that their functionality isn't implemented in the same daemon and same event loop or thread pool as the SMTP listeners.   The front-end and back-ends will be completely separate with API linkages between them, because experience tells us that is the best way to achieve it.

I'm not adverse to having some HTTP functionality or enablers in core; this simply doesn't *feel* like the right way to do it or indeed the right reasons to do it this way (e.g. a single plugin that requires this to work as it is currently written).

> I waited more months

Sorry - but that's how it is.  There's been PR's open for ages because people dumped the code on us and didn't make requested changes or because the code fell short in some way.   Submitting a PR doesn't require us to merge it if we don't agree on the changes that are being brought in.   There's no entitlement - and I wouldn't expect any if it was a PR of my own and both of you had concerns with it.   There's three of us with the commit bit here and two of them are raising concerns - it's up to *you* if you care about this so much to address those concerns and invest the necessary time to come up with something acceptable and build the necessary consensus.   If @baudehlo was comfortable to merge this code, then I'd have absolutely no issue with it being merged because there's a majority agreement on it.

Ok - that said; let's actually try and make this constructive and try and move this forward into something that we can all agree on...

Thanks for adding the docs; I can now actually see how it's supposed to work - the only bit that gives me some concern is only being able to have one plugin handling the websocket connection - how is that going to work given that any plugin could hook init_wss?  I'm guessing the first plugin to register on it is going to be used?

I've noticed one more issue with this PR; the way you have this right now - the Express app will be running as the user that *starts* Haraka, which is usually going to be *root* (as root is required to bind to port 25).   I think we can all agree that is not good.

Throwing some other ideas out there:

If the HTTP server functions were in a separate child_process that is started before the plugins are loaded; you could register the routes in the plugins by sending the necessary data to the child via the child_process.send('....') functions, on receiving these the HTTP server can add the necessary routes and/or load the plugin code itself to be able to use the code inside it.   That would provide the same functionality with proper separation between the processes.

For the websockets - the HTTP server could implement the necessary functions to handle the broadcasting of incoming data to all the connected clients and listen for the data on a UNIX domain socket which any plugin could then write to.   That would remove the current limitation of only allowing one plugin to work in this way although it would require some namespacing to prevent collisions.

baudehlo:
OK, I didn't review the code as thoroughly as I probably should, but at quick glance it looks fine, and the more I think about this the more it makes sense. I might even work on building a config editor plugin.

PR #916, opened 2015-04-08
fix 8-bit msg not displayed properly in gmail

msimerson:
reported by iksik on #haraka


transaction.js
```diff
@@ -130,14 +138,14 @@ Transaction.prototype.end_data = function(cb) {
         var data = this.body.parse_end();
         if (data.length) {
             data = data.replace(/^\./gm, '..').replace(/\r?\n/gm, '\r\n');
```
smfreegard: This should probably be:

`````
data = data.toString('binary').replace(/^\./gm, '..').replace(/\r?\n/gm, '\r\n');
`````

PR #934, opened 2015-04-18
connection: add an errors counter

msimerson:
that counts a bunch of errors that unrecognized doesn't

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -204,6 +204,7 @@ function Connection(client, server) {
     this.max_line_length = config.get('max_line_length') || 512;
     this.max_data_line_length = config.get('max_data_line_length') || 992;
     this.results = new ResultStore(this);
+    this.errors = 0;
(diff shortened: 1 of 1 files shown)


smfreegard:
Why don't you increment this counter on an unhandled unrecognized command as well?  Then there's no need to count these in the hook.

I'd also add this counter to the disconnect log line.

msimerson:
> Why don't you increment this counter on an unhandled unrecognized command as well?

Oversight, and I'm not sure how to identify that an unrecognized command was not handled. Where would I detect that?

smfreegard:
Adding it here: https://github.com/baudehlo/Haraka/blob/master/connection.js#L677 should do it.

Nice catch on this one - I'd forgotten that there were a bunch of placed that did this.

msimerson:
PR updated as requested, and squashed.

PR #940, opened 2015-04-25
watch SMTP connections in browser

msimerson:
* watch/client: display transactions in new rows
* watch/server: get results via get_all
    instead of checking based on list of plugins
* watch: prune connections > 100
* watch/server: don't sent data if same as last sent

### DIFF
--- a/config/watch.ini
+++ b/config/watch.ini
@@ -1,0 +1,12 @@
+; watch - a web interface for viewing Haraka activity
+
+; Sampling:Limit display to 1 connection per second (Default: false)
+; sampling=false
+
+[wss]
+; url (Default: same URL as HTTP client used)
+; The WebSocket client will attempt to connect via the same URI (changing only
+; the scheme) as the initial HTTP connection. WSS is stricter than typical
+; HTTP so the scheme and hostname *must* match else it silently fails.
+;
+; url=wss://mail.example.com/
--- a/docs/plugins/watch.md
+++ b/docs/plugins/watch.md
@@ -1,0 +1,33 @@
+# watch - smtp connections in HTTP browser
+
+Watch SMTP traffic to your Haraka server in a web interface. 
+
+![Watch Screen Capture](http://www.tnpi.net/internet/mail/haraka-watch.png)
+
+
+## Enable Watch
+
+1. Enable Haraka's HTTP server (see `listen` in http.ini)
+2. Add 'watch' to config/plugins
+3. Point your web browser at http://mail.example.com/watch/
+
+Enjoy the blinky lights. 
+
+
+## Tips
+
+* Hover your mouse pointer or tap (with touch devices) on table data to see more
+details. 
+* Copy that connection UUID at left and use it to grep your logs for even more.
+* Edit the files in watch/html and play with the appearance. If you make it
+  better, post a screen shot somewhere and create an Issue or PR.
+
+
+## Interpretation Key
+
+* Green: tests passed
+* Light Green: tests passed, but with conditions
+* Yellow: poor results, but not awful.
+* Light red: tests failed, but no rejection
+* Red: tests failed causing rejection
+
--- a/plugins/watch/html/client.js
+++ b/plugins/watch/html/client.js
@@ -1,0 +1,398 @@
+'use strict';
+/* jshint jquery: true, maxlen: 130 */
+/* global window, XMLHttpRequest, WebSocket */
+
+var ws, connect_cols, helo_cols, mail_from_cols, rcpt_to_cols, data_cols, total_cols;
+var cxn_cols, txn_cols;
+
+var connect_plugins  = ['connect.geoip','connect.p0f','connect.asn','dnsbl', 'early', 'connect.fcrdns'];
+var helo_plugins     = ['helo.checks','tls','auth','relay','spf'];
+var mail_from_plugins= ['spf','mail_from.is_resolvable'];
+var rcpt_to_plugins  = ['access','rcpt_to.in_host_list']; // ,'rcpt_to.qmail_deliverable'];
+var data_plugins     = ['bounce','data.headers','karma','spamassassin','clamd', 'data.uribl'];
+// 'seen' plugins are ones we've seen data reported for. When data from a new
+// plugin arrives, it gets added to one of the sections above and the table is
+// redrawn.
+var seen_plugins     = connect_plugins.concat(helo_plugins, mail_from_plugins,
+                       rcpt_to_plugins, data_plugins);
+var ignore_seen      = ['local_port', 'remote_host', 'helo', 'mail_from', 'rcpt_to', 'duration', 'queue'];
+
+var rows_showing = 0;
+
+function newRowConnectRow1 (data, uuid, txnId) {
+    var host = data.remote_host || { title: '', newval: ''};
+    var port = data.local_port ? (data.local_port.newval || '25') : '25';
+
+    if (txnId > 1) {
+        return [
+            '<tr class="'+uuid+'">',
+            '<td rowspan=2 class="uuid got uuid_tiny" title="">' + txnId + '</td>',
+            '<td rowspan=2 colspan="' + cxn_cols + '"></td>'
+        ];
+    }
+
+    return [
+    '<tr class="spacer"><td colspan="'+total_cols+'"></td></tr>',
+    '<tr class="'+uuid+'">',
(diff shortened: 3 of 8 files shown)


smfreegard:
Just about to review the code - but what immediately jumps out is that there's no documentation in this commit??

plugins/watch/public_html/client.js
```diff
@@ -0,0 +1,393 @@
...
+}
+
+function update_seen(plugin) {
+    if (seen_plugins.indexOf(plugin) !== -1) return;
+    if (ignore_seen.indexOf(plugin) !== -1) return;
+
+    $('#messages').append(', refresh('+plugin+') ');
+    seen_plugins.push(plugin);
+
+    var bits = plugin.split('.');
```
smfreegard: I can see a much better way to handle this.  I'll put it in the main comments.

smfreegard:
Ok - this looks OK to me, there's obviously a lot of hardcoding in this at the moment and plenty of room for improvement in the future.

Just had a quick look at plugins.js and I think this might be a useful change that could potentially reduce quite a bit of hardcoding that you have in this plugin:

`````
--- a/plugins.js
+++ b/plugins.js
@@ -12,6 +12,8 @@ var utils       = require('./utils');
 var util        = require('util');
 var states      = require('./connection').states;

+exports.registered_hooks = {};
+
 function Plugin(name) {
     this.name = name;
     this.base = {};
@@ -26,6 +28,13 @@ function Plugin(name) {
 }

 Plugin.prototype.register_hook = function(hook_name, method_name) {
+    if (!Array.isArray(exports.registered_hooks[hook_name])) {
+        exports.registered_hooks[hook_name] = [];
+    }
+    exports.registered_hooks[hook_name].push({
+        plugin: this.name,
+        method: method_name,
+    });
     this.hooks[hook_name] = this.hooks[hook_name] || [];
     this.hooks[hook_name].push(method_name);
`````

This would allow the plugins to see this variable as `this.registered_hooks` and would also allow me to remove some hardcoding to the plugin order option in bin/haraka too.

msimerson:
> there's no documentation

True, that will come, and it's going to be pretty short:

1. Enable http in Haraka (http.ini)
2. Add 'watch/server' to config/plugins
3. Point your web browser at http://my.haraka.com/watch/

>  I think this might be a useful change that could potentially reduce quite a bit of hardcoding

Sorry to not understand, but which hard coding would that remove? Beside perhaps the SPF plugin which looks at the scope to determine HELO vs MAIL FROM, I'm not seeing how having the plugin hook names would be a benefit.

>> update_seen

> I can see a much better way to handle this. I'll put it in the main comments.

One way to learn which plugins exist would be another AJAX call to watch/server.js that reads config/plugins and returns the list. The problem with doing that, versus having an assortment of plugin names hard coded and then dynamically redrawing the display table is making whatever comes back look good in the web interface. If they're being learned dynamically, there'd be an obvious desire to move the hard coded plugin names from client.js into watch.ini, so folks can choose the display ordering. Or make it deterministic based on order in config/plugins. Or...

smfreegard:
Ok - let me be a bit more clearer on all points.

Docs: all core plugins require them to be accepted for inclusion.  We don't have any plugins shipped in core without docs.   I'd have already committed this after I reviewed it if this were the case.  After all - how is anyone not already with Haraka and this plugin going to know what this is, why they would want it and how to set it up or indeed even know of its existence as it won't appear on the web site.

Regarding the code I supplied: I noticed for example that you're splitting the plugin name by dot to work out where it runs e.g. mail_from.is_resolvable which is redundant if you know which plugins run on which hook.

There also seemed to be other places in the code that having knowledge of this would also be useful as this isn't something you can work out by reading `config/plugins`.  This would give you the exact run order as determined by plugins.js at start-up which I thought would eliminate a whole bunch of guess work.

It was merely a suggestion - nothing more.  I'm going to add that code anyway as it reduces the amount of hardcoding of a function I added to `bin/haraka` and will automatically pick up new hooks without me having to remember to add them.

It's also just occurred to me that `watch` could run it's own hook to have plugins report their status to it.  That would also allow it to work out which plugins can report to it and in what order and would provide a way for other plugins to integrate with it without having to modify configuration files to do it.

msimerson:
> all core plugins require them

No surprises there.

> I'd have already committed this

That surprises me, and that's partly why there's no docs yet. There's almost zero options, and I don't expect everyone will have the same goals and expectations of the plugin that I do. I expected at least a few constructive comments that would shape the plugin more, and I'll write the docs when there's at least one other person who thinks the code is "good enough."  I'm happy to hear you think so.

> I noticed for example that you're splitting the plugin name by dot to work out where it runs e.g. mail_from.is_resolvable which is redundant if you know which plugins run on which hook.

Thank you for clarifying. Having another sharp mind thinking about it is helpful and I wasn't sure what you meant.

More specifically, I'm parsing the plugin names to work out where I should **display** them. That is only is only loosely correlated with where a plugin runs. For example, if I were to look at plugins that hook `mail`, I'd get this list:

````
access
bounce
data.headers
data.uribl
karma
mail_from.is_resolvable
queue/smtp_proxy
rcpt_to.host_list_base
rcpt_to.qmail_deliverable
record_envelope_addresses
spf
````

Of those, the only ones that have any particular affinity for `mail` are the mf.is_resolvable and spf. I thought about having overrides for plugins that "hook the world," like data.uribl, access, and karma, but that still doesn't whittle down the field enough. Nor does it control the aesthetics of placement and ordering in the display. And I surely don't want to add to the reasons for having plugin ordering "matter" in config/plugins.

> It's also just occurred to me that watch could run it's own hook to have plugins report their status to it.

Complicate every plugin for the convenience of watch? I thought avoiding that was the entire reason we went down the road to result_store (to complicate every plugin for the benefit of **several** other plugins (like elasticsearch)). 

> That would also allow it to work out which plugins can report to it and in what order and would provide a way for other plugins to integrate with it without having to modify configuration files.

The difficulty I see in being able to benefit from such an arrangement is deciding how best to display the data. Should access, uribl and karma have a box in every phase because they report a result there?  Or should they be aggregated into one box as they are now with a tooltip for more info? Which phase? There's a slew of similar issues like that boil down to opinions about how best to present the data. I expect some evolution of watch in this area.

Thanks again for you feedback. I'll some docs written up soon.

msimerson:
@smfreegard, docs added.

smfreegard:
> That surprises me, and that's partly why there's no docs yet. There's almost zero options, and I don't expect everyone will have the same goals and expectations of the plugin that I do. I expected at least a few constructive comments that would shape the plugin more, and I'll write the docs when there's at least one other person who thinks the code is "good enough." I'm happy to hear you think so.

You should have labelled this as 'Wants Review' and 'Work in Progress' is you didn't want it committed after review then...

> More specifically, I'm parsing the plugin names to work out where I should display them. That is only loosely correlated with where a plugin runs

Ok - but what a hacky way to determine this and why I was dead against (re-)naming all the plugins in this way.

> Complicate every plugin for the convenience of watch? I thought avoiding that was the entire reason we went down the road to result_store

We've already gone down this road, it's already a partially trodden path - the plugin naming being one step, result_store being another - and we went that route because trying to put that behaviour into *.notes would have caused all sorts of issues (outbound being the main issue, where txn.notes gets copied into hmail.notes).

Adding some redundant code into each plugin to provide optional metadata to another core plugin - I'd be fine with, as we wouldn't be adding the hook into core; it would be implemented inside your watch plugin.  It would be a far less pervasive change than results_store turned out to be (that particular patch-bomb was so big that I haven't bothered to merge my own plugins with core since, so improvements I've made locally haven't been contributed because I haven't had time to merge it and shake out all the bugs).  It also wouldn't be modifying the actual plugin code, just adding some hook_init_watch and hook_watch_result functions, that without the watch plugin being loaded, will never get called.

> The difficulty I see in being able to benefit from such an arrangement is deciding how best to display the data. Should access, uribl and karma have a box in every phase because they report a result there? Or should they be aggregated into one box as they are now with a tooltip for more info? Which phase? There's a slew of similar issues like that boil down to opinions about how best to present the data. I expect some evolution of watch in this area.

That's why I'm suggesting that you do it with metadata supplied by the actual plugin.  Surely you can come up with some sort of meta API to do that?  A init_watch hook called by watch at init_master would seem to be an obvious choice for that.   It could identify each plugin and which hooks they want to report on and where the data to report is stored etc.

> Thanks again for you feedback.

No problem - you're welcome.

PR #943, opened 2015-04-26
permit plugins to be directories

msimerson:
This change makes this possible:

```diff
@@config/plugins
-watch/server
+watch
```

````
ls plugins/watch
index.js
html
````

### DIFF
--- a/plugins.js
+++ b/plugins.js
@@ -19,6 +19,7 @@ function Plugin(name) {
     var full_paths = [];
     this._get_plugin_paths().forEach(function (pp) {
         full_paths.push(path.resolve(pp, name) + '.js');
+        full_paths.push(path.resolve(pp, name) + '/index.js');
     });
     this.full_paths = full_paths;
     this.config = config;
--- a/tests/plugins.js
+++ b/tests/plugins.js
@@ -80,7 +80,10 @@ exports.get_plugin_paths = {
         );
         test.deepEqual(
             this.plugin.full_paths,
-            [ path.join(__dirname, '../plugins/testPlugin.js') ],
+            [
+                path.join(__dirname, '../plugins/testPlugin.js'),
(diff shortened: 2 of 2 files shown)


baudehlo:
Should we do this the way node_modules works and read a package.json for the main module name?

smfreegard:
@baudehlo - yeah I think so.  But also shouldn't this also be looking in `<haraka_install_dir>/node_modules/<plugin_name>/{index.js/package.json}` instead of `<haraka_install_dir>/plugins/<plugin_name>/{index.js/package.json}`??

We ship a `<haraka_install_dir>/package.json` don't we, so if you put the plugins you want to install in that and do an `npm install`; the plugins are going to end up in `<haraka_install_dir>/node_modules` aren't they?   Or am I misunderstanding the purpose of that package.json?

I definitely think this a good improvement though.

baudehlo:
Yeah that is stage two.

msimerson:
I smell three different steps all leading in the same direction:

1. baby step: this PR, adding index.js support
2. one step: read and parse index/package.json and follow the main property.
3. big step: install plugins as npm packages

msimerson:
I think moving in that direction is **certainly** a good idea. We'd want a separate `haraka-plugin-test` harness that plugins in separate repos could import (at test time, perhaps, with a `git submodule`) and then run their tests under. 

With a nice web admin panel, we wouldn't need to ship any plugins with Haraka. Install Haraka, configure an admin user, point at https://my.mail.com/admin/ and then click checkboxes for which plugins you want to use. Under the covers, clicking the checkboxes does a:

````
cd $haraka_install; npm install $package
````

With a web ui, every plugin gets a nice name, description, travis build test results icon, and versioning. There's a whole lotta fun that can be had down that road.

baudehlo:
Agreed.

PR #944, opened 2015-04-27
Changing createCredentials to tls.createSecureContext

DarkSorrow:
I had a deprecation message when haraka start about createCredentials that told me to use the tls function instead.
Here is the doc: https://nodejs.org/api/crypto.html#crypto_crypto_createcredentials_details
It shows that it is the same options as well but i'm not sure about the bug mentioned in the file comment and how to exactly test it so if someone want me to make some more test with something specific let me know


msimerson:
Thanks for the PR. Since this passes tests on 0.10.38, I'm inclined to just merge it. But [I / we] haven't had a chance to test it in a 0.10 production environment first. That's all this is waiting for.

smfreegard:
Just tested it on 0.10 - appears to be fine.   Thanks for the patch!

PR #948, opened 2015-05-02
karma: prune features moved to limit plugin and

msimerson:
* remove concurrency checks
* remove max recipient checks
* add ability to select redis index
* remove penalty feature, more humorous than helpful, and doesn't fit future plans

### DIFF
--- a/config/karma.ini
+++ b/config/karma.ini
@@ -5,6 +5,7 @@
 [redis]
 server_ip   = 127.0.0.1
 server_port = 6379
+dbid        = 0
 
 ; karma IP history is expired after this many days. This value is refreshed
 ;       at every connection. Frequent senders may never expire.
@@ -40,10 +41,6 @@ positive=2
 ; until they have made > 5 connections and achieved a score lower than this.
 history_negative=-3
 
-; punish: the negative threshhold below which a connection must score to be
-; sent to the penalty box.
-punish=-15
-
 
 [deny]
 ; hooks to disconnect on. When the score is below [threshholds]negative,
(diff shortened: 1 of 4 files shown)


celesteking:
> remove penalty feature, more humorous than helpful, and doesn't fit future plans

What does that mean? I don't get the joke. The feature's been extremely helpful for us, it deters badly behaving hosts with very low # of false positives.

msimerson:
Not a joke, but I confess to having second thoughts.

### penalty feature
* Pros
    * really bad hosts don't waste resources try, try, trying again
    * really bad hosts get many fewer chances to send spam. This is the part I miss most, as I've noticed hosts that send about 15 total messages to my server over the course of a month (snowshoe spammers). Their scores range from -6 (barely passing) to -32. It's fairly obvious that anything they send is going to be spam. With `penalty` enabled, their odds of getting even that one  -6 message through are much, much smaller.
    * the taunt was humorous
* Cons
    * there's no "clear" button for penalty (gotta use redis CLI)
    * penalty occasionally nabs legit senders. Most recent example: a new server, fellow didn't have SPF configured correctly, IP space was dirty (previous tenant earned DNSBL listing), and there was some other issue too. Needless to say, he was sent to the penalty box. Without penalty, he gets to try again, after fixing SPF an then his message gets through.
    * the taunt was not helpful to humans inadvertently penalized.
    * a "very low # of false positives" is too many (when there's no way to easily "fix" it)

> doesn't fit future plans

* karma will focus on scoring
* `penalty` will be resurrected as its own plugin, using karma scoring as an input.
* the `penalty` plugin will have a CLI utility and HTTP interface (with JS and captcha anti-bot) that permits humans to clear their listing.
* the [Why Blocked](https://mail.theartfarm.com/haraka/) landing page replaces the taunt.

celesteking:
So, what happened with penalty feature? Was it added back?

msimerson:
Most of the penalty features have returned in the new [limit](https://github.com/haraka/Haraka/blob/master/plugins/limit.js) plugin. It handles limiting connection concurrency, max errors, max unrecognized commands, and max recipients, based on the connection quality, as scored by karma.

The *why blocked* feature has been implemented. [real live sample](https://mail.theartfarm.com/logs/87AFF565-7EDD-4FF9-AD93-2AA1391C6EF1) from today.

I haven't added a *penalty* plugin for a few reasons:

* the limit plugin provides 90% of the benefits while shedding 90% of problems (false positives).
* The only cost I can extract from naughty senders is one of their finite quantity of outbound TCP connections. By limiting them to a single very slow connection, I've greatly reduced my own exposure to them as well as making it take a little longer to hit the next victim.
* I **do** have the web server in place now, but I don't yet have an authentication mechanism in place that could permit **local** users to authenticate and clear a penalty listing.

celesteking:
speaking of limit plugin (haven't read docs on it), how is it different from rate_limit? Can I use both?

msimerson:
how different? read the docs. Can both be used? Yes. Should both be used? Probably.

In general terms:

`limit` imposes hard limits on things: concurrent connections, recipients-per-transaction, errors per connection, etc. 

`rate_limit` is more along the lines of "N connections per Y units of time" type limits. Both are useful.

celesteking:
Have read docs, I'm a bit perplexed of purpose of this plugin. 
For ex., max recipients limit, what's the point of returning tempfail if sender would just retry next time tempfailed recipients? Also, having it throw out error would definitely stop mail client from sending message, but won't stop spammers.

msimerson:
> ...but won't stop spammers.

There's no [Final Ultimate Solution for the Spam Problem](http://slashdot.org/comments.pl?sid=98024&cid=8373855) so pretending that something must stop spammers to be useful is based on a faulty premise. Instead, consider that perhaps this plugin does something implied in the name, such as reducing the attack surface and, oh I don't know, *limiting* spammers.

smfreegard:
It's more about preventing denial of service and saving resources than it is about preventing spam.  Although rate limiting your outbound traffic by the envfrom to say 500 recipients in 24 hours is a good way to spot compromised accounts quickly.   Particularly if you add alerting into the mix...

PR #950, opened 2015-05-03
connection: save queue results to result_store

msimerson:
### Goal
* make queue attempt failures visible to watch and elasticsearch plugins via result_store.

### change
* store queue result in transaction.results, providing more insight into what happened when queue attempts != 'ok'

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -1573,9 +1573,28 @@ Connection.prototype.queue_msg = function (retval, msg) {
     }
 };
 
+Connection.prototype.store_queue_result = function (retval, msg) {
+    var plugin = {name: 'queue'};
+    switch (retval) {
+        case constants.ok:
+            this.transaction.results.add(plugin, { pass: msg });
+            break;
+        case constants.deny:
+        case constants.denydisconnect:
+        case constants.denysoft:
+        case constants.denysoftdisconnect:
+            this.transaction.results.add(plugin, { fail: msg });
+            break;
+        default:
+            this.transaction.results.add(plugin, { msg: msg });
+            break;
+    }
+};
+
 Connection.prototype.queue_outbound_respond = function(retval, msg) {
     var self = this;
     if (!msg) msg = this.queue_msg(retval, msg);
+    this.store_queue_result(retval, msg);
     if (retval !== constants.ok) {
         this.lognotice('queue code=' + constants.translate(retval) + ' msg="' + msg + '"');
     }
@@ -1619,7 +1638,9 @@ Connection.prototype.queue_outbound_respond = function(retval, msg) {
                         if (!msg) msg = self.queue_msg(retval, msg);
                         self.respond(550, msg, function() {
                             self.msg_count.reject++;
-                            self.reset_transaction(function () { self.resume();});
+                            self.reset_transaction(function () {
+                                self.resume();
+                            });
                         });
                         break;
                     default:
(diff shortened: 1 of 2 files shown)


msimerson:
This is now cleaned up (noops merged separately) and ready for merge. Storing the results here is instead of #962, as this avoids the need to have each queue plugin save queue results.

Net result, is this bit of data in ES output:
```json
    "queue": {
      "pass": [
        "2.6.0 <03B521FEF618B64588A9F1AB6FCD34D0020F4347A3C2@IIG-EXCHANGE.*******.com> [InternalId=14493] Queued mail for delivery (CA08BB21-6DA2-43D6-A075-EE0010DAA991.1)"
      ]
    }
```

whereas doing it in the queue plugin (#962) adds this:
```json
    "queue/smtp_forward": {
      "pass": [
          "2.6.0 <03B521FEF618B64588A9F1AB6FCD34D0020F4347A3C2@IIG-EXCHANGE.********.com> [InternalId=14493] Queued mail for delivery"
      ]
    }
```

celesteking:
Hello dear. Why did you merge #962 if this one is the proper way of doing it?

smfreegard:
I agree - what's the point of merging this and #962??

msimerson:
> Why did you merge #962 if this one is the proper way of doing it?

Answer: timing.

* I proposed #962 first
* Then I started looking at adding the same thing to another queue plugin, and decided to do this PR instead.
* @smfreegard merged #962 
* I preferred this approach (works for all queue plugins) and merged this, making #962 redundant.

celesteking:
----
Next question is default outbound handling. That is, when all ```queue_outbound``` hooks return ```OK``` and mail delivery is handled by Haraka's ```outbound```.

My proposal is to log/index a separate document for each outbound transaction and *update* it with subsequent delivery attempts' statuses. I don't see any point of creating a separate document for each delivery attempt as it would contain basically all the same info, but with the additional status message and code returned from remote MTA (plus, maybe, new TTL of the mail). This is especially true when remote MTA is graylisting or is just down.

We just add status of queued message [to some list in that document](https://www.elastic.co/guide/en/elasticsearch/guide/master/partial-updates.html) and are done with it. Once message is accepted, we update the status field value to "accepted", for example. If not, ```{status: bounced}``` with additional field ```{bounce_reason: ("TTL_exceeded" | $remote_mta_rejection_msg)}``` added.

Not sure yet about "schema" of "outbound" document, need to work it out.

connection.js
```diff
@@ -1573,9 +1573,28 @@ Connection.prototype.queue_msg = function (retval, msg) {
...
+            this.transaction.results.add(plugin, { pass: msg });
+            break;
+        case constants.deny:
+        case constants.denydisconnect:
+        case constants.denysoft:
+        case constants.denysoftdisconnect:
+            this.transaction.results.add(plugin, { fail: msg });
+            break;
+        default:
+            this.transaction.results.add(plugin, { msg: msg });
```
celesteking: With ```CONT``` returned (Haraka handles outbound), ```queue``` results object will be logged with msg set to null.

smfreegard:
I already do this in my ES; you have to update status per-recipient and I keep the data in the main transaction.  It's not too hard - you have to create a groovy script to do it and store them in /etc/elasticsearch/scripts on each cluster node, the script is then called to upsert the value via the update API.

connection.js
smfreegard: That's not how Haraka works though.

hook_queue is only called when connection.relaying = false otherwise
hook_queue_outbound is called.

There must be a hook_queue plugin that calls OK otherwise the message is
rejected.

celesteking:
Oh, good point. Have totally forgotten remote MTA can reject per recipient, not just per message.

PR #951, opened 2015-05-03
Plugin ordering

smfreegard:
This commit adds support for overriding the plugin hook run order, by explicitly registering a hook and specifying a `priority` between -100 (highest priority) and 100 (lowest priority) with the default priority being 0 (zero).

Additionally it improves the `bin/haraka -o` output by showing the plugin priorities and now only shows hooks with plugins actually registered on them:

`````
Hook: connect ------------------------------------------------------------------
Plugin                                Method                                Prio
--------------------------------------------------------------------------------
access                                rdns_access                              0
access                                any                                      0
dnsbl                                 connect_first                            0
max_unrecognized_commands             hook_connect                             0
`````

If I then modify the dnsbl plugin to specify:

`````
plugin.register_hook('connect',  'connect_first', -100);
`````

The output then becomes:

`````
Hook: connect ------------------------------------------------------------------
Plugin                                Method                                Prio
--------------------------------------------------------------------------------
dnsbl                                 connect_first                         -100
access                                rdns_access                              0
access                                any                                      0
max_unrecognized_commands             hook_connect                             0
`````

Additionally - I found a small speed improvement in doing this, because previously plugins.run_hooks() went through the entire plugin list and checked each plugin in turn to see if it was to run on the current hook, this commit does this via an Object to do a keyed lookup instead and therefore saves quite a number of Array loops on a system with a lot of plugins (like mine!).

### DIFF
--- a/bin/haraka
+++ b/bin/haraka
@@ -372,11 +372,11 @@ else if (parsed.order) {
     for (var h = 0; h < hooks.length; h++) {
         var hook = hooks[h];
         console.log(sprintf('%\'--80s', 'Hook: ' + hook + ' '));
-        console.log(sprintf('%-37s %-37s %-4s', 'Plugin', 'Method', 'Prio'));
+        console.log(sprintf('%-35s %-35s %-4s %-3s', 'Plugin', 'Method', 'Prio', 'T/O'));
         console.log(sprintf("%'-80s",''));
         for (var p=0; p<plugins.registered_hooks[hook].length; p++) {
             var item = plugins.registered_hooks[hook][p];
-            console.log(sprintf('%-37s %-37s %4d', item.plugin, item.method, item.priority));
+            console.log(sprintf('%-35s %-35s %4d %3d', item.plugin, item.method, item.priority, item.timeout));
         }
         console.log('');
     }
--- a/bin/haraka
+++ b/bin/haraka
@@ -361,31 +361,22 @@ else if (parsed.order) {
                 (path.join(process.env.HARAKA, 'node_modules'));
         require('module')._initPaths(); // Horrible hack
     }
+    var sprintf = require('sprintf-js').sprintf;
     var logger = require(path.join(base, "logger"));
     if (!parsed.verbose)
         logger.log = function () {} // disable logging for this
     var plugins = require(path.join(base, "plugins"));
     plugins.load_plugins();
-    var hooks = ['init_master', 'init_child', 'lookup_rdns', 'connect', 'helo', 'ehlo', 
-                 'capabilities', 'unrecognized_command', 'mail', 'rcpt', 'rcpt_ok', 'data', 
-                 'max_data_exceeded', 'data_post', 'queue', 'queue_outbound','queue_ok', 
-                 'quit', 'disconnect', 'vrfy', 'noop', 'rset', 'reset_transaction', 'deny',
-                 'get_mx', 'bounce', 'delivered', 'send_email','deferred'];
     console.log('');
+    var hooks = Object.keys(plugins.registered_hooks);
     for (var h = 0; h < hooks.length; h++) {
         var hook = hooks[h];
-        console.log(hook + '\n----------------------');
-        for (var i = 0; i < plugins.plugin_list.length; i++) {
-            var plugin = plugins.plugin_list[i];
-            if (plugin && plugin.hooks[hook]) {
-                var j;
-                for (j = 0; j < plugin.hooks[hook].length; j++) {
-                    var hook_code_name = plugin.hooks[hook][j];
-                    (hook_code_name === 'hook_' + hook) ?
-                        console.log(plugin.name) :
-                        console.log(plugin.name + ' => ' + hook_code_name);
-                }
-            }
+        console.log(sprintf('%\'--80s', 'Hook: ' + hook + ' '));
+        console.log(sprintf('%-37s %-37s %-4s', 'Plugin', 'Method', 'Prio'));
+        console.log(sprintf("%'-80s",''));
+        for (var p=0; p<plugins.registered_hooks[hook].length; p++) {
+            var item = plugins.registered_hooks[hook][p];
+            console.log(sprintf('%-37s %-37s %4d', item.plugin, item.method, item.priority));
         }
         console.log('');
     }
--- a/docs/Plugins.md
+++ b/docs/Plugins.md
@@ -249,10 +249,23 @@ This is important as some plugins might rely on `results` or `notes` that have
 been set by plugins that need to run before them.   This should be noted in the
 plugins documentation though, so you should make sure that you read it.
 
-If you are writing complex plugins of your own, you may find that you have to 
-split your plugins into multiple plugins if you require your hooks to run in a 
-specific order e.g. you want hook_deny to run last after all other plugins and
-hook_lookup_rdns to run first, then you'd need two plugins to achieve this.
+If you are writing complex plugins of your own, you may find that you require 
+your hooks to run in a specific order e.g. you want hook_deny to run last after 
+all other plugins and hook_lookup_rdns to run first, then you can explicitly 
+register your hooks and provide a `priority` value which is an integer between
+-100 (highest priority) to 100 (lowest priority) which defaults to 0 (zero) if
+not supplied.  You can apply a priority to your hook in the following way:
+
+`````
+exports.register = function() {
+    var plugin = this;
+    plugin.register_hook('connect',  'hook_connect', -100);
+}
+`````
+
+This would ensure that your hook_connect function will run before any other
+plugins registered on the `connect` hook, regardless of the order it was 
+specified in `config/plugins`.
 
 You can check the order that the plugins will run on each hook by running:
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by msimerson** [comment]
- My first reaction is that this makes a complicated aspect of Haraka (plugin / hook ordering) more complicated. I lean towards ideas like adding "next(!CONT) not allowed here" hooks that [mostly] remove the ordering contention. I presume that's insufficient for your purposes.

- My second reaction is that this makes me want to replace a bunch of hook_[hook_name] function names with what_it_does names, so that `haraka -o` output is more meaningful. IE, instead of this:

````
lookup_rdns
----------------------
process_title
connect.p0f
connect.fcrdns
data.uribl
karma
limit                            incr_concurrency
````

we'd have this:

````
lookup_rdns
----------------------
process_title                   increment
connect.p0f                     get_tcp_fingerprint
connect.fcrdns                  check_fcrdns_match
data.uribl                      check_rdns_hostname
karma                           get_reputation
limit                           incr_concurrency
````

- Putting the plugin hooks into an object at register() to check at run time instead of iterating over an array listing every plugin seems an obvious win.

- Why limit the range of priority numbers? Wouldn't a 'lower numbers have higher priority" rule would be sufficient? I'm expecting that most Haraka installs will never use priority values until they run into an issue, and then they're only going to change the priority of a plugin they want to run sooner/later, and leave everything else at zero. Removing the limit avoids a "sorry, you have to modify plugin Y and increase its priority because that plugin author decided his plugin needed to run sooner than plugin Z and you disagree."

**Comment by msimerson** [comment]
My one other thought, and this might be worth no more than the electrons it's printed on, is that I wish that the plugin ordering was maintained somewhere besides in the plugin hooks. I'm thinking of the future when plugins can be npm packages, and the interface for enabling/disabling them is a web page, and the output of `haraka -o` too is a web page. It'd be nice if there was a way to influence the plugin / hook ordering in that interface with little up and down arrows or a drag-n-drop list.

**Comment by smfreegard** [comment]
> My first reaction is that this makes a complicated aspect of Haraka (plugin / hook ordering) more complicated. I lean towards ideas like adding "next(!CONT) not allowed here" hooks that [mostly] remove the ordering contention. I presume that's insufficient for your purposes.

I considered this - I was going to create a runtime flag that basically marked a plugin as 'always run' e.g. any plugin marked with the flag would always be run regardless of the earlier plugins return value.  I decided against it as it was much more complicated to code and still wouldn't fix the run order issue where you want one hook in your plugin to run before everything else and another hook to run after everything else.

I'm not keen on adding a boatload of hook_*_pre and hook_*_post hooks either - it's going to make things a lot more complicated as is making these hooks only allow specific return codes.

Limiting return types for specific hooks also isn't any less complicated IMO than this.   We'll have people turning up on IRC asking why when they `return next(DENY)` on them that it doesn't do as they intend (and it's more complicated to code).

> My second reaction is that this makes me want to replace a bunch of hook_[hook_name] function names with what_it_does names

And that would make Haraka better how?  helo.checks already does this, it's not really necessary elsewhere as most hooks do one thing and one thing well; adding method names to dnsbl would be pretty worthless, same with spamassassin or messagesniffer or data.uribl.   It's just needless lines added to the code for no gain.

> Putting the plugin hooks into an object at register() to check at run time instead of iterating over an array listing every plugin seems an obvious win.

*nod* - it's a reasonable win.  Even if a hook doesn't have any hooks registered it would still be iterating over the whole list, so it's definitely better.

> Why limit the range of priority numbers? Wouldn't a 'lower numbers have higher priority" rule would be sufficient? I'm expecting that most Haraka installs will never use priority values until they run into an issue, and then they're only going to change the priority of a plugin they want to run sooner/later, and leave everything else at zero. Removing the limit avoids a "sorry, you have to modify plugin Y and increase its priority because that plugin author decided his plugin needed to run sooner than plugin Z and you disagree."

Because there's always going to be an upper/lower limit e.g. it's an integer after all.   I decided it was easier to limit this to +-100 for simplicity.  I figured that we would allow the core plugins to go +-10 or +-50 or whatever, so it's still simple to make local plugins take precedence.   In reality I don't think this will be much of an issue given that this is going to be pretty rare to need to specify a priority.

> is that I wish that the plugin ordering was maintained somewhere other than the plugin register() call. I'm thinking of the future when plugins can be npm packages, and the interface for enabling/disabling them is a web page, and the output of haraka -o too is a web page. It'd be nice if there was a way to influence the plugin / hook ordering in that interface with little up and down arrows or a drag-n-drop list.

Wouldn't you do this by modifying `config/plugins` accordingly?  I'm not sure that having a priority for each plugin is the way to go for something like that.

Maybe making `config/plugins` a directory e.g. `config/plugins.d` and then dropping the files in lexical order and reading the directory instead as a method of loading plugins instead might be an idea or allowing the priority to be expressed in `config/plugins` somehow might be another way (or making `config/plugins` able to be read as `config/plugins.json` etc. etc. etc.

**Comment by msimerson** [comment]
> this makes me want to replace a bunch of hook_[hook_name] function names with what_it_does names

>> that would make Haraka better how? 

I showed you how in the output above. Quite a few Haraka users don't even know where their plugin files are installed, yet you continue to assume that everyone that uses haraka wants to go read the source to determine what a plugin does. For myself and many others, this output:

````
data_post
----------------------
data.uribl                      check_from_header
data.uribl                      check_reply_to_header
data.uribl                      check_message_id_header
data.uribl                      check_body_urls
````

is much more informative and useful than this:
````
data_post
----------------------
data.uribl
````

> It'd be nice if there was a way to influence the plugin / hook ordering
>> Wouldn't you do this by modifying config/plugins accordingly?

For plugins ordering sure, but that only goes so far with hook ordering.

>  I'm not sure that having a priority for each plugin is the way to go for something like that.

I misspoke:
```diff
- I wish that the plugin ordering was maintained somewhere other than the plugin register() call. 
+ I wish that the hook ordering was maintained somewhere other than the plugin register() call. 
```

I was thinking of hook priority. With a change like this PR, one needs to edit a plugins source file to change a hooks priority. If priority was specified in config/hook_priority, with simple file format for overriding the default priority of 0, it would be clearly within the control of the administrator.

smfreegard:
> I showed you how in the output above. Quite a few Haraka users don't even know where their plugin files are installed, yet you continue to assume that everyone that uses haraka wants to go read the source to determine what a plugin does. For myself and many others, this output:
> 
> data_post
> ----------------------
> data.uribl                      check_from_header
> data.uribl                      check_reply_to_header
> data.uribl                      check_message_id_header
> data.uribl                      check_body_urls
> is much more informative and useful than this:

> data_post
> ----------------------
> data.uribl

So you'd modify data.uribl to run 4 different methods just to make this output more expressive for people that can't be arsed to read the documentation to know that data.uribl "Parses URIs from a message an looks them up against blacklist sources".   You might as well add a long description to the plugin and show that instead...

 > It'd be nice if there was a way to influence the plugin / hook ordering

>> Wouldn't you do this by modifying config/plugins accordingly?
>For plugins ordering sure, but that only goes so far with hook ordering.

>> I'm not sure that having a priority for each plugin is the way to go for something like that.
> I misspoke:

> I wish that the hook ordering was maintained somewhere other than the plugin register() call. 
> I was thinking of hook priority. With a change like this PR, one needs to edit a plugins source file to change a hooks priority. If priority was specified in config/hook_priority, with simple file format for overriding the default priority of 0, it would be clearly within the control of the administrator.

Huh?  you can't change the priority of a *hook* - it's run based on the input of SMTP commands.  You can only influence the ordering of a plugin inside a given hook.

If you want to be able to express hook -> plugin -> priority in `config/priorities` or equivalent - you're welcome to add it in a future PR once this is merged.

smfreegard:
Just added a display of the plugin timeout to the `-o` option:

`````
Hook: connect ------------------------------------------------------------------
Plugin                              Method                              Prio T/O
--------------------------------------------------------------------------------
access                              rdns_access                            0  30
access                              any                                    0  30
dnsbl                               connect_first                          0  30
max_unrecognized_commands           hook_connect                           0  30
`````

@baudehlo: I'd appreciate a yay/nay or any comments when your visitors have gone in a couple of weeks.  This fixes all of the issues I've had with plugin ordering where I've had to split a single logical plugin into multiple pieces so that the hooks are run in the correct order.

msimerson:
> Just added a display of the plugin timeout

Nice. :+1:

msimerson:
> So you'd modify data.uribl to run 4 different methods just to make this output more expressive

The URIBL data_post hook is already running 4 different functions wrapped in a bigger one. This would be a reduction in function calls.

smfreegard:
>> So you'd modify data.uribl to run 4 different methods just to make this output more expressive
> The URIBL data_post hook is already running 4 different functions wrapped in a bigger one. This would be a reduction in function calls.

I don't really care about the function calls; I'm merely stating that data.uribl (and most other plugins) have a single purpose:  in data.uribl's case it is to look up any URIs in URI DNS blacklists, having it show each separate function with a pretty name, just for the sake of it isn't particularly useful, interesting or helpful.   A regular user of Haraka probably isn't ever going to run `-o` unless they have an issue and having data.uribl being shown 4 times instead of once in the output is just going to add confusion.   Additionally, you'd then have to potentially set 4 different priorities instead of one if you wanted to move things around - it simply doesn't make sense for data.uribl that has one task to do.

PR #962, opened 2015-05-09
smtp_forward: save to result_store

msimerson:
adds this bit of data to the ES results:
```json
    "queue/smtp_forward": {
      "pass": [
          "2.6.0 <03B521FEF618B64588A9F1AB6FCD34D0020F4347A3C2@IIG-EXCHANGE.********.com> [InternalId=14493] Queued mail for delivery"
      ]
    }
```

### DIFF
--- a/plugins/queue/smtp_forward.js
+++ b/plugins/queue/smtp_forward.js
@@ -54,6 +54,10 @@ exports.hook_queue = function (next, connection) {
 
         var dead_sender = function () {
             if (smtp_client.is_dead_sender(plugin, connection)) {
+                var rs = connection.transaction ?
+                         connection.transaction.results :
+                         connection.results;
+                rs.add(plugin, { err: 'dead sender' });
                 return true;
             }
             return false;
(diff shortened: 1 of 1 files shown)


smfreegard:
Dead sender???

It's a dropped connection.  We guard against this behaviour in all async plugins where the remote host could hang up between operations.

It's not really worthy of storing per-plugin IMO.   You can detect it in the connection close event.

msimerson:
I accept a message from a remote MTA and `smtp_forward` it to the destination MTA. Sometimes, while forwarding to dMTA (usually Exchange) it doesn't work out. What I'm after here is more visibility into why that forwarded connection failed. Did the Exchange server refuse the connection? Was there not enough smtp clients in the pool? Did the smtp_forward connection time out? Was there an SSL error? 

Nothing that would help answer those questions is currently making it into Elasticsearch, and saving more details to result_store seemed like a good start. 

> You can detect it in the connection close event.

Please explain further, I'm not understanding you.

msimerson:
adds this bit of data to the ES results:
```json
    "queue/smtp_forward": {
      "pass": [
          "2.6.0 <03B521FEF618B64588A9F1AB6FCD34D0020F4347A3C2@IIG-EXCHANGE.********.com> [InternalId=14493] Queued mail for delivery"
      ]
    }
```

In addition to #950, adds result_store error when a dead sender is encountered.

PR #964, opened 2015-05-09
Add Rspamd plugin

fatalbanana:


msimerson:
I have reviewed and tested this plugin and the version in this PR is working. There's one more issue that I've seen and that's that rspamd returns data structures in this format:

```json
{
    { name: A_SPF_RESULT, score: -1 }
}
```

Those nested data structures don't lend themselves well to logging not to nice indexable answer in Elasticsearch. I have some local changes I'll submit soon that address this.

smfreegard:
@msimerson

That format shouldn't be a problem for ES provided it's mapped as an object.  You would then simply reference it via parent_field.name and parent_field.score.   Kibana 3 didn't handle those formats though IIRC, haven't tried Kibana 4.

msimerson:
I've already got [an open PR](fatalbanana/haraka-rspamd#2) that makes it spit out useful log messages and ES results that don't require custom mappings. I expect that when @fatalbanana wakes up, he'll merge and we'll all be happy with this new plugin.

smfreegard:
Ok

fatalbanana:
I've added passing of sender/recipient/user/uuid information- it should be functionally-complete now.

smfreegard:
Thanks Andrew - I looked at Rspamd years ago and TBH I'd forgotten about it.   It looks much improved since then and I'm looking forward to trying it again now with your plugin.

msimerson:
I've been testing [rspamd](http://rspamd.com). I delayed rejecting spammy connections until data_post, so that most connections would be exposed to both spamassassin and rspamd. In general, if SpamAssassin would have detected it as spam, so too does rspamd. When SA is confident it's spam, so too is rspamd. 

When handling indeterminate mails in that are not clearly ham/spam, neither seems to have a particular advantage. Rspamd seems a fair bit more reliant, and scores more heavily on URIBL lookups. Once a domain gets listed (I've watched a few dozen [youreyesightabout | glucosehistorica l diabetescureforyou].us domains arrive on the URIBL as they're cycled through, before the spamming IP or it's URLs gets blacklisted. Of those *certain* spams, sometimes SA is better, sometimes rspamd is better. Of the messages I've inspected where rspamd suggested an action of `greylist`, about half were  spam and the rest were hammy transactional messages (examples: amctheatres.com, famousfootwear.com, and other small reputable businesses using an ESP).

Of the indeterminate messages that neither scanner properly detected as ham/spam, most (~75%) were blocked by karma, based upon neighbor (ASN) reputation. To me, this suggests that neither SA nor rspamd weigh ASN reputation as heavily as karma. The other big win for blocking those certain spams (~20%) is limiting the QoS for senders with poor IP or ASN reputation to a single (limit plugin) very slow (karma plugin progressive tarpits) connection. By making each of their attempts take 30-60 seconds to fail, the DNSBLs get time to update and future connections are much more likely to get detected by SA and rspamd (as well as DNSBL and URIBL plugins). 

The biggest difference between SA and rspamd is processing time. I added timers to both plugins, so that I could compare the processing time spent by both.The average time spent by SA sampled over 1000 messages is 1.4 seconds per message. With rspamd, the average is 0.3s per message. My volume isn't high enough on this server to effectively measure CPU usage but it seems apparent that rspamd would support much higher throughput.

The WebUI is also quite pretty.

rspamd webui

smfreegard:
Interesting stuff Matt.   I'm not going to be in a position to try rspamd in production for a week or two.  Given that I'm far more aggressive with up-front rejections and I have a very effective greylisting plugin that will run before the rspamd plugin - I suspect that my results will likely be worse, given that URIBL hits are rejectables here.

One thing that does interest me is the fuzzy algorithms as I have a 'training mode' in my Haraka implementation which currently trains DSPAM on stuff that I'm confident is spam (e.g. honeypot addresses, DNSBL + other markers), so it might be interesting to use this to generate fuzzy patterns and OSB bayes tokens in rspamd and see what effectiveness that provides.   The only problem is that the docs say you can use the HTTP interface to tell rspamd to learn the fuzzy matches from a message, but it doesn't actually say how to do that (it gives an example using rspamc binary).

msimerson:
> the docs say you can use the HTTP interface to tell rspamd to learn the fuzzy matches from a message

I noticed that too, and I'm asking the authors for details about the HTTP options that are passed to rspamd in the request. I have a lot more connection metadata that could be passed in, if it could be useful for training fuzzy and bayes.

However, that `Learning` tab you see in the image above? You can copy/paste messages in there for training, so it wouldn't be too hard to inspect that HTTP request and replicate it in your own training routines.

fatalbanana:
To train bayes & fuzzy post at the controller (post 11334):

/fuzzyadd and /fuzzydel with following headers:
Weight:
Flag: 
Password(?)

/learnham & learnspam
Password(?)

msimerson:
Nice API... thanks @fatalbanana!

vstakhov:
@msimerson , that's really a splendid comparison. I'm asking if I can use your results somewhere in rspamd documentation? And you have really moved ASN reputation implementation closer to the top of my TODO list :)

msimerson:
Certainly, use away.

PR #966, opened 2015-05-13
earlytalker: store results when relayclient talks early

msimerson:
Part of this PR is a question: Why would you put that line `if (!connection.early_talker) { return next(); }` inside the setTimeout?  It looks to me like a non-optimization.

This also adds a log message when a relaying client trips over the early_talker plugin. Yes, I've seen it happen 3 times now, so I'm sure it's happening more often. Just not quite sure how often.

### DIFF
--- a/plugins/early_talker.js
+++ b/plugins/early_talker.js
@@ -19,7 +19,13 @@ exports.load_config = function () {
 exports.early_talker = function(next, connection) {
     var plugin = this;
     if (!plugin.pause      ) { return next(); } // config set to 0
-    if (connection.relaying) { return next(); } // Don't pause AUTH/RELAY clients
+    if (connection.relaying) {   // Don't pause AUTH/RELAY clients
+        if (connection.early_talker) {
+            connection.results.add(plugin,
+                    { skip: 'an early talking relaying client?!'});
+        }
+        return next();
+    }
 
     setTimeout(function () {
         if (!connection.early_talker) { return next(); }


plugins/early_talker.js
```diff
@@ -18,11 +18,14 @@ exports.load_config = function () {
 
 exports.early_talker = function(next, connection) {
     var plugin = this;
-    if (!plugin.pause      ) { return next(); } // config set to 0
-    if (connection.relaying) { return next(); } // Don't pause AUTH/RELAY clients
+    if (!plugin.pause           ) { return next(); } // config set to 0
+    if (!connection.early_talker) { return next(); }
+    if ( connection.relaying    ) {   // Don't pause AUTH/RELAY clients
+        connection.lognotice('an early talking relaying client?!');
```
smfreegard: This shouldn't be a lognotice; logdebug is more appropriate.

smfreegard:
I'm going to reject this.   I think the 'return next()' inside the setTimeout() is what makes this plugin actually work.  e.g. it needs to wait on all clients to see if they are pipelining data when they shouldn't be and then returns next() for those that haven't.

plugins/early_talker.js
```diff
@@ -18,11 +18,14 @@ exports.load_config = function () {
 
 exports.early_talker = function(next, connection) {
     var plugin = this;
-    if (!plugin.pause      ) { return next(); } // config set to 0
-    if (connection.relaying) { return next(); } // Don't pause AUTH/RELAY clients
+    if (!plugin.pause           ) { return next(); } // config set to 0
+    if (!connection.early_talker) { return next(); }
+    if ( connection.relaying    ) {   // Don't pause AUTH/RELAY clients
```
smfreegard: If you do this; you'll got this logged for every relay client regardless as to whether or not they are earlytalkers or not.

msimerson:
>  I think the 'return next()' inside the setTimeout() is what makes this plugin actually work.

Ahhh, I see.

PR #976, opened 2015-05-18
Add optional return message to auth_base

DarkSorrow:
i had to distinguish between different reason on why the login was denied from my plugin. As smf suggested on irc, i added a message variable when i return the boolean for exemple: 
exports.check_plain_passwd = function (connection, user, passwd, cb) {
      if (user === 'me' && passwd === 'imAnoob') {
          return cb(true, 'you are logged');
      }
      return cb(false, 'get lost');
}
If no message is provided the default message will be shown

### DIFF
--- a/plugins/auth/auth_base.js
+++ b/plugins/auth/auth_base.js
@@ -83,12 +83,12 @@ exports.check_user = function (next, connection, credentials, method) {
         return;
     }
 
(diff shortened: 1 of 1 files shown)


smfreegard:
LGTM - thanks for the patch!

PR #982, opened 2015-05-26
make rate_limit redis keys expire whatever would happen

### LINKED ISSUE #980: rate_limit not atomic
It seems that sometimes rate_limit doesn't set TTL on newly created entries.

I've got a couple of cases of that and had to clean them up manually. Not having a TTL means user will certainly hit the rate limit. 

So, investigate the code and wrap it in a transaction.

Thanks.

celesteking:
Fix #980 : my proposal.

### DIFF
--- a/plugins/rate_limit.js
+++ b/plugins/rate_limit.js
@@ -150,25 +150,31 @@ exports.rate_limit = function (connection, key, value, cb) {
 
     connection.logdebug(self, 'key=' + key + ' limit=' + limit + ' ttl=' + ttl);
 
-    client.incr(key, function(err, val) {
+    client.get(key, function(err, val) {
         if (err) return cb(err);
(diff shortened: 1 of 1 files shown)


plugins/rate_limit.js
```diff
@@ -150,26 +150,30 @@ exports.rate_limit = function (connection, key, value, cb) {
...
+                return cb(null, true);
+            }
+            else {
+                // OK
+                return cb(null, false);
+            }
+
+        };
+
+        if (val == null) // new key
```
smfreegard: Missing {'s

celesteking:
okay, anything else? it works fine without {} there.

smfreegard:
It works - but it's not the house style: https://github.com/baudehlo/Haraka/wiki/Contributing#style-conventions

Otherwise - looks good.

celesteking:
Well here you go. :)

PR #984, opened 2015-05-27
custom message and status codes in pass check callback

celesteking:
Which is basically an enhanced #976.

Magically, it also makes this code possible:
```
cb(false, {message: DSN.no_such_user()});
```

### DIFF
--- a/plugins/auth/auth_base.js
+++ b/plugins/auth/auth_base.js
@@ -83,12 +83,11 @@ exports.check_user = function (next, connection, credentials, method) {
         return;
     }
 
-    // otps: {
-    //   message, code
-    // }
+    // valid: (true|false)
+    // opts: ({ message, code }|String)
     var passwd_ok = function (valid, opts) {
-        var status_code = (opts && opts['code']) || (valid ? 235 : 535);
-        var status_message = (opts && opts['message']) ||
+        var status_code = (typeof(opts) == 'object' && opts['code']) || (valid ? 235 : 535);
+        var status_message = (typeof(opts) == 'object' ? opts['message'] : opts) ||
                 (valid  ? '2.7.0 Authentication successful' : '5.7.8 Authentication failed');
 
         if (valid) {
--- a/plugins/auth/auth_base.js
+++ b/plugins/auth/auth_base.js
@@ -83,12 +83,19 @@ exports.check_user = function (next, connection, credentials, method) {
         return;
     }
 
-    var passwd_ok = function (valid, message) {
+    // otps: {
+    //   message, code
+    // }
+    var passwd_ok = function (valid, opts) {
+        var status_code = (opts && opts['code']) || (valid ? 235 : 535);
+        var status_message = (opts && opts['message']) ||
+                (valid  ? '2.7.0 Authentication successful' : '5.7.8 Authentication failed');
+
         if (valid) {
             connection.relaying = true;
             connection.results.add({name:'relay'}, {pass: 'auth'});
             connection.results.add(plugin, {pass: method});
-            connection.respond(235, ((message) ? message : "Authentication successful"), function () {
+            connection.respond(status_code, status_message, function () {
                 connection.authheader = "(authenticated bits=0)\n";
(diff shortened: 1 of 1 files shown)


plugins/auth/auth_base.js
```diff
@@ -83,12 +83,19 @@ exports.check_user = function (next, connection, credentials, method) {
         return;
     }
 
-    var passwd_ok = function (valid, message) {
+    // otps: {
```
msimerson: I'm thinking you meant, "opts"

msimerson:
Besides the comment typo, LGTM.

msimerson:
Note that this isn't backwards compatible with @DarkSorrow patch. As this PR would break his code, it'd be nice for either:
1. him to chime in here and bless the change (and update his code)
2. for this to check `typeof opts`, and use it as the message if it's a string

smfreegard:
I'd prefer option 2 as it's more how Haraka works internally with the DSN module and is more developer friendly e.g. pass a string or a DSN object and it works it out automatically.

In a future Haraka major release; we can change *all* of the error messages internally and in all of the plugins to use enhanced status codes and then advertise ENHANCEDSTATUSCODES as an ESMTP extension.

celesteking:
Here you go. (I don't know how to write tests yet)

plugins/auth/auth_base.js
```diff
@@ -83,12 +83,18 @@ exports.check_user = function (next, connection, credentials, method) {
         return;
     }
 
-    var passwd_ok = function (valid, message) {
+    // valid: (true|false)
+    // opts: ({ message, code }|String)
+    var passwd_ok = function (valid, opts) {
```
smfreegard: You should use === over == as it's stricter and we can also ensure that we've got a DSN object too, additionally if we aren't using the DSN module, we shouldn't return any enhanced codes.

I would do it this way:

`````javascript
// valid:  true/false
// message:  optional response string or DSN object
var passwd_ok = function (valid, message) {
    var code, reply;
    if (typeof(message) === 'object' && message.constructor.name === 'DSN') {
        code = message.code;
        reply = message.reply;
    }
    else {
        code = (valid ? 235 : 535);
        reply = (message ? message : (valid ? 'Authentication successful' : 
                                              'Authentication failed'));
    }

    ....
    connection.respond(code, reply, function () ....
`````

**Comment by smfreegard** [comment]
Crap - I've just realised.   This whole patch is unnecessary.
See https://github.com/baudehlo/Haraka/blob/master/connection.js#L463

I already handled this in the respond() function....

**Comment by celesteking** [comment]
I noticed that first, when trying to understand why the heck status code was skipped when DSN obj was supplied. 

I need to accept both message & code because DSN doesn't support x.7.8, and I found no option to craft custom extended codes there... So that's why I can do the following in auth plugin: 
```return cb(success, {message: '5.7.8 Login or password incorrect', code: 535});```
and it "just works"

**Comment by smfreegard** [comment]
It's pretty simple:

`````
var DSN = require('dsn');
...
return cb(success, return new DSN(535, 'Login or password incorrect', 535, 7, 8));
`````

The DSN module was the very first thing I wrote for Haraka, and it would seem that I didn't include RFC4594 codes to it; should be pretty simple to add though.

**Comment by celesteking** [comment]
no, it's not simple.

```
> var DSN =  require('Haraka/dsn');
> new DSN(535, 'Login or password incorrect', 535, 7, 8);
TypeError: object is not a function
```

smfreegard:
Ok - see https://github.com/baudehlo/Haraka/pull/985 which fixes that:

`````javascript
[smf@mail1-ec2 Haraka]$ node 
> var DSN = require('./dsn');
undefined
> new DSN.DSN(535, 'Login or password incorrect', 535, 7, 8);
{ code: 535,
  msg: 'Login or password incorrect',
  cls: 5,
  sub: 7,
  det: 8,
  default_msg: '',
  reply: '5.7.8 Login or password incorrect' }
`````

smfreegard:
Fixed in #991 closing.

celesteking:
this is work in progress actually i guess

PR #988, opened 2015-05-28
consolidate queue message default into function

msimerson:
replace 14 of these constructs:
```js
msg || "Message denied"
```
with 6 of these:
```js
if (!msg) msg = queue_msg(...)
```

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -637,7 +637,7 @@ Connection.prototype.resume = function () {
 
 Connection.prototype.lookup_rdns_respond = function (retval, msg) {
     var self = this;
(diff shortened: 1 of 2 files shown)


msimerson:
Added tests.

````
$ ./run_tests tests/connection.js 
Running tests:  [ 'tests/connection.js' ]

connection.js
✔ connection - get_capabilities
✔ connection - queue_msg, defined
✔ connection - queue_msg, default deny
✔ connection - queue_msg, default denysoft
✔ connection - queue_msg, default else

OK: 7 assertions (120ms)
````

PR #996, opened 2015-05-31
karma: simplifying further

msimerson:
* removed unused ipaddr.js
* rename connect to score (easier to understand what it is)
* in addition to computed history (good - bad), report good and bad
* be less verbose by default (when loglevel=info)
* don't tarpit quit (rarely works anyway)
* removed early_talker check
* added all_good and all_bad results
* on hook_deny, also permit DENYSOFTDISCONNECT to pass through
* moved static -1 penalty for env_user_match to karma.ini
* emit logs much less frequently
* added [asn]report_as feature

### DIFF
--- a/config/karma.ini
+++ b/config/karma.ini
@@ -15,6 +15,7 @@ expire_days = 60
 [asn]
 ; track karma for AS? (network neighborhood)
 enable=true
+;report_as=connect.asn
 
 
 [tarpit]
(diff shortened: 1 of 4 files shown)


smfreegard:
No docs for these changes??

msimerson:
Docs and tests are now updated.

Issue #947, opened 2015-05-02
add connect_pre hook to reduce plugin ordering issues


msimerson:
* lookup\_rdns - called to look up the rDNS - return the rDNS via `next(OK, rdns)`

There are 0 plugins that use this hook in this way.

There are a handful of plugins that use this hook in different ways: graph, karma, limit, process_title, profile, rate_limit, log. The most common use of this hook is to increment counters and initialize data structures. As such, any plugin that uses lookup_rdns in the documented way is likely to negatively impact that usage, unless careful attention to ordering in config/plugins is attended to.

Proposal:
* add a `connect_pre` hook, that always runs upon every connection, doing exactly what all the current plugins that use lookup_rdns expect.
* the connect_pre hook ignores plugin return codes

msimerson:
Here's a similar / related issue:

My `karma` plugin initializes its "to-score" list on the lookup_rdns hook. In general, I want karma to run **after** every other plugin on every hook, as karma is processing the results of those other plugins. On the lookup_rdns hook, `data.uribl` checks the rdns hostname and may attempt to reject the connection. When it does, `karma` will catch the DENY and permit the connection to continue (likely to reject later, after capturing some metadata). The mechanism available to override that DENY is `next(OK)`, which also has the undesired effect of skipping other plugins from processing on the lookup_rdns hook. Since `karma` runs after uribl, its init hook doesn't get called. Having rDNS listed on the URIBL then becomes a handy way to bypass karma.

This is not an issue that can be solved by rearranging the order of plugins on the hook.

### potential solutions
* have a connection setup and teardown hooks that plugins are always run on
* [create a runtime flag that marks a plugin as 'always run'](https://github.com/baudehlo/Haraka/pull/951)
* ...

### workarounds
* modify data.uribl and add reject=false options (like I did for many other plugins)

### hacks
* have karma try to init itself on every hook (bandaid, doesn't solve the problem)

smfreegard:
The most obvious answer and simplest solution (other than adding a bunch of extra hooks) is that you simply modify the hook_deny logic to continue running plugins after next(OK) is returned.

smfreegard:
I guess you could also add a different constant to all hook_deny to maintain the existing functionality too.  e.g. OKCONT for the new behaviour or something like that.

msimerson:
> The most obvious answer ... is ... modify the hook_deny logic to continue running plugins after next(OK) is returned.

[...]

PR #999, opened 2015-06-01
add connect_init hook

msimerson:
plugins: ignore retval for connect_init


msimerson:
This is an attempt to satisfy #947 

I'm not sure if this is the The Right Way to do it, but after much tinkering, it does work.

smfreegard:
LGTM

baudehlo:
+1 from me.

PR #1016, opened 2015-06-06
smtp_forward: revert part of #962 w/overlap #950

msimerson:
### changes:

1. Don't save txn results in queue/smtp_forward, redundant with #950
2. Don't append txn ID to queue message in queue/smtp_*
3. Do append txn ID in connection.queue_*_respond

Question: Is there any reason for the txn ID to be appended? Log
messages are already prefixed with the txn ID and txn results already have
the txn ID available, so at that point, it's entirely redundant data.

Near as I can tell, the only reason to append it there is so that it's
returned to the sending MTA in queue_*_respond. If that's the case, I'd rather
append it there.

Thoughts?

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -1571,7 +1571,7 @@ Connection.prototype.queue_msg = function (retval, msg) {
 
     switch (retval) {
         case constants.ok:
-            return 'Message Queued (' + this.transaction.uuid + ')';
+            return 'Message Queued';
         case constants.deny:
         case constants.denydisconnect:
             return 'Message denied';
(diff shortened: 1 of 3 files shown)


smfreegard:
> Don't append txn ID to queue message
> Question: Is there any reason for the txn ID to be appended? Log
messages are already prefixed with the txn ID and txn results already have
the txn ID available, so at that point, it's entirely redundant data.

> Near as I can tell, the only reason to append it there is so that it's
returned to the sending MTA in queue_*_respond. If that's the case, I'd rather
append it there.

Bingo - that's exactly why it was added so that the sending MTA gets the UUID and definitely agree; it's better off added in:

`````
Connection.prototype.queue_msg = function (retval, msg) {
    if (msg) return msg + ' (' + this.transaction.uuid + ')';
`````

**Comment by msimerson** [comment]
Hmm, appending it in `queue_msg()` would be bad for two reasons:

1. queue_msg is conditional, so frequently it wouldn't get appended
2. it would still be appended in the result_store results, where it's redundant.

Instead, I've removed it from the uuid appending from the places it was previous added, and now unconditionally append it immediately after saving to result_store. 

```json
"queue": {
  "pass": [
    "2.6.0 <7628E5543F3B4F248372CBB0D417B3C8@smtpserve.com> [InternalId=23694] Queued mail for delivery"
  ]
}
```

No redundant UUID appendage. :-)

Issue #945, opened 2015-04-27
Save the SMTP conversation


msimerson:
I think I'd like to store a summary of the SMTP conversation, assembled as an array on the connection, something like this:
````
<-  220 mail.theartfarm.com ESMTP Haraka 2.6.2-msimerson ready
 -> EHLO rmbp.local
<-  250-mail.theartfarm.com Hello wsip-70-168-115-6.ks.ks.cox.net [70.168.115.6], Haraka is at your service.
<-  250-PIPELINING
<-  250-8BITMIME
<-  250-SIZE 35000000
<-  250 STARTTLS
 -> STARTTLS
<-  220 Go ahead.
 ~> EHLO rmbp.local
<~  250-mail.theartfarm.com Hello wsip-70-168-115-6.ks.ks.cox.net [70.168.115.6], Haraka is at your service.
<~  250-PIPELINING
<~  250-8BITMIME
<~  250-SIZE 35000000
<~  250 AUTH PLAIN LOGIN CRAM-MD5
 ~> AUTH CRAM-MD5
<~  334 *********
 ~> ******************==
<~  235 Authentication successful
 ~> MAIL FROM:<matt@tnpi.net>
<~  250 sender <matt@tnpi.net> OK
 ~> RCPT TO:<matt@tnpi.net>
<~  250 recipient <matt@tnpi.net> OK
 ~> DATA
<~  354 go ahead, make my day
<~  250 ok 1430153916 qp 82679 (1A11ED1E-51CF-4512-B1B2-BF8CB1179CC6.1)
 ~> QUIT
<~  221 mail.theartfarm.com closing connection. Have a jolly good day.
````

such that it could be assembled into a string like so `connection.conv.join('\n');`

Having this in ES would be quite helpful for support purposes.

Thoughts?

baudehlo:
Core or plugin? It should be easy to do in a plugin. There's
connection.original_string.

smfreegard:
You probably want to store the pre and post transaction conversation in cxn.notes and the transaction stuff in txn.notes and then merge the two before you write the transaction data to ES otherwise if you get >1 transaction per connection, you'll end up with a potentially large structure that contains all of the transactions which I suspect you probably don't want.

msimerson:
> Core or plugin?

I was thinking core.

 > connection.original_string

[...]

PR #1021, opened 2015-06-12
connection: add MAIL and RCPT to results

msimerson:
* stores all sender and recipient attempts in results_store
* move rcpt incrementing into rcpt_incr, which also saves to results

replaces 10 of these:

    this.transaction.rcpt_count[action]++;
    this.rcpt_count[action]++;

with 10 of this:

    this.rcpt_incr('accept');  // or 'reject' or 'tempfail'

This change drains some impetus for #945

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -915,36 +915,68 @@ Connection.prototype.mail_respond = function(retval, msg) {
         'code=' + constants.translate(retval),
         'msg="' + (msg || '') + '"',
     ].join(' '));
+
+    function store_results (action) {
+        var res = {
+            action: action,
+            sender: sender.format(),
+            code: constants.translate(retval),
+        };
+        if (msg && action !== 'accept') res.msg = msg;
+        self.transaction.results.add({name: 'mail_from'}, res);
+    }
+
     switch (retval) {
         case constants.deny:
             this.respond(550, msg || dmsg + " denied", function() {
+                store_results('reject');
                 self.reset_transaction();
             });
             break;
         case constants.denydisconnect:
             this.respond(550, msg || dmsg + " denied", function() {
+                store_results('reject');
                 self.disconnect();
             });
             break;
         case constants.denysoft:
             this.respond(450, msg || dmsg + " denied", function() {
+                store_results('tempfail');
                 self.reset_transaction();
             });
             break;
         case constants.denysoftdisconnect:
             this.respond(450, msg || dmsg + " denied", function() {
+                store_results('tempfail');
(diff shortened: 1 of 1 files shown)


smfreegard:
Sorry - I don't like this.

If you're scraping logs or using something like Splunk (probably Logstash too), then having a log format this is the same for every instance of that log line is desirable and this changes it completely depending on conditions.

Quite often I do things like this:

`````
$ grep ' recipient ' /var/log/maillog | grep -Po 'code=\S+' | sort | uniq -c | sort -rn
   7359 code=DENY
   4682 code=OK
   1027 code=CONT
    303 code=DENYSOFT
     42 code=DENYSOFTDISCONNECT
`````

That becomes impossible after these changes.

I also don't like it logging under `[mail_from]` and `[rcpt_to]` .vs. `[core]`.  Currently you can grep your logs for only core output easily and discard the plugin output and this change will break any existing parsers.

You're also not quoting strings so; `msg=MX without an A record` is going to be far harder to capture than msg="MX without an A record" e.g. via `msg="([^"]+)"`

You appear to have found a bug somewhere as `  1027 code=CONT` should be impossible on my install and indeed looking at the logs; somehow the code returned by core was CONT when in actual fact the plugin returned OK - I'm looking in to why now.

Logging changes should be treated like API changes and things that fundamentally change the format should only be done on a major version bump, documented and warned about accordingly.

Adding a new `disposition=[accepted|tempfailed[+drop]|rejected[+drop]]` field on the end of the existing sender, recipient and message log lines would be a fine compromise though to make it clearer for those that don't understand what the `code=constant` means whilst retaining the existing format and functionality.

msimerson:
My main concern is getting the data into ResultStore so that it ends up in Elasticsearch. Perhaps another solution is to leave the existing lognotice lines in place and turn off the `emit: true` options to result_store.

smfreegard:
As long as the logging remains as it is and we still have the existing counters; then I've got no problem at all with adding this stuff to ResultStore as it seems like a pretty sensible thing to do.

I've found the cause of the `code=CONT` being logged incorrectly; it's because `hook_rcpt` returned `OK` at which point the log output is generated by `rcpt_ok_respond()` instead, so in this case if `hook_rcpt_ok` returns `CONT`; it should actually be logged as `OK` instead.

msimerson:
Is removing the empty msg="" strings okay?

smfreegard:
No - sorry.

msimerson:
Log message affecting changes have been removed.

PR #1023, opened 2015-06-12
attach: consolidate config handling into load_ini()

msimerson:
Now the extension loading get gets run once at register() and once per time
the config file changes.

Previously, the config wrangling was done on every connection.


msimerson:
Updated:

* saves results to result_store
* calculates and also stores attachment size

PR #1026, opened 2015-06-14
config: make ini values optional

msimerson:
I'd like the INI key to be permissible without an = character. Currently, this
works:

    key=

But this does not:

    key

It seems not-very-intuitive to require that = sign to be present.

### use case

```ini
[hosts]
es1.nyc.example.com
es2.nyc.example.com
es1.lax.example.com=protocol:https, dc:lax
es2.lax.example.com=protocol:https, dc:lax
```

There's also quite a few places where we shove data off to separate files to get `list` loading ability, and this permits storing that data in an ini [section].

### DIFF
--- a/configfile.js
+++ b/configfile.js
@@ -8,7 +8,7 @@ var yaml = require('js-yaml');
 // for "ini" type files
 var regex = exports.regex = {
     section:        /^\s*\[\s*([^\]]*?)\s*\]\s*$/,
-    param:          /^\s*([\w@\._\-\/]+)\s*=\s*(.*?)\s*$/,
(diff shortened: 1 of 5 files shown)


smfreegard:
I've wanted to do this before as I much prefer using .ini files over list types for index style lookups because of the obvious speed advantage in doing so and this obviously makes that possible.

I'd resisted changing it like this though as we're drifting further away from the standard.

So whilst I think this is a good change; it 1) needs documenting and 2) I think it needs @baudehlo to approve.

baudehlo:
+1

msimerson:
docs added.

PR #1027, opened 2015-06-14
attach: load regex files early and pre-compile

msimerson:
* them in advance (rather than upon every connection)
* add periods to file extensions when not present (ie, no longer required in config)
* fixed a doc-code mismatch
* store archive extensions in object instead of array
    * replace list.indexOf(thing) === -1 with !list[thing]
* added disallowed_extensions config option, which blocks attachments with
  those extensions in file attachments and archives.
* lots of delicious tests

### DIFF
--- a/docs/plugins/attachment.md
+++ b/docs/plugins/attachment.md
@@ -36,23 +36,28 @@ Configuration
 -------------
 
 * attachment.ini
+    * default settings shown
 
-  - archive\_max\_depth
-    (default: 5)
+  - timeout=30
+
+    Timeout in seconds before the plugin will abort.
+
+  - disallowed_extensions=exe,com,pif,bat,scr,vbs,cmd,cpl,dll
+
+    File extensions that should be rejected when detected.
+
+    [archive]
+  - max\_depth=5
 
     The maximum level of nested archives that will be unpacked.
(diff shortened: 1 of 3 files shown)


smfreegard:
This has crashed into a change I'd already made locally - I'd basically moved the regexp files into the INI file as a wildcard list.

I'll look at merging the two, but it doesn't really make much sense to have a disallowed_extensions list to go with it.

msimerson:
> I'd basically moved the regexp files into the INI file as a wildcard list.

That's an improvement.

Another would be having a section like this:

```ini
[blocked_extensions] 
exe
dll
scr
```

and 

```ini
[archive_extensions]
zip
tar
tgz
taz
z
gz
rar
7z
```

now that PR #1026 has landed.  The options_to_array() can go away entirely.

smfreegard:
> now that PR #1026 has landed. The options_to_array() can go away entirely.

No - it can't.   I use it elsewhere and it's still useful for other things where this type of format isn't suitable.

> [blocked_extensions] 
> exe
> dll
> scr

This is a poor way to do it.   It prevents you from being able to block specific files by name using the same list .vs. :

file_extensions=*.exe *.dll *.scr winmail.dat
archive_extensions=:file_extensions: message.dat

etc.

PR #1028, opened 2015-06-14
watch & graph updates for compatibility with #1015

msimerson:
watch and graph plugins


baudehlo:
Not that relevant to this commit, but should we deprecate graph? I created it because it was cute but I don't use it any more and I doubt anyone else does either.

msimerson:
> should we deprecate graph?

I'd say create an issue for it with a 3.0 milestone, and add it to the TODO list. If anyone sees value in it, they'll have plenty of time to speak up before the sunset date.

PR #1030, opened 2015-06-14
deprecate graph plugin

msimerson:
Per @baudehlo comment in PR #1028


baudehlo:
If plugins as npm modules work, we should put this on npm.

msimerson:
I haven't had time to get it installed and write install instructions, but it's got a new home at haraka/haraka-graph and it's [published on npm](https://www.npmjs.com/package/haraka-plugin-graph)

PR #1033, opened 2015-06-15
Allow hook_rcpt to return OK with message

smfreegard:
This fixes the behaviour when hook_rcpt returns `next(OK, 'message here')`.   Currently the message is ignored as hook_rcpt_ok takes the message sent by any plugins that it ran instead.   This patch applies any message from hook_rcpt where the message is unset in hook_rcpt_ok.

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -211,6 +211,7 @@ function Connection(client, server) {
     this.max_data_line_length = config.get('max_data_line_length') || 992;
     this.results = new ResultStore(this);
     this.errors = 0;
+    this.last_rcpt_msg = null;
     setupClient(this);
 }
 
@@ -985,6 +986,7 @@ Connection.prototype.rcpt_ok_respond = function (retval, msg) {
         this.results.add(this, {err: 'rcpt_ok_respond found no transaction'});
         return;
     }
+    if (!msg) msg = this.last_rcpt_msg;
     var rcpt = this.transaction.rcpt_to[this.transaction.rcpt_to.length - 1];
     var dmsg = "recipient " + rcpt.format();
     // Log OK instead of CONT as this hook only runs if hook_rcpt returns OK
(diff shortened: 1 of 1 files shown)


smfreegard:
BTW - my reason for patching this is because I was rewriting recipients at hook_rcpt and because of the way the logging works; the response sent to the client 'leaks' the rewritten recipient, so to prevent that I returned `next(OK, 'Recipient OK')` which of course didn't work ... ;-)

connection.js
```diff
@@ -985,6 +986,7 @@ Connection.prototype.rcpt_ok_respond = function (retval, msg) {
         this.results.add(this, {err: 'rcpt_ok_respond found no transaction'});
         return;
     }
+    if (!msg) msg = this.last_rcpt_msg;
     var rcpt = this.transaction.rcpt_to[this.transaction.rcpt_to.length - 1];
```
msimerson: curiosity question:
Couldn't the first change, setting `last_rcpt_msg=null` be omitted? 

And likewise, wouldn't this be better?

```diff
- if (!msg) msg = this.last_rcpt_msg;
+ if (!msg && this.last_rcpt_msg) msg = this.last_rcpt_msg;

- this.last_rcpt_msg = msg;
+ if (msg) this.last_rcpt_msg = msg;
```

smfreegard: No - this is a V8 optimization.  Over the life of the Haraka process we create thousands of connection objects and creating each object with identical properties prevents V8 from creating a hidden class.  See https://developers.google.com/v8/design

There's also a paper from a talk about this about v8 optimizations but I can't find it right now.

I also think this is quite developer friendly because one look at the constructor tells you all the properties it might have.

smfreegard: Here you go, now I'm actually at my computer instead of using my phone...

This is worth a read: http://www.html5rocks.com/en/tutorials/speed/v8/ - the references to the original talk and slides are at the bottom, but this is a good summary of everything discussed in it.

msimerson: That's a very good read, thanks!

PR #1036, opened 2015-06-16
Incorrect plugin name

abhas:
I had problems with a missing plugin while trying to rcpt_to.ldap -- looks like this inherits line should include the full name of the plugin or else it doesn't work.


msimerson:
thanks @abhas !

PR #1042, opened 2015-06-17
Add object.hook property and don't run deny on init_* hooks

smfreegard:


msimerson:
> Add object.hook property and don't run deny on init_ hooks

+1 and +1

> Remove hook property from transaction (not needed)

That commit seems to have gotten removed from this PR by 106163c. If that was purposeful, squashing away that commit message would be a good idea.

PR #1051, opened 2015-06-27
Do something about errors from Rspamd

fatalbanana:


plugins/rspamd.js
```diff
@@ -159,6 +159,13 @@ exports.parse_response = function (rawData, connection) {
         return;
     }
 
+    if (Object.keys(data).length === 1 && data.error) {
+        connection.transaction.results.add(plugin, {
+            err: 'Rspamd returned error: ' + data.error
+        });
+        return;
```
msimerson: Why have the 'Rspamd returned error:' prefix?

fatalbanana: Just to distinguish it as some particular type of error. :)

fatalbanana: I guess it is not strictly necessary in this context. :)

plugins/rspamd.js
```diff
@@ -159,6 +159,13 @@ exports.parse_response = function (rawData, connection) {
         return;
     }
 
+    if (Object.keys(data).length === 1 && data.error) {
+        connection.transaction.results.add(plugin, {
```
fatalbanana: Would it be preferred to send errors to connection.logerror or connection.transaction.results.add (or both)? Currently the plugin has some mixed behaviour in this respect.

msimerson: Errors stored to results are automatically emitted as log error, so that is preferable.

PR #1057, opened 2015-07-06
Fix early_talker to work at connect_init as well as data

smfreegard:


baudehlo:
Lacks docs and tests (not sure it will be easy to test though)

smfreegard:
Huh?  It's removing stuff from core that wasn't documented (and was dubious)?

baudehlo:
Yes but we expect plugins to have docs.

smfreegard:
Sorry - I misunderstood; I've updated the plugin docs to correctly match reality.

baudehlo:
I feel bad for commenting now. I forgot this was already a plugin - I thought it was new. Mea culpa.

PR #1066, opened 2015-07-10
Bumped version of node-syslog to fix build error on node v0.12

cbestall:
Bump to 1.2.0 for node-syslog to prevent error when installing using node v0.12.


msimerson:
backwards compat tested on node 0.10, looks good.

PR #1073, opened 2015-07-12
dnsbl: improve reported result error

msimerson:
this is another of those, "now that I can **see** those results in Elasticsearch, lets make 'em better."


baudehlo:
Is there any benefit to supporting errors as objects instead of just putting the message in there? I figure err.code might be useful, no?

msimerson:
not to me, and not really in this particular context. (Everything else logs strings to err, so this is an anomaly. When we run into a case where having err.code is useful, that should likely be saved separately, or perhaps even as an errObj list in results).

baudehlo:
Ok.

+1

PR #1074, opened 2015-07-12
limit: pass default when history missing

msimerson:
and also save pass result


baudehlo:
What's the value in removing the debug logging here?

msimerson:
It's not removed, that comes along for free with the results.add line.

baudehlo:
Then +1

Issue #1065, opened 2015-07-09
Release 2.7.0


baudehlo:
Ticket for discussion on releasing 2.7.0.

Discuss any open issues we need to get fixed before release here.

I think (gut feel) we are very close to being able to cut a release. Thoughts?

msimerson:
I have a very significant update to karma that's mostly done. I've changed the config file format of the rules (all new methods, old config settings still work) such that rules are simpler, can be read, written, scored, and then [reported on](https://mail.theartfarm.com/haraka/logs/6E9A2F70-2CD7-42DE-A09D-323A292DC3D2) posthumously.

msimerson:
Oh, and it uses the new redis pub/sub support in Results, so at some future date when I choose to longer support the old config settings, karma can stop "hooking the world"

smfreegard:
Sounds fine to me; I don't have anything pressing to go in.

baudehlo:
ETA on karma changes?

msimerson:
Code is all done, config file is updated, I just need to finish updating the docs. Highly likely this weekend.

msimerson:
The karma changes are in, a few other bug fixes have merged in as well since then. I think we should release, and then look forward to 2.7.1.

smfreegard:
#1081 must be merged prior to release.   If not - then a lot of potential breakage will occur as the plugin ordering in `config/plugins` is not honoured.

#1075 should also be merged; I've commented on the change from milliseconds to seconds.

I'd also like to see #952 go in.  The tests are broken, not the PR.

The rest are all pretty much dead in the water.

msimerson:
Ah yes, I'll try to get them tested and merged in the next few days.

dontforget:
Hello guys!

When we can expece v2.7 release date?

Thanks.

baudehlo:
We will probably push out RC1 on Monday.

soundslocke:
Exciting!

Dexus:
Please merge #1118 before release v2.7

It has many improvments, fixes and makes compatible to each iojs (that V8 build with g++ >= 4.8).

PR #1075, opened 2015-07-12
Early talker

msimerson:
contains @smfreegard 's #1057 squashed:

* set early_talker to delay at connect_init and data …
* Remove early_talker delay in core

and a few of my changes:

* added reject option
* save pass/fail results
* move config to early_talker.ini
    * with fallback to early_talker.pause
* added reject option (default: true)
* only calculate elapsed when hook=connect_init
* when pausing, subtract elapsed from pause
* make sure connection exists after delay (else a CRIT happens)
* brought docs the rest of the way up-to-date
* renamed the test file to match the plugin name

Closes #1057


config/early_talker.ini
```diff
@@ -0,0 +1,6 @@
+
+; delay in seconds
+pause=5
```
smfreegard: I don't think this should be in seconds.   My default delay here is 500ms...

baudehlo:
There's a comment in #1075 about a change to milliseconds. Is this a block on merging this?

baudehlo:
Sorry, #1065 I mean.

msimerson:
@smfreegard thinks so. I think seconds is far more user-friendly, but I understand Steve's desire to use portions of seconds. Even after this change, he can continue to use milliseconds in `config/early_talker.pause`.

baudehlo:
Not reading the code right now so two questions:

1) is that a breaking change?

2) does it support decimal seconds?

If it supports decimal seconds and doesn't break things then it seems fine.

msimerson:
Non-breaking.

Yes, it certainly supports decimal seconds.

````
 $ node 
> 0.5 * 1000
500
> '0.5' * 1000
500
````

msimerson:
Having thought more about this, I feel that values in config files really **ought**  to be seconds. 

Milliseconds are still the default in `early_talker.pause` and are also available as decimal values in the `.ini` file.

PR #1081, opened 2015-07-17
Fix plugin ordering

smfreegard:
Just realised that my plugin ordering changes in https://github.com/baudehlo/Haraka/commit/24d15b6006789b630392eac4343ba5100ec25483 incorrectly cased the object ordering to change even when no priorities are set.

This patch fixes that and retains the original order as per the ordering in `config/plugins` when no priorities are set.

### DIFF
--- a/plugins.js
+++ b/plugins.js
@@ -15,6 +15,7 @@ var states      = require('./connection').states;
 exports.registered_hooks = {};
 exports.registered_plugins = {};
 exports.plugin_list = [];
+var order = 0;
 
 function Plugin(name) {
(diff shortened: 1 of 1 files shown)


msimerson:
Seems like an easy thing to add test coverage for.

smfreegard:
Easy for you maybe, but I can assure you it would take me far longer to work out how to test this than I did to find the issue and fix it.

baudehlo:
Since this is a 2.7 blocker I'm going to merge and add a ticket for tests.

PR #1090, opened 2015-07-22
Move data_post hook timer into transaction object

smfreegard:
The delay between the ending dot and when the message is accepted is a key statistic as most senders do not honour the RFC timer of 5 minutes for this phase, typically most senders timeouts are much shorter, so the delay must be kept to a minimum.

I've moved the recording and storage of this to the transaction object as it seemed the more logical place to keep it and so that this can be used later by a plugin e.g. one that logs the delay to a database.

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -204,7 +204,6 @@ function Connection(client, server) {
         tempfail: 0,
         reject:   0,
     };
-    this.data_post_start = null;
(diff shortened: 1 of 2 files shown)


msimerson:
+1 for moving a transaction oriented timer into transaction.js.

PR #1093, opened 2015-07-23
Tidy-up some ResultStore logging

smfreegard:
Fixes a couple of annoyances:

1) Error logging

There's a bunch of plugins that do:

`if (err) connection.results.add(plugin, {err: err});` where `err` is actually a proper Error object.   This causes ResultStore to log:

`Jul 23 12:38:12 localhost haraka[25434]: [ERROR] [546435EB-BDF0-44F3-89C4-46A53351A8A6] [Error]`

Which is clearly wrong.   So rather than fixing all of the various plugins, I figured it was likely a fairly common mistake that could be easily handling the case and logging `err.message` instead which then results in the following output:

`Jul 23 13:20:17 localhost haraka[26832]: [ERROR] [42496B88-A3C2-4427-AF05-4DE835E10840] [defendermx/dnswl] IPv6 not supported`

2)  Logging of objects

It seems pretty commonplace for some plugins to build an object, then stuff it into results which causes things like this to be logged:

`[connect.asn] asn: 15169, net: 1.2.3.0/24, cymru: [object Object]`

This patch checks the value being inserted - if it's an object, then it's skipped - so the above line then becomes:

`[connect.asn] asn: 15169, net: 1.2.3.0/24`

In the future I'd like to change the logging to match the rest of Haraka log output as currently the ResultStore output isn't very parseable by tools like grep.

e.g. 

`[connect.asn] asn: 15169, net: 1.2.3.0/24`

should be

`[connect.asn] asn=15169 net=1.2.3.0/24`

As Haraka logging uses key=value entries where booleans are converted to Y/N values and any strings containing spaces are double-quoted.

### DIFF
--- a/result_store.js
+++ b/result_store.js
@@ -3,6 +3,7 @@
 'use strict';
 
 var config = require('./config');
+var util = require('util');
 
 // see docs in docs/Results.md
 var append_lists = ['msg','pass','fail','skip','err'];
@@ -165,7 +166,6 @@ ResultStore.prototype.get_all = function () {
 };
 
 ResultStore.prototype.private_collate = function (result, name) {
-
     var r = [], order = [], hide = [];
 
     if (cfg[name]) {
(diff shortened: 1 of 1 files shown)


msimerson:
> In the future I'd like to change the logging to match the rest of Haraka log output

No objections here.

PR #1097, opened 2015-07-24
Store HAProxy IP in connection object

smfreegard:
Allows you to do something like this in a plugin:

`````
        // Do an A record lookup of the host name
        dns.resolve(host, 'A', function (err, addrs) {
            if (err) return next();
            if (addrs.indexOf(server.notes.public_ip) !== -1) {
                connection.loginfo(plugin, 'host matches public IP');
                return check_user();
            }
            if (addrs.indexOf(connection.haproxy_ip) !== -1) {
                connection.loginfo(plugin, 'host matches HAProxy IP');
                return check_user();
            }
            return next();
        });
````

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -213,6 +213,7 @@ function Connection(client, server) {
     this.errors = 0;
     this.last_rcpt_msg = null;
     this.hook = null;
+    this.haproxy_ip = null;
     setupClient(this);
 }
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by msimerson** [comment]
+1 for the feature.

-1 for the implementation.

Not that the implementation is bad, but I'd **much** rather see our connection data stored in a format like this:

```js
connection = {
    local = {
        ip:   '1.2.3.4',
        proxy: undefined,
        port: 25,
        host: 'mail.example.com',
    },
    remote = {
        ip:   '5.4.3.2',
        port: 23432,
        host: 'mail.sender.com',
    },
    ....
}
```

Yeah, I know, not a change to be made before Haraka 3.0. But I still want it.

smfreegard:
Nice - give me a -1 for something that I can't change yet. ;-)

And yeah - I agree FWIW; the TLS data could live inside there as well.

msimerson:
Please add a comment to #1098 that shows your idea for the TLS data.

PR #1100, opened 2015-07-27
karma: skip deny on outbound hooks

msimerson:
as there's no connection.results to work with


msimerson:
and a silly little early_talker tweak for node 0.12 and iojs-1.8.

PR #1116, opened 2015-09-02
Include dep installation step in git instructions

jaredj:
Installing Haraka with NPM directly takes care of dependencies, but before you can run it out of a git repo, you have to install dependencies.  I assume `npm install` (installing the deps locally rather than globally) is the preferred way to do this


smfreegard:
Yes - that's correct.

Issue #1114, opened 2015-09-01
HELO/rDNS issue while using multiple outbound ip


MojaveKid:
Hi guys,

Im using the multiple outbound ip plugin by baudehlo (#442). 

However, HELO/rDNS check (an anti-spam technique) fails because the HELO string used is different from the rDNS name.

Say you have a mail server mail.xyz.com at 111.111.111.111 and you have an interface  111.111.222.222 (foomail.xyz.com)

If I understand correctly, Haraka issues

HELO mail.xyz.com 

when connecting to a targeted mail server.

Suppose the email went out using interface 111.111.222.222

If the targeted mail server performs rDNS lookup, it will use 111.111.222.222 and resolves this to foomail.xyz.com

HELO and rDNS will not match:   mail.xyz.com != foomail.xyz.com

Is there a way where we can set the HELO connection string before the IP binding to the mx ?

Thanks.

Dexus:
I will take a look what i can make, perhaps a new plugin/config needed for that...

Dexus:
a new param is now available: `mx.bind_helo`
use at the sameway as mx.bind in #442

Dexus:
@smfreegard can you close this now?

@MojaveKid for docs look into https://github.com/baudehlo/Haraka/blob/master/docs/Outbound.md

PR #1127, opened 2015-09-07
Fix for #1114

Dexus:
HELO/rDNS issue while using multiple outbound ip #1114


Dexus:
early_talker - is an early talker fail on 0.12 but is a false positive - perhaps a other bug. 
(As of my repo its okay: https://travis-ci.org/Dexus/Haraka/builds/79097857 - same base as master)

PR #1136, opened 2015-09-08
tls: tmp disable for hosts that fail STARTTLS

### LINKED ISSUE #770: disable TLS negotiation for remotes that fail STARTTLS
# the idea

If a remote attempts to connect and fails, cache their IP in an in-memory no-tls object. The next time that IP connects:

* don't offer STARTTLS
* remove their IP from the no-tls object

This would:

* permit mail to make it though from MTAs and MUAs with old/broken SSL/TLS implementations
* permit Haraka to require more secure TLS settings with less impact
* automatically enable TLS when those hosts fix/upgrade their TLS

msimerson:
fixes #770 

* disabled by default
* progressive enhancement: does nothing if redis unavailable.

### DIFF
--- a/plugins/tls.js
+++ b/plugins/tls.js
@@ -28,7 +28,7 @@ exports.register = function () {
         plugin.logcrit("config/tls_cert.pem not loaded. See 'haraka -h tls'");
         return;
     }
-    
+
     plugin.register_hook('capabilities', 'tls_capabilities');
     plugin.register_hook('unrecognized_command', 'tls_unrecognized_command');
 };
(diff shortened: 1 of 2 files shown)


plugins/tls.js
```diff
@@ -67,16 +68,50 @@ exports.tls_capabilities = function (next, connection) {
...
 
-    /* Let the plugin chain continue. */
-    next();
+    redis.get(dbkey, function (err, dbr) {
+        if (err) {
+            connection.results.add(plugin, {err: err});
+            return enable_tls();
+        }
+
+        if (dbr === null) {
```
smfreegard: Given that there's a PR to merge in ioredis instead of node-redis, you'd be better off doing this:  if (!dbr) otherwise this won't work if we change to ioredis later as it returns an empty string instead of null.

msimerson: suggested change is made.

PR #1138, opened 2015-09-08
uribl: prevent calling next() more than 1x

msimerson:
and a more informative results fail message.


plugins/data.uribl.js
```diff
@@ -166,14 +166,16 @@ exports.do_lookups = function (connection, next, hosts, type) {
     // Perform the lookups
     var pending_queries = 0;
     var called_next = false;
+    var call_next = function (code, msg) {
+        if (called_next) return;
```
smfreegard: You should move the clearTimeout()'s here as well.

msimerson: agreed, and updated.

PR #1140, opened 2015-09-08
fcrdns: don't log error for ENODATA

msimerson:
(now that every request is being made for IPv4 and IPv6, so at least one fails
on every request)

### DIFF
--- a/plugins/connect.fcrdns.js
+++ b/plugins/connect.fcrdns.js
@@ -171,6 +171,7 @@ exports.handle_a_error = function(connection, err, domain) {
     switch (err.code) {
         case 'ENOTFOUND':
         case 'NXDOMAIN':
+        case 'ENODATA':
         case dns.NOTFOUND:
         case dns.NXDOMAIN:
             connection.results.add(plugin, {fail: 'ptr_valid('+domain+')' });


ScOut3R:
Hi,

I am playing with version 2.7.0 and still see error messages related to fcrdns:

```
[ERROR] [046E946F-4C38-4D15-8D3A-128732E54CBF] [connect.fcrdns] queryAaaa ENODATA
```

Also I am seeing the fail:forward_dns(ENODATA) checks in watch and I am getting -1 karma points for the specific host, although the host has a valid HELO name, with a matching forward and PTR record in the ipv4 realm. Am I missing something obvious here or in the configuration?
screen shot 2015-11-21 at 02 11 25

baudehlo:
Yeah that should probably only error if the A lookup fails too. Matt?

msimerson:
Hi @ScOut3R, I'm pretty sure the error is harmless, unless you happen to have a karma rule that penalizes fcrdns errors. If so, remove it, and instead score on particular rDNS errors. The default karma.ini has these entries:

```ini
084 = connect.fcrdns | fail    | match  | ptr_valid   | -4  | FCrDNS has no valid PTR | Set up https://en.wikipedia.org/wiki/Forward-confirmed_reverse_DNS
085 = connect.fcrdns | fail    | match  | valid_tld   | -6  | FCrDNS has no valid TLD | Set up https://en.wikipedia.org/wiki/Forward-confirmed_reverse_DNS
086 = connect.fcrdns | fail    | equals | has_rdns    | -6  | FCrDNS has no rDNS      | Set up https://en.wikipedia.org/wiki/Forward-confirmed_reverse_DNS
087 = connect.fcrdns | ip_in_rdns   | equals | true  |  -1  | FCrDNS has IP in rDNS   | Configure proper hostnames for your MTA
088 = connect.fcrdns | generic_rdns | equals | true  |  -1  | FCrDNS host is generic
089 = connect.fcrdns | fcrdns  | match  |  google.com |  2  | FCrDNS Well Known Sender
090 = connect.fcrdns | fcrdns  | match  | twitter.com |  2  | FCrDNS Well Known Sender
```

Never-the-less, #1247 ignores that error. You can manually apply that if the error messages bothers you.

ScOut3R:
@msimerson Thank you for the quick help and for the fix! :)

Things getting strange here, because I have the default fcrdns rules in karma.ini and I guess the logged fcrdns error confused me. If you check the watch screenshot the forward_dns fail comes from helo.checks. I have to figure out why it failed, because for other, valid domains it is working.

ScOut3R:
@msimerson I think I narrowed down the problem. In helo.check.js (https://github.com/haraka/Haraka/blob/master/plugins/helo.checks.js#L527) the helo name's forward record is checked against both ipv4 and ipv6. I assume if any of those checks returns an error, like ENODATA, then the forward check will fail, but according to the code it shouldn't. I wasn't right when I said that some domains were working right. I made some tests and if a domain does not have an ipv6 A record for its helo name then it will fail the check. Also I've tested the DNS module outside of Haraka and it is working right, so it is not a DNS resolver issue.
I'll try to test it by adding an ipv6 record for a domain.

msimerson:
Hey @ScOut3R , checkout my update to PR #1247 as I think it'll clean that error up in a better way than PR #1249. 

The problem with #1249 is that inside the async block, you really can't know if those errors matter or not. If *both* lookups get an ENODATA, that's an error. If only AAAA does, that's an error. If there's no hosts found and either errors, then that's an error. So the error handling really has to be done after the final async callback returns.

ScOut3R:
@msimerson Thank you for fixing this so quickly! As I said my PR wasn't the best solution. ;)

PR #1146, opened 2015-09-08
replace node_redis with ioredis

Dexus:
switch to ioredis, for more stable version, better performance and iojs compatibility

### DIFF
--- a/docs/Results.md
+++ b/docs/Results.md
@@ -196,7 +196,7 @@ This is from the karma plugin, subscribing on the `connect_init` hook.
 ```js
 var redis = {
     patt: 'result-' + connection.uuid + '*',
-    conn: require('redis').createClient(),
+    conn: require('ioredis').createClient(),
 };
 redis.conn.on('psubscribe', function (pattern, count) {
     connection.loginfo(plugin, 'psubscribed to ' + pattern);
--- a/docs/plugins/dnsbl.md
+++ b/docs/plugins/dnsbl.md
@@ -31,7 +31,7 @@ dnsbl.ini - INI format with options described below:
 
 * enable\_stats
 
-    To use this feature you must have installed the 'redis' module and
+    To use this feature you must have installed the 'ioredis' module and
     have a redis server running.
 
     When enabled, this will record several list statistics to redis.
--- a/docs/plugins/dnswl.md
+++ b/docs/plugins/dnswl.md
@@ -36,7 +36,7 @@ dnswl.ini - INI format with options described below:
 
 * enable\_stats
 
-    To use this feature you must have installed the 'redis' module and
+    To use this feature you must have installed the 'ioredis' module and
     have a redis server running.
       
     When enabled, this will record several list statistics to redis.
--- a/docs/plugins/rate_limit.md
+++ b/docs/plugins/rate_limit.md
@@ -10,11 +10,11 @@ tarpit the connection by adding a delay before every response sent back to the
 client instead of sending a DENYSOFT.  To do this requires the 'tarpit' plugin 
 to run immediately after this plugin.
 
-To use this plugin you will need a Redis server and will need the redis, 
+To use this plugin you will need a Redis server and will need the ioredis, 
 hiredis and ipaddr.js packages installed via:
 
     cd /path/to/haraka/home
-    npm install redis hiredis ipaddr.js
+    npm install ioredis hiredis ipaddr.js
     
 Configuration
 -------------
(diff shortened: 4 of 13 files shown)


msimerson:
I have a few reservations about this.

1. I'm happy with node-redis. I haven't even heard of *someone* having stability issues with node-redis.
2. To get iojs compat, we just need to update our ancient redis module dep from 0.10.1 (Feb 17, 2014) to the later versions.
3. To get the same higher performance with node-redis, just install hiredis (same as with ioredis).
4. I really dislike dependencies that require compilation (hiredis). 
    * they can blow up later because an underlying library / binary was updated. ("If you use hiredis, be sure to rebuild it whenever you upgrade your version of node. There are mysterious failures..." -- hiredis docs).
    * Our (very few) compile-time dependencies generate an inordinate amount of support / install issues.
    * IMO (I worship at the church of uptime), hiredis should **not** be a dependency. Add it to the docs, suggest it, emphatically even, but please don't **require** it.
5. redis transactions are in use but not all uses (see karma plugin) are updated by this PR.
6. Because redis is loaded in so many places, I'd much rather see it get encapsulated in a library so that switching between node-redis and ioredis and whether or not to install/update/use hiredis was a config file setting.

Dexus:
> I'm happy with node-redis. I haven't even heard of someone having stability issues with node-redis.
I use redis in a 60-80k req/s API and there sometimes (more as i like) brake the node-redis but ioredis doesn't. And I have even a little room for improvement with ioredis.

> To get iojs compat, we just need to update our ancient redis module dep from 0.10.1 (Feb 17, 2014) to the later versions.
Thats right, this is an option.

> To get the same higher performance with node-redis, just install hiredis (same as with ioredis).
This is recommend on production systems. ;)

> I really dislike dependencies that require compilation (hiredis).
I have added it because i used it in production systems that have massive mail to process. So its a win, but it is not a must have dep.

> redis transactions are in use but not all uses (see karma plugin) are updated by this PR.
What is not updated and need to be updated? My tests works like excepted. But will not rule out that I've forgotten something.

> Because redis is loaded in so many places, I'd much rather see it get encapsulated in a library so that switching between node-redis and ioredis and whether or not to install/update/use hiredis was a config file setting.
That is a good point. I'll think about it again.

smfreegard:
Here's my 2c.

Personally - I've used node_redis with hiredis since this beginning.  I've never used node-redis standalone and I've not had any issues with it so far.

However - I hadn't discovered ioredis until you pointed it out and it's obvious that it is a major improvement over node_redis (e.g. support for sentinel and cluster (completely missing in node_redis) and much better pubsub and Lua support).

I have extensive non-public plugins that all use node_redis that will need to be modified to run with ioredis - but it's something that I'm going absolutely going to do.

Just because we have a bunch of plugins that use redis now doesn't mean we shouldn't convert them all to use ioredis and discard node_redis completely particularly as we're moving to a new point release.   Why bother with the effort of writing a library to encapsulate it?  Seems totally unnecessary to me and I can see it causing more issues than it's worth given that ioredis and node_redis treat empty replies differently - there's far more scope for bugs.

Anyone with non-public plugins (like me) just has to deal with it - it's hardly a problem running both modules in the interim and it's not like this is going to break an existing installation using node_redis already.

Dexus:
redis 2.0.x is there, with breaking changes (maybe) same as ioredis. i'll check differences today.
Ref: https://github.com/haraka/haraka-nosql/issues/2

msimerson:
There shall be [one Redis to rule them all](https://github.com/luin/ioredis#consolidation-its-time-for-celebration). Eventually.

msimerson:
I'm closing this now because:

* it has branch conflicts (my fault)
* since the merge of the new `redis` plugin, the other plugins that use redis now inherit from a common `redis` plugin, erasing my points 5 and 6 above. 
* a future PR that replaced node_redis with hiredis would be much smaller
* a PR could support **both** node_redis and hiredis, much like how we choose which syslog module, or geoip module, based on which one successfully loads at run time.

celesteking:
So what have you decided? Another PR for ioredis? Or no ioredis in foreseeable future?

sumpton:
My 2 cents. We have a complicated stack for our primary project, with Redis our primary store, supported by Elasticserch and MongoDB. Our main Redis is a cluster, so we follow the driver adoption pretty closely. We are seeing others adopt ioredis -- which is supported by http://www.alibaba.com/. I really can't see ioredis going away now.

msimerson:
The primary reasons this PR wasn't merged straight away were that it affected a number of plugins and it *forced* everyone to use ioredis. Now that we have a redis module which most (all?) the redis-using plugins inherit from, a new PR could be created that made choosing between node-redis and ioredis a config file setting, or even a "detect which is installed and use it."

PR #1147, opened 2015-09-08
Update package.json latest versions and .travis.yml to build correct …

Dexus:
…iojs and nodeja v4 used v8 engine with modules

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,21 +2,24 @@ language: node_js
 node_js:
     - "0.12"
     - "0.10"
-#   - "0.8"   # no longer supported by iconv
-#   - "0.6"   # no longer supported by async
-    - "iojs-v1.8.1"
-    - "iojs"
-
+#   - "0.8"     # no longer supported by iconv
+#   - "0.6"     # no longer supported by async
+    - "stable"
+    - "iojs-v3"
 services:
     - redis-server
-
-matrix:
-    allow_failures:
-        - node_js: "0.8"
-        - node_js: "iojs"
-#       - node_js: "0.6"
-
+    
+# these are required for building on node.js 4.0 / io.js v3
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - g++-4.8
+env:
+  - "CXX=g++-4.8"
+# end: these are required for building on node.js 4.0 / io.js v3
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,21 +2,20 @@ language: node_js
...
     - "0.10"
-#   - "0.8"   # no longer supported by iconv
-#   - "0.6"   # no longer supported by async
-    - "iojs-v1.8.1"
-    - "iojs"
-
+    - "stable"
+    - "iojs-v1"
+    - "iojs-v2"
+    - "iojs-v3"
```
msimerson: I doubt we'll ever care to deploy to iojs 1, 2, or 3, and instead target node.js 4.0 (the unforked branch). So not much point in adding them here.

.travis.yml
```diff
@@ -2,21 +2,20 @@ language: node_js
...
-#       - node_js: "0.6"
-
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - g++-4.8
+env:
+  - "CXX=g++-4.8"
```
msimerson: I think these options are no longer required.

Dexus: This lines need. See https://travis-ci.org/Dexus/Haraka/builds/81483117

msimerson: Okay, so put a comment there telling why: "these are required for building on node.js 4.0"

.travis.yml
Dexus: ok but iojs gets future updates... i dont know if only for bugs or also for new features. So I think that iojs-v3 should remain there.

.travis.yml
```diff
@@ -2,21 +2,20 @@ language: node_js
 node_js:
     - "0.12"
     - "0.10"
-#   - "0.8"   # no longer supported by iconv
-#   - "0.6"   # no longer supported by async
```
msimerson: These comments, telling **why** we don't support node 0.8 and 0.6 tend to be helpful, which is why we leave them in there.

package.json
```diff
@@ -14,36 +14,36 @@
     "node": ">= v0.10.0"
   },
```
baudehlo: We should update this to 10.36 at least due to bad certs in the older nodes.

Issue #542, opened 2014-04-02
Can't get 'aliases' plugin working


felixrabe:
I'm running on master (the legendary 5f5fa3be commit), and have a working inbound server set up. Now I tried adding the aliases plugin and configured some test aliases. But they don't work for me, as Haraka tries to relay them on to Dovecot as all other mail, where it gets rejected as that user does not exist, causing a "failure notice" reply. The aliases plugin actually does not execute its `exports.aliases` hook.

This is my configuration: http://nutzwertanalyse.ch/mail.rabe.io-inbound.tar.gz

config/aliases:

    {
      "to-gmail@mcr.io": { "action": "alias", "to": "f.daniel.cuervo@gmail.com" },
      "to-gmx@mcr.io":   { "action": "alias", "to": "felixrabe@gmx.ch" }
    }

rabe.io == mcr.io. The only change (cf https://gist.github.com/felixrabe/20cacf0dba5e3d22d342) in my aliases_felix plugin is adding the line

    connection.logdebug(plugin, 'aliases plugin running');

I do not see `aliases plugin running` in my logs even though they show debug messages indicating that `aliases_felix` plugin has been loaded. Gist https://gist.github.com/felixrabe/e60b4a178cab720856f8 shows the former log output before switching to `aliases_felix`.

Did I miss something or is this a bug?

Dexus:
@felixrabe sorry for the delay, but now I have update the aliase plugin. If should do now what you want.

@msimerson please close, thank you.

msimerson:
@Dexus , when you submit a PR that fixes / closes an issue, include the special syntax: "closes #542" in the PR comments, and this Issue would have been closed automatically.

PR #1149, opened 2015-09-08
Fix for #542

Dexus:
- added support for "user@host" alias, "@host" alias
- before check the "user" alias.


msimerson:
with docs and tests, yay!

Dexus:
:grinning:

PR #1151, opened 2015-09-09
Clean up ip_fixtures array - remove duplicates, sort

Synchro:
I spotted that the fixtures list contained several duplicates (about 20) so I stripped them out and then sorted the whole lot, so it's more readable, easier to find a specific example.


smfreegard:
Thanks @Synchro

Issue #1125, opened 2015-09-06
IPv6 compatibility: ToDoList


Dexus:
To make Haraka full IPv6 compatible we need to add more checks that covers the IPv4 checks, to get the same result on IPv6 connects.

- [x] review all current IPv4 checks in helo.checks and add IPv6 compatible checks
- [x] data.uribl
- [x] all IPv4 RegExp Test, must also check against IPv6
- [x] check that also used IPv6 BLs
- [x] resolve must be made compatible for use with IPv6
- [x] update regexp with "IPv6:" Prefix in IPv6 helo check statements

---
If you have some additional tasks, please let it me know.
Thats my first small look, what need to be done.

https://www.ietf.org/rfc/rfc2821.txt

msimerson:
I'd suggest moving all those checks into net_utils. We've already got some helper functions in there and having additional ones in there promotes reuse.

Dexus:
:+1: the first steps are made.

Dexus:
@baudehlo @msimerson @smfreegard 
Any known task that i don't have covered in my list?

PR #1154, opened 2015-09-09
IPv6 review - finished

Dexus:
Notices, discuss and todolist at #1125


plugins/spf.js
```diff
@@ -73,7 +73,7 @@ exports.hook_helo = exports.hook_ehlo = function (next, connection, helo) {
 
     // RFC 4408, 2.1: "SPF clients must be prepared for the "HELO"
     //           identity to be malformed or an IP address literal.
-    if (net_utils.is_ipv4_literal(helo)) {
+    if (net_utils.is_ip_literal(helo)) {
         connection.results.add(plugin, {skip: 'ipv4_literal'});
```
smfreegard:
`````
 connection.results.add(plugin, {skip: 'ipv4_literal'});
`````

should now be 


`````
 connection.results.add(plugin, {skip: 'ip_literal'});
`````

**Comment by Dexus** [comment]
Notices, discuss and todolist at #1125

**Comment by Dexus** [comment]
So now i think its finished.

File: net_utils.js [hunk 2]
```diff
@@ -421,3 +428,32 @@ exports.get_ipany_re = function (prefix, suffix, modifier) {
         modifier
     );
 };
+
+exports.ipv6_reverse = function(ipv6){
+    ipv6 = new IPv6.Address(ipv6);
```
smfreegard: Missing 'var' declaration.

Additionally - do we really need the 'ip-address' module for this being as we already have the ipaddr.js module.

net_utils.js
```diff
@@ -421,3 +428,32 @@ exports.get_ipany_re = function (prefix, suffix, modifier) {
...
+        var done = false;
+        var data;
+        var reserveHost = ipCheck.reverseForm().replace('.ip6.arpa.', '')+'.v6.fullbogons.cymru.com.';
+        dns.resolve4(reserveHost, function (err, ip){
+            if(!err) {
+                data = ip[0];
+            }
+            done = true;
+        });
+        deasync.loopWhile( function () { return !done; } );
```
smfreegard: Yuck.   Don't like this at all.   Another module dependency and an async function should return a callback IMO.

plugins/data.uribl.js
```diff
@@ -106,11 +109,13 @@ exports.do_lookups = function (connection, next, hosts, type) {
...
             }
-            // Convert in-addr.arpa into bare IPv4 lookup
+            // Convert in-addr.arpa into bare IPv4/v6 lookup
             var arpa = host.split(/\./).reverse();
-            if (arpa.shift() === 'arpa' && arpa.shift() === 'in-addr') {
-                if (arpa.length < 4) continue; // Only full IP addresses
-                host = arpa.join('.');
+            if (arpa.shift() === 'arpa'){
+                if ( ['in-addr','ip6'].indexOf(arpa.shift()) !== -1) {
+                    if (arpa.length < 4) continue; // Only full IP addresses
```
smfreegard: This code is looking for 4.3.2.1.in-addr.arpa and converting that into 4.3.2.1 instead.

You've added a check for ip6.arpa, but that's not going to work properly being as there's no length check when it is IPv6 e.g. arpa.length < 4 is fine for IPv6, but not IPv6, so it's slightly wrong.

net_utils.js
Dexus: `ipaddr.js` has not the power of `ip-address` and build-in functions.

smfreegard: addr.toNormalizedString() would suffice for this function without the additional dependency.

net_utils.js
```diff
@@ -421,3 +428,32 @@ exports.get_ipany_re = function (prefix, suffix, modifier) {
...
     );
 };
+
+exports.ipv6_reverse = function(ipv6){
+    ipv6 = new IPv6.Address(ipv6);
+    return ipv6.reverseForm().replace('.ip6.arpa.', '');
+};
+
+exports.ipv6_bogus = function(ipv6){
+    var ipCheck = new IPv6.Address(ipv6);
```
smfreegard: https://github.com/whitequark/ipaddr.js/blob/master/src/ipaddr.coffee#L226

net_utils.js
Dexus: This will not work, because `ip-address` make it to full lenght other than `ipaddr.js`

```
ipaddr.js - toNormalizedString: 2001:db8:8:66:0:0:0:1
ip-address - canonicalForm: 2001:0db8:0008:0066:0000:0000:0000:0001 << needed to reverse
```

net_utils.js
Dexus: Okay, checks there are top :+1:

net_utils.js
smfreegard: toNormalizedString().split(':').reverse().join('.') + '.ip6.arpa' 

Should provide the same I think.

Dexus: nope, thats the problem... but i will make the change that needed.

msimerson:
@Dexus , do you know how to squash a PR?  See the [instructions](https://github.com/baudehlo/Haraka/wiki/Contributing) and catch me on #haraka if you need assistance.

Dexus:
Updated: Fixes overwriting the addresses that were previously fetched. (L98++ in mail_from.is_resolvable.js)

plugins/data.uribl.js
```diff
@@ -4,6 +4,7 @@
 var url       = require('url');
 var dns       = require('dns');
 var isIPv4    = require('net').isIPv4;
+var isIPv6    = require('net').isIPv6;
```
msimerson: For consistency, we've been updating these to only export `net`, and then when used, reference `net.isIPv4` and `net.isIPv6`. Otherwise, this PR looks good to me.

PR #1123, opened 2015-09-05
ipv6 compatible ...

Dexus:
ipv6 compatible connect.fcrdns.js + helo.checks.js + fix/improve connect.geoip.js

PR #1156, opened 2015-09-10
repair ptr_valid errors introduced with #1123

msimerson:
After merging #1123, every domain has an A and AAAA query performed. The AAAA query failures were calling handle_a_error, which certainly was the wrong thing. 

This logs and otherwise ignores the errors, so a ptr_valid failure is only emitted if neither query returns a result.


Dexus:
Fix please the tests.

msimerson:
tests are now fixed.

Issue #937, opened 2015-04-20
RFC3464 Compliance for Bounce messages


smfreegard:
https://tools.ietf.org/html/rfc3464

related to #936 
see also #848

msimerson:
closed by #1189

PR #1157, opened 2015-09-11
Improved outbound testability

hatsebutz:
Added a basic outbound test here: Injecting an 5XX in outbound-mail-conversation and asserting that hmail.bounce gets called.
Does have (small) impact on outbound.js: try_deliver_host was splitted, so it is possible to inject a mocked socket.

As I am somewhat new to nodeunit/nodejs comments and feedback is highly appreciated. 

Maybe this basic testing of outbound could be helpful to write tests for #937 too.


smfreegard:
Currently failing tests on node 0.10.x

hatsebutz:
Fixed tests for node node 0.10.40

Dexus:
Tests dont fail with last commit. Travis has swallowed during the build.

PR #1158, opened 2015-09-13
greylisting - finished - running live test

### LINKED ISSUE #1110: greylisting plugin
https://github.com/celesteking/Haraka/pull/1

Can't diff against your repo because apparently my tracking is f.. up.

Dexus:
close #1110 orgin: celesteking/Haraka/pull/1

Plugin seems to consider only IPv4. IPv6 review shall be conducted.

* clean up
* one config file and not 4...
* remove "special-sender"
* greylist is greylist and dont need reject/learn actions

## Comments from commit promoted here by @msimerson

*Fix IPv6 incompatibility
* Update docs / config info for action "learn".
* rework: remove Q deps
* rework: async - done
* set "return" to prevent double callbacks
* fix bug cb not a function

### DIFF
--- a/config/greylist.ini
+++ b/config/greylist.ini
@@ -1,0 +1,43 @@
+; Config for greylisting plugin
+
+; greylisting action text
+text = Greylisted. Please come back later.
+
+[redis]
+host = 127.0.0.1
+; port = 6379
+db = 11
+
+[skip]
+; skip for DNSWL hosts having high reputation
+dnswlorg = true
+mailspikewl = true
+
+[period]
+# transition path: first_connect --> black (defer) --> grey(allow) --> white (allow) --> expired
+
+# 14 minutes
+black = 850
+# 25 hours
+grey = 90000
+# 35 days
+white = 3024000
+
+[envelope_whitelist]
+# Envelope emails or domains, one per line
+
+
+[ip_whitelist]
+# IP or Subnet, one per line
+
+
+[recipient_whitelist]
+# Recipient emails or domains, one per line
+
+
+[special_dynamic_domains]
+# Put domains that should be always treated as dynamic here.
+# Pattern is matched at the end of rdns
+
+# SiteGround VPS service
+sgvps.net
--- a/docs/plugins/greylist.md
+++ b/docs/plugins/greylist.md
@@ -1,0 +1,90 @@
+# Greylist
+
+Basic greylisting plugin that follows common practices found on internets.
+
+## Principles of work
+
+### Notation
+
+The so-called _tuple_ consists of the following:
+
+* First subdomain of rDNS is stripped off (but no shorter than the domain boundary). This is considered a _hostid_.
+* Envelope sender is the _sender_.
+* RCPT TO would supply the _recipient_.
+
+_hostid_ in above notation is chosen unless:
+
+1. The connecting host has no PTR record, a.k.a. reverse DNS (rDNS). [gl]
+1. The rDNS record contains the first two or last two octets of the IP address. [fcrdns]
+1. The rDNS record contains the ‘short’, decimal, or hex representation of the full IP address. [fcrdns] [gl]
+1. Multiple rDNS records are returned. [gl]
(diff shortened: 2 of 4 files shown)


Dexus:
@smfreegard 
Have restored with IPv6 compatibility. 

>from: celesteking/Haraka/pull/1#issuecomment-132935885
>
>Initial comments after a quick scan:
>
> - LOL - someone read my greylisting paper ;-).
> - actions are undocumented. I'm a bit fuzzy on how 'learn' can possibly work (and be remotely accurate) and how 'black' isn't crazy dangerous.
> - 'special cases' are undocumented?

I don't have looked for the undocumented actions in detail. But __learn__ action will collect data but not reject or defer mails.

The 'special cases' i don't know. But it looks as if dynamic rDNA of certain hosters, only the IP should use instead of the domain.

Dexus:
[ ] remove Q and use call backs.

Dexus:
Works on live system like a busy bee :)
Last hours 29k greylisted; 100 whitelisted - works.
Now I'll be watching the next week.

tests/plugins/data.greylist.js
```diff
@@ -0,0 +1,76 @@
...
+            "ip":[
+                [{"octets":[123,123,123,234]},32],
+                [{"parts":[10754,33284,54784,32864,31008,16448,8430,38528]},128],
+                [{"octets":[123,210,123,234]},27],
+                [{"parts":[10754,33284,54784,32864,31008,61166,0,0]},96]
+            ]
+        };
+        this.plugin.list = {"dyndom":["sgvps.net"]};
+        test.expect(7);
+        test.equal(true,  this.plugin.addr_in_list('mail', 'josef@examlpe.com'));
```
msimerson: Ahah, I think I see your problem.

````
this.plugin.whitelist = {
+            "mail":{"josef@examlpe.com":true},
````

You are setting this.plugin.whitelist.mail above.

Here you are accessing this.plugin.mail.

smfreegard:
So the 'learn' mode is staying?   
What constitutes a 'special-sender' and why would you want to defer the deferral until hook_data?

msimerson:
> So the 'learn' mode is staying?

I sure hope so. IMO, the only sensible way to deploy greylisting is to run it in learn mode for a few weeks before deploying.

Dexus:
@smfreegard 
special-sender are (postmaster|double-bounce(\-?\d{8,12})?) they never should be greylisted.
i call that now on rcpt_ok

smfreegard:
>> So the 'learn' mode is staying?

> I sure hope so. IMO, the only sensible way to deploy greylisting is to run it in learn mode for a few weeks before deploying.

Good luck with that.   You might as well not bother running greylisting at all if you do that.

It's *impossible* to 'learn' in this case.   You can't determine if a remote system implements a retry queue without testing and observing it.

'Learn' in this context basically means, whitelist *everything* that comes along.  Which means that when you do switch this into 'defer' mode - it will do nothing, because everything was whitelisted whether it queued or not.

smfreegard:
> special-sender are (postmaster|double-bounce(-?\d{8,12})?) they never should be greylisted.

Why not?   A sending host either implements a retry queue or it doesn't - the actual sender is irrelevant to that fact.

Dexus:
So now i removed special-sender, default defer action and clean up hook_data step to hook_rcpt_ok.

celesteking:
>> special-sender are (postmaster|double-bounce(-?\d{8,12})?) they never should be greylisted.

>Why not? A sending host either implements a retry queue or it doesn't - the actual sender is irrelevant to that fact.

That's for cases when sender should not be deferred in RCPT TO phase. For ex., email address verification. Returning temp error will probably count as error for them, thus the need for additional code.

celesteking:
> 'Learn' in this context basically means, whitelist everything that comes along

Yes, learning is broken in this regard, sadly. I'm busy working on other things, so I guess no learn mode in near future. On bright side, it won't do much harm to HAM messages, you'll only end up with a bunch of whitelisted spammy hosts that will expire in a month.

PR #1165, opened 2015-09-21
Outbound.send_email: added dot_stuffing

hatsebutz:
added dot-stuffing to outbound.send_email function (included additional options-param to allow disabling this changed behaviour)

### DIFF
--- a/outbound.js
+++ b/outbound.js
@@ -354,7 +354,10 @@ exports.send_email = function () {
     var from = arguments[0],
         to   = arguments[1],
         contents = arguments[2];
-        var next = arguments[3];
+    var next = arguments[3];
+    var options = arguments[4];
+
+    var dot_stuffed = ((options && options.dot_stuffed) ? options.dot_stuffed : false);
 
     this.loginfo("Sending email via params");
 
@@ -408,6 +411,9 @@ exports.send_email = function () {
     while (match = re.exec(contents)) {
         var line = match[1];
         line = line.replace(/\r?\n?$/, '\r\n'); // make sure it ends in \r\n
+        if (dot_stuffed === false && line.length > 3 && line.substr(0,1) === '.') {
+            line = "." + line;
+        }
         transaction.add_data(new Buffer(line));
         contents = contents.substr(match[1].length);
         if (contents.length === 0) {
--- a/docs/Outbound.md
+++ b/docs/Outbound.md
@@ -276,3 +276,10 @@ should queueing to disk fail e.g:
 
     outbound.send_email(from, to, contents);
 
+
+In case you are passing your content dot-stuffed (a dot at the start of a line
+is doubled, like it is in SMTP conversation, 
+see https://tools.ietf.org/html/rfc2821#section-4.5.2), you should pass the
+```dot_stuffed: true``` option, like so:
+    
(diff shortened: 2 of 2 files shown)


outbound.js
```diff
@@ -352,7 +352,10 @@ exports.send_email = function () {
     var from = arguments[0],
         to   = arguments[1],
         contents = arguments[2];
-        var next = arguments[3];
+    var next = arguments[3];
+    var options = arguments[4];
+
+    var dot_stuffing = ((options && typeof(options.dot_stuffing) !== 'undefined') ? options.dot_stuffing : true);
```
smfreegard: This needs to be documented.

I would be tempted to default this on, as we should probably presume that 'contents' isn't dot-stuffed when passed and the option would disable this functionality rather than enable it.

@baudehlo: Any thoughts on this?

hatsebutz:
Renamed the option from ```"dot_stuffing"``` (default: true) to ```"dot_stuffed"``` (default: false), as dot_stuffed (refering to the passed content) seems more logical. Added a short documentation about the option to Outbound.md

outbound.js
hatsebutz: I've renamed it to "dot_stuffed" and negated the logic (default ```false```), so it describes the passed ```contents```.

baudehlo:
Yeah it should probably be on by default. It seems like bogus behaviour to expect it to be pre-dot-stuffed.

But we should check if we're doing outbound.send_email anywhere in the current code to make sure we do the right thing.

hatsebutz:
There are 2 types of usage:
* ```send_email(transaction, next)```: Not affected by this change (used in Connection.queue_outbound_respond, plugin/queue/deliver, plugin/queue/lmtp)
* ```send_email(from,to,contents,callback,options)```: Used only in outbound.bounce_respond (should be ok here since the bounce message is not expected to be dot-stuffed)

smfreegard:
This looks OK now.  Can you pull master again and rebase so I don't have to merge this manually?

baudehlo:
Yeah it's used in the mailing list code too.
https://github.com/baudehlo/haraka-mailinglist

Not that anyone uses that code except the mailing list itself (and nobody
really uses the mailing list).

msimerson:
@hatsebutz, you rebase like this:

````
git remote update
git rebase -i baudehlo/master
````

and in your editor, squash all the merges. That'll also get rid of the "merge branch" stuff from this branches history.

msimerson:
shout if you have issues and I'll rebase and squash it for you.

hatsebutz:
@msimerson, I would like to ask if you could do that rebase/squash for me. Did something wrong and I am now stuck with it.... probably should restart from an older commit I think, or make a new branch/PR, but don't really sure

msimerson:
generally, I find that branches rebase easily, until/unless you merge in another branch. IE, if you use rebase exclusively and don't "git merge," it works clean, fast, and easy. After a merge, yeah, it's usually easier to create a new branch/PR.

hatsebutz:
@msimerson thanks! In the meanwhile I tried rebase on a new branch, and I think I've got it (and can now see my previous mistake, starting with rebasing on outdated local repo).

PR #1169, opened 2015-09-22
introduce eslint and consistent 4-space indents

msimerson:
* personally, I prefer 2-space indents, like 90+% of the JS community, but
I value internal consistency higher than external conformity (and am not motivated enough to write the script to automate that).
* lots of whitespace changes

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,8 @@
+{
+  "env": {
+    "node": true
+  },
+  "rules": {
+    "indent": [2, 4, {"SwitchCase": 1}],
+  }
+}
--- a/plugins/clamd.js
+++ b/plugins/clamd.js
@@ -16,49 +16,49 @@ exports.load_excludes = function() {
     for (var i=0; i < list.length; i++) {
         var re;
         switch (list[i][0]) {
-        case '!':
+            case '!':
 
-            if (list[i][1] === '/') {
-                // Regexp exclude
+                if (list[i][1] === '/') {
+                    // Regexp exclude
+                    try {
+                        re = new RegExp(list[i].substr(2, list[i].length-2),'i');
+                        new_skip_list_exclude.push(re);
+                    }
+                    catch (e) {
+                        plugin.logerror(e.message + ' (entry: ' + list[i] + ')');
+                    }
+                }
+                else {
+                    // Wildcard exclude
+                    try {
+                        re = new RegExp(
+                                utils.wildcard_to_regexp(list[i].substr(1)),'i');
+                        new_skip_list_exclude.push(re);
+                    }
+                    catch (e) {
+                        plugin.logerror(e.message + ' (entry: ' + list[i] + ')');
+                    }
+                }
+                break;
+            case '/':
+                // Regexp skip
                 try {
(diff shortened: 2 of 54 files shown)


plugins/clamd.js
```diff
@@ -16,49 +16,49 @@ exports.load_excludes = function() {
...
+                    }
+                }
+                break;
+            case '/':
+                // Regexp skip
                 try {
-                    re = new RegExp(list[i].substr(2, list[i].length-2),'i');
-                    new_skip_list_exclude.push(re);
+                    re = new RegExp(list[i].substr(1, list[i].length-2),'i');
+                    new_skip_list.push(re);
```
smfreegard: ??

smfreegard:
This is not a noop change.

If you're going to modify code e.g. clamd, backscatterer, then it needs to be done in separate pulls otherwise a patch bomb like this is impossible to review effectively.

There should be one pull for eslint e.g. .eslintrc, grunt and the sourcecode after eslint has been run on it and separate pulls for the other changes.

.eslintrc
```diff
@@ -0,0 +1,8 @@
+{
+  "env": {
+    "node": true
+  },
+  "rules": {
+    "indent": [2, 4, {"SwitchCase": 1}],
```
smfreegard: Having had a quick look as eslint - I would also suggest:

`````
one-var: [2, "never"]
`````

As that was always our preferred style (I've added it to the contributing page on the Wiki as it was missing).

plugins/clamd.js
msimerson: This really is **only** a whitelist change. If you use side-by-side diffs with a really wide screen it makes the change easier to see.

This does raise an interesting question. To indent case or not?  I prefer no but I chose yes, based on **most** of the `case` calls being already indented.

.eslintrc
msimerson: There's plenty more settings that could/should be added, but that would make this PR even bigger, so just indention is more than sufficient for this PR.

msimerson:
The backscatter change is this:

````
this.first(connection.remote_ip, [ 'ips.backscatterer.org' ], function (err, zone, a) {
    // some stuff
});
````

-vs- 

````
function resultCb (err, zone, a) {
    // some stuff
}

this.first(connection.remote_ip, [ 'ips.backscatterer.org' ], resultCb);
````

In addition to *wanting* deeper indention, the first example is also creates a line longer than 80 chars. The simplest solution is to name and declare the function.

smfreegard:
But it then isn't a no-op change is it?   We're changing code rather than just whitespace and indents (which is what I presume the noop was for).

And this:

`````
this.first(connection.remote_ip, [ 'ips.backscatterer.org' ], function (err, zone, a) {
`````

Could just as easily be:

`````
this.first(connection.remote_ip, 
           [ 'ips.backscatterer.org' ], 
           function (err, zone, a) 
{
...
});
`````

But I'll agree having a named function is better than an anonymous function should any stack traces be output.

My rant is that this is a *huge* patch to review and if you mix whitespace, tabs and actual code changes together makes it really hard to spot errors which is the whole point of reviewing it in the first place.

So I haven't actually reviewed this properly - I got to the first functional change and then stopped.

PR #1170, opened 2015-09-22
Add license clause

Synchro:
Keeps npm quiet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "author": "Matt Sergeant <helpme@gmail.com> (http://baudehlo.com/)",
   "name": "Haraka",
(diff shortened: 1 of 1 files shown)


Synchro:
Ah, I just spotted that npm complains because it has an old-style license that npm no longer supports, will correct.

Dexus:
@Synchro please sqash it together thanks.

Synchro:
I just did that and force-pushed - not sure why it kept the old commits.

Dexus:
Git only adds live but cant remove old data live ;)

Synchro:
Thanks

Dexus:
NP

PR #1189, opened 2015-09-25
Outbound Bounce messages according to RFC3464

hatsebutz:
This PR adds bounce messages in RFC3464 format for outbound failures:

- config/outbound.bounce_message is used for the headers and the human-readable part
- bounce message contain original mail headers only
- re-uses dsn.js for errors before talking to remote-MTA
- uses remote-MTA codes on failures during SMTP conversation
- includes tests
- includes no docs except some lines in source
- reorganized outbound_protocol tests to reuse fixtures/vm_harness

Feedback welcome

### DIFF
--- a/config/outbound.bounce_message
+++ b/config/outbound.bounce_message
@@ -11,6 +11,3 @@ This is a permanent error; I've given up. Sorry it didn't work out.
 
 Intended Recipients: {recipients}
 Failure Reason: {reason}
-
---- Below this line is a copy of the message.
-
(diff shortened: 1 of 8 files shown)


smfreegard:
+1 From me - this looks great.

@msimerson - are you happy with the tests?

msimerson:
My response to this PR is to blow @hatsebutz a kiss.

 I think there's an outstanding issue related to DSN and we should check if this closes it.

Dexus:
:+1: have test it live and i'm happy :)

baudehlo:
There's one bit of the code where the previous call to .bounce() is just
commented out. I'd like to get that removed not just commented out.

smfreegard:
Sorry @baudehlo - I missed that.   Addressed in #1190 along with some others that were also in outbound.js.

PR #1200, opened 2015-10-04
Troubleshooting for the most recurring issues.

Dexus:


docs/Troubleshooting.md
```diff
@@ -0,0 +1,14 @@
...
+============
+
+Listen EACCES error
+---------
+
+For ports below 1024, you need root privileges to listen on port 25, 465 or 587 etc.
+
+Useful links:
+- https://en.wikipedia.org/wiki/Authbind
+- http://manned.org/authbind.1
```
msimerson: This is a markdown file, so why not make proper links? IE:

````
http://manned.org/authbind.1
```` 

becomes

````
[http://manned.org/authbind.1](http://manned.org/authbind.1)
````

and then renders to: [http://manned.org/authbind.1](http://manned.org/authbind.1)

Bonus points if you replace the first link with a good description. 

Double extra super bonus points if you instead publish this on the Wiki instead, and perhaps put a single tiny little file here in docs, such as HELP.md with nothing else but a link to the Wiki FAQ / Help / Troubleshooting page.

Dexus: :) ok

PR #1212, opened 2015-10-07
rename plugins

Dexus:


msimerson:
Be warned, there's a significant amount of controversy regarding renaming plugins with "connection phase" prefixes. This should not be merged until everyone agrees.

Dexus:
On my TODO for tomorrow
- [ ] rename config
- [ ] change docs
- [ ] change tests

Dexus:
@smfreegard @baudehlo 
what do you think can we merge this? Or any objections?

smfreegard:
Personally I'm -1 on this.   I think the whole connect/mail/rcpt/data/data_post naming is crap.   If it were up to me; I'd go completely the other way e.g. has_mx or require_mx is way less typing than mail_from.is_resolvable.   Additionally, you *could* do MX checks on the Reply-To/From/Sender headers (which I do here), which if you do it in the same plugin makes the whole naming scheme total bollocks.

But I'll go with the majority, so if @baudehlo is OK with it - then go ahead.

smfreegard:
Oh - and one other example.   *data*.uribl.js checks the rDNS, HELO, MAIL arguments in addition to any URIs parsed out from the mail during hook_data_post, so that's already a good example of why the naming scheme is wrong.   It should just be called uribl - like it was in rbldnsd IIRC.

Dexus:
maybe, we should refactor the Plugin process and give each hook a priority.
but only as an aside in this PR ...

smfreegard:
Huh?  I added that recently: https://github.com/haraka/Haraka/blob/master/docs/Plugins.md#plugin-run-order

Dexus:
ups i miss that... perfect, so we dont need to rename - but i think its not a bad idea to "categorize" the plugins to the "main" hook.

smfreegard:
Sure - if you like the extra typing.   Because it's not just the plugin name, it's the configuration files too.  I just think it brings nothing to the table, it's not "easier" for a newbie to understand (maybe they'll incorrectly assume that a plugin prefix denotes where it runs...).

PR #1213, opened 2015-10-07
add auth support to smtp_forward

msimerson:
I think we should merge this and close #430 

Thoughts?

closes #430

### DIFF
--- a/plugins/queue/smtp_forward.js
+++ b/plugins/queue/smtp_forward.js
@@ -50,6 +50,35 @@ exports.hook_queue = function (next, connection) {
 
     var smc_cb = function (err, smtp_client) {
         smtp_client.next = next;
+        
+        var config = plugin.cfg;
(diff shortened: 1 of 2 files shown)


Dexus:
if we merge it, we need also docs updated.
But from me its ok.

baudehlo:
What are the hard-coded base64 values in there?

smfreegard:
@baudehlo:   Username: and Password:   It's for AUTH LOGIN that prompts for each that way with 3xx responses.

smfreegard:
What about auth_proxy?  That was what #430 did.

msimerson:
PR can't be merged as-is, so if that feature is still desired, and someone takes it upon themselves to fix it up, they'll have to open a new PR anyway. So might as well close it, eh?

smfreegard:
(thumbs-up)

PR #1221, opened 2015-10-09
Make plugin/attachment smarter with only one config file and rename…

Dexus:
… it to its use field 'data'

Update docs

### DIFF
--- a/UPGRADE
+++ b/UPGRADE
@@ -2,6 +2,12 @@
 
 v3.0.0
 
+   2015.10.09
+   create config/data.attachment.ini
+   rename plugins/attachment.js to plugins/data.attachment.js
+   delete config/attachment.ctype.regex moved into config/attachment.ini file
+   delete config/attachment.filename.regex moved into config/attachment.ini file
+   
    2015.10.07
 
    deprecated plugins/graph
--- a/config/attachment.ctype.regex
+++ b/config/attachment.ctype.regex
@@ -1,2 +1,0 @@
-executable
-partial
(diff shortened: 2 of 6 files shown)


msimerson:
:+1:

msimerson:
This PR and #1027 are being resurrected and merged as [haraka/haraka-plugin-attachment](https://github.com/haraka/haraka-plugin-attachment), making it easier to maintain a fork that doesn't collide with the @smfreegard version that isn't in Haraka yet.

smfreegard:
Don't worry about me.  I've given up trying to keep my Haraka plugin in step with haraka/Haraka - making attachment a separate plugin won't help that.  I don't see the point of splitting it off like that.

msimerson:
There's two points: one is that this is a place where we had/have conflicting versions, and it's a whole lot easier to `sed .... package.json` to specify a 'local' version, which can be installed from npm or even a private github repo).

Regardless, we want to make it easier to have `local` [versions of] plugins that cannot (due to conflicts, company policy, whatever) be included in Haraka proper.

The second issue is that we need a reference plugin published as an npm module, something we can point plugins authors at and say, "if you want to publish your module on npm, please do it **that** way." In order to permit npm-published plugins to have tests, there's a few things that need to get working:

* config file loading ([haraka-config](https://github.com/haraka/haraka-config)), which is free of Haraka dependencies
* tests/fixtures/* - TODO (probably as haraka-test-fixtures)
* constants.js - TODO (probably as haraka-constants)

Publishing this plugin apart from Haraka is also an exercise in "how little can I get away with doing in order to get npm-published plugin testing to work."

PR #1242, opened 2015-11-15
Clean up plugin loading a tiny bit

baudehlo:
This prevents reading in the contents of package.json as the source code. And removes the useless "use strict" on that code path.

### DIFF
--- a/plugins.js
+++ b/plugins.js
@@ -202,7 +202,7 @@ plugins._load_and_compile_plugin = function (name) {
             continue;
         }
     }
-    if (!(rf || hasPackageJson) {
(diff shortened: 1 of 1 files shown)


plugins.js
```diff
@@ -201,7 +202,7 @@ plugins._load_and_compile_plugin = function (name) {
             continue;
         }
     }
-    if (!rf) {
+    if (!(rf || hasPackageJson) {
```
msimerson: you're missing a closing paren there.

baudehlo:
Fixed.

msimerson:
LGTM

Dexus:
LGTM too.

PR #1245, opened 2015-11-18
Problem: connect.geoip emits non-ASCII codes into mail header

bsmr:
I did some tests with the ```connect.geoip``` plugin, and found this result:
```
X-Haraka-GeoIP: EU, DE, 02, N??rnberg
```
This looks like UTF-8 (the location name should be "Nürnberg") which isn't escaped, and as far as I remember mail headers _must_ only contain 7-bit ASCII.

With the help of ```mimelib``` (and ```encodeMimeWord```) this could be fixed.
But this fix adds the dependency of ```mimelib```, and I am not sure if You like this.

The result looks somewhat akeward, but is 7-bit clean:
```
X-Haraka-GeoIP: EU, DE, 02, =?UTF-8?Q?N=C3=BCrnberg?=
```

_Notes:_
  *  I am no JavaScript or Node.js programmer, so I may have made some mistakes.
  *  The helper ```isAscii7Bit``` is needed, because I couldn't find a better solution.
  *  The code currently runs on a semi-productive server.


deburau:
There is a function encode_qp in utils.js which I used for this.

Am 18.11.2015 um 15:57 schrieb Boris Mühmer:
> With the help of |mimelib| (and |encodeMimeWord|) this could be fixed.
>
> But this fix adds the dependency of |mimelib|, and I am not sure if 
> You like this.
>

msimerson:
Mmmm yes, utils.encode_qp is a better solution, especially since it knows how to wrap long lines, which converting to Quoted Printable is likely to introduce.

bsmr:
I will have a look at it tomorrow.
Thank You for the input... I didn't have a look at ```utils.js```, yet.

smfreegard:
@msimerson @baudehlo Maybe the safest thing here is to run encode_qp in the add_header method?

bsmr:
I removed the ```mimelib``` dependency, and switched to ```utils.encode_qp()```.
The result of my test is ```X-Haraka-GeoIP: EU, DE, 02, N=FCrnberg``` now.

I also (renamed and) moved ```isSevenBitAscii()``` closer to the location where it is needed, but I am not sure if this is a good place.

deburau:
Shouldn't it be

DE, 02, =?ISO-8859-1?Q?N=FCrnberg?=

Or is the 'ü'  displayed correctly for you?

Am 19.11.2015 um 14:26 schrieb Boris Mühmer:

bsmr:
With the _tools_ (Outlook -> Mail Properties -> notepad++) at my destination, I don't see a 'ü'.
My primary goal was to get rid of characters over 127.

Basically I would prefer Your suggestions: I think it is the correct one.

Maybe a "bug" in ```utils.encode_qp()```?

deburau:
A year ago, I tried to solve the same problem but never managed to 
publish my changes.
I have attached what I have done so far, feel free to use it, if it is 
of any help for you. It is working on my private mail server for a year now.

Am 19.11.2015 um 14:51 schrieb Boris Mühmer:

baudehlo:
It's not a bug per-se, the function is/was designed for encoding to QP for
the mail body, not headers. Maybe we could have an options object as last
param that can specify things like that.

Dexus:
As per https://www.ietf.org/rfc/rfc2047.txt the full header like "from:...." need to encoded. Not only a part of the line. And the encoding must included.

Because of the not included info about the encoding outlook will not correct decode the information.

As matt said we can maybe make the function with a optional argument that handle it.

bsmr:
@Dexus That's why I used ```mimelib``` (and ```encodeMimeWord```) at first... well, and not knowing the existance of ```utils.js```... ;-)

I don't know how much unnecessary code lives in ```mimelib```, but I think it is worth using an exisiting library, instead of reinventing the wheel, so to speak... aka DRY.

Dexus:
@bsmr no problem, can you please make a rebase and squash the commits together?
Its better for the overview. After you have done, i will fork your changes and work on a fix for that.

Geetings to Nürnberg from Osnabrück :)

bsmr:
@Dexus Good... I wouldn't have been able to start with it until Saturday evening... It's good You started!

msimerson:
Hey guys, is anyone working on this? I'd like to get this issue resolved and into the forthcoming 2.7.3 release (#1300)

Dexus:
Unfortunately, I am currently busy elsewhere..

baudehlo:
I'd like to make it not use mimelib, and have an extra param (or extra
function) in utils.js.

I'll take a look.

baudehlo:
I've now fixed this with tests in #1401

PR #1247, opened 2015-11-21
ignore ENODATA DNS lookup err

msimerson:
* ignore the 'queryAaaa ENODATA' error.
    * corrects an error pointed out in the comments of #1140 
* better error handling, more DRY by reusing net_utils.get_ips_by_host (#1249)


msimerson:
Also improve similar error handling in `helo.checks` by using utils.

closes #1249

PR #1256, opened 2015-11-26
Update SettingUpOutbound.md

Currerius:
auth requires tls for obvious reasons. auth will not advertise AUTH before STARTTLS.
tls has to be run before auth/flat_file in config/plugins, otherwise neither STARTTLS nor AUTH will be working.
I am not a native speaker so someone might edit my suggestion for the TLS-AUTH explanation in this pull request....


Currerius:
struggled with git :P ... might be ok now ... my first pr ..

PR #1261, opened 2015-12-09
show array when it has length

msimerson:
```js
if (typeof result[key] === 'object') continue;
```

Because Array's are a type of object, that line was inadvertently suppressing the display of arrays with content. The line above makes it obvious Arrays with contents are supposed to be shown.


Dexus:
LGTM

PR #1275, opened 2015-12-22
karma: for sub conn, use same params as results pub

msimerson:
Previously, karma only connected to redis on the default host (localhost). Now karma can subscribe to redis on other hosts to obtain ResultStore publish messages.

### DIFF
--- a/plugins/karma.js
+++ b/plugins/karma.js
@@ -23,6 +23,7 @@ exports.register = function () {
     );
 
     plugin.load_karma_ini();
+    plugin.load_redis_ini();
 
     plugin.register_hook('init_master',  'init_redis_connection');
     plugin.register_hook('init_child',   'init_redis_connection');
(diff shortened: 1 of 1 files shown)


msimerson:
Fixes this "no Redis on localhost (b/c it's not there!) crasher.

````
Dec 21 13:05:20 haraka haraka[80693]: [CRIT] [-] [core] Error: Redis connection to 127.0.0.1:6379 failed - connect ECONNREFUSED 127.0.0.1:6379
Dec 21 13:05:20 haraka haraka[80693]: [CRIT] [-] [core]     at Object.exports._errnoException (util.js:837:11)
Dec 21 13:05:20 haraka haraka[80693]: [CRIT] [-] [core]     at exports._exceptionWithHostPort (util.js:860:20)
Dec 21 13:05:20 haraka haraka[80693]: [CRIT] [-] [core]     at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1060:14)
````

PR #1278, opened 2015-12-24
Fix loading plugins as packages

baudehlo:
The problem was that it might find index.js first,
and then not set hasPackageJson=true. Then it would
try and load via the original method.

We may want to move the index.js line up too...?


msimerson:
I'm wondering about the precedence here. If all these exist, which should load first?  I'm thinking the order should be one of the follows:

# 1

* redis.js
* redis/package.json 
* redis/index.js

# 2

* redis/package.json 
* redis/index.js
* redis.js

I tend to think that 1 seems like the most **expected** behavior.

msimerson:
PS: I use this all the time:  `alias gruh='git remote update haraka && git rebase haraka/master && git push -f'`

baudehlo:
The point is to differentiate between packages vs plugins. I don't see how plugins follow #1.

msimerson:
> The point is to differentiate between packages vs plugins.

That's just semantics. Both are plugins, and both are located in the `plugins` dir. Whether ones a "package" or a "file" is merely a packaging difference. I think the bigger question is what happens if there's a name collision between a "package plugin" and a "file plugin"?

msimerson:
>  both are located in the plugins dir. 

Not quite: `path.join(__dirname, '../node_modules', 'testPlugin','package.json'),`

Hrmm, that could result in some **very** surprising behavior when we have plugins such as redis.js which have a name identical to it's parent module it loads from **node_modules**.

msimerson:
```js
    paths.push(path.join(__dirname, 'plugins'));
    paths.push(path.join(__dirname, 'node_modules'));
```

I'm thinking that code right above is really the source of the surprising behavior I'm foreseeing. Maybe instead of this partial list of possibilities:

````
path.join(__dirname, '../plugins', 'testPlugin','package.json'),
path.join(__dirname, '../plugins', 'testPlugin','index.js'),
path.join(__dirname, '../plugins', 'testPlugin.js'),
path.join(__dirname, '../node_modules', 'testPlugin','package.json'),
path.join(__dirname, '../node_modules', 'testPlugin','index.js'),
path.join(__dirname, '../node_modules', 'testPlugin.js'),
````

It should instead be:

```
path.join(__dirname, '../plugins', 'testPlugin.js'),
path.join(__dirname, '../plugins', 'testPlugin','index.js'),
path.join(__dirname, '../node_modules', 'testPlugin','package.json'),
path.join(__dirname, '../node_modules', 'testPlugin','index.js'),
```

That way, plugins can be plugin.js or plugin/index.js while plugin modules are expected to have "npm" semantics.

**Comment by baudehlo** [comment]
Maybe the code just needs to be more explicit in what it looks for. If it finds a folder, expect it to be a package (with a package.json) period. And load via require.

**Comment by baudehlo** [comment]
Of your 1 and 2 above, a) you shouldn't do that :) But b) If we want it to be 1 we should make sure we don't look for node_modules/redis.js as that would just be wrong.

I pushed in another commit, but it's a work in progress based on this.

**Comment by baudehlo** [comment]
OK, this seems to get you to 1 now.

**Comment by baudehlo** [comment]
OK not quite - I boo-booed.

**Comment by baudehlo** [comment]
Wow it passed finally :)

**Comment by msimerson** [comment]
I think I like this. 

If I'm reading this correctly, when `plugin/path` is a directory, then require a package.json file. Otherwise, we load the plugin from full/path.js. Simple. Predicable. It just happens to break the watch plugin, which now needs a package.json file, right?

**Comment by msimerson** [comment]
Rebased and squashed. `plugins/watch/package.json` added.

**Comment by baudehlo** [comment]
Was `watch` loaded via require() or the older "inject" method?

**Comment by msimerson** [comment]
>  older "inject" method

I know now what this older inject method is/was, but I'm [very very confident](https://github.com/haraka/Haraka/commit/cc0b5533db8b50338cfdaf6889216d1e1227c37f#diff-779ba29ca31786b5ba28f320e69586b2) that watch was loaded with the check for `$dir/index.js`

**Comment by baudehlo** [comment]
The jnject method is the one that puts the constants in the file.

Whereas require is only used if it finds package.json. (Search for hasPackageJson).

I think it might be best to use the inject method for watch and other folders - that way they can have the constants injected.

If you agree, I'll fix this to work that way again.

**Comment by baudehlo** [comment]
OK, this is now up to speed with what I intended. Let me know what you think @msimerson

**Comment by msimerson** [comment]
At first perusal, I'm pretty sure that'll work just fine. I'll test it on a 2.7.2 server later tonight or tomorrow.

**Comment by baudehlo** [comment]
Any joy with this?

**Comment by msimerson** [comment]
Almost there, just built a fresh VM with haraka, testing soon...

**Comment by msimerson** [comment]
In a fresh new VM, with Haraka installed from NPM, I installed this new plugins.js

```sh
cd /usr/local/lib/node_modules/Haraka/ 
[root@haraka /usr/local/lib/node_modules/Haraka]# fetch -o plugins.js https://raw.githubusercontent.com/haraka/Haraka/fix_npm_installs/plugins.js
plugins.js                                    100% of   15 kB  138 MBps 00m00s
# service haraka restart
loglevel: LOGINFO
Starting up Haraka version 2.7.1
[INFO] [-] [core] Loading plugins
[INFO] [-] [core] Loading plugin: process_title
[CRIT] [-] [core] Caught exception: "Loading plugin process_title failed: Error: ENOENT: no such file or directory, open '/usr/local/haraka/plugins/process_title.js'"
```

Hrmm, looking...

msimerson:
I found your logic error/oversight.

msimerson:
The problem is here:

```js
if (!/\bnode_modules\b/.test(pp)) {
```

You end up not loading plugins in the "system" Haraka plugins dir because it has `node_modules` in its path:

````
[INFO] [-] [core] Loading plugin: process_title
[INFO] [-] [core] not dir: /usr/local/haraka/plugins/process_title
[INFO] [-] [core] checking: /usr/local/haraka/plugins/process_title.js
[INFO] [-] [core] not dir: /usr/local/haraka/node_modules/process_title
[INFO] [-] [core] not dir: /usr/local/lib/node_modules/Haraka/plugins/process_title
[INFO] [-] [core] not dir: /usr/local/lib/node_modules/Haraka/node_modules/process_title
[CRIT] [-] [core] Caught exception: "Loading plugin process_title failed: undefined"
````

If I invert the logic and instead search for 'plugins' like this:

```js
if (/\bplugins\b/.test(pp)) {
```

Then it works as I think you intended.

baudehlo:
I kind of feel like again we need to be a bit more strict.

I should document exactly what we need to support (which I need to do on this branch anyway). I'll work on that.

baudehlo:
Situations we need to support, in priority order:

## In Development mode, or install via a plain "git clone":

* Plugin in node_modules. Contains a `package.json` file.
* Plain plugin in `plugins/` folder
* Plugin in a folder in `plugins/<name>/` folder. Contains an index.js with the plugin code. (Do we want to support `plugins/<name>/package.json` here?)

## In "installed" mode (via `haraka -i <path>`):

* Plugin in `<path>/node_modules`. Contains a `package.json` file.
* Plain plugin in `<path>/plugins/` folder
* Plugin in a folder in `<path>/plugins/<name>/`folder. (same concept as above)
* Plugin in `<core_haraka_dir>/node_modules`.
* Core plugin in `<core_haraka_dir>/plugins/` folder
* Plugin in a folder in `<core_haraka_dir>/plugins/<name>/` folder. (same concept as above)

Is that all we need to support?

msimerson:
> Plugin in a folder in plugins/<name>/ folder. Contains an index.js with the plugin code. (Do we want to support plugins/<name>/package.json here?)

Yes.  It's much easier to explain to plugin authors that "plugins are a single file OR must be packaged like npm" than to document and explain our own "special" behavior. In the interests of simplicity, I'd even consider dropping <name>/index and require a package.json, so all plugins are packaged like npm modules. For that matter, making all plugins into directories is also worth considering:

* plugin
    * index.js    <-- plugin
    * README.md   <-- docs
    * config.ini    <-- default config



> Is that all we need to support?

Looks complete to me.

baudehlo:
Pushed up the change related to the comments above. Not done the tests yet so they will fail. Working on that, but it's a complete change to how the tests work.

baudehlo:
@msimerson I'd appreciate any feedback on the implementation before I finish working on the tests.

msimerson:
> paths.push(
>    path.resolve(__dirname, 'node_modules', name, 'package.json'),
>    path.resolve(__dirname, 'plugins', name + '.js'),
>    path.resolve(__dirname, 'plugins', name, 'package.json')
> );

When loading our `redis` plugin, the actual values will be:

````
/usr/local/lib/node_modules/Haraka/node_modules/redis/package.json
/usr/local/lib/node_modules/Haraka/plugins/redis.js
/usr/local/lib/node_modules/Haraka/plugins/redis/package.json
```

Doesn't this namespace collision pose a problem?

````
root@node:~ # head /usr/local/lib/node_modules/Haraka/node_modules/redis/package.json
{
  "name": "redis",
  "version": "2.1.0",
  "description": "Redis client library",
  "keywords": [
    "database",
    "redis"
  ],
  "author": {
    "name": "Matt Ranney",
...
root@node:~ # head /usr/local/lib/node_modules/Haraka/plugins/redis.js
'use strict';

var redis  = require('redis');

    var plugin = this;

    plugin.load_redis_ini();

    plugin.register_hook('init_master',  'init_redis_connection');
````

Part of the old plugin logic that was covered by tests was making sure to load plugins/name.js first. That doesn't "solve" the namespace collisions but it worked for most cases.

This solution should work properly now with plugins/redis.js, but also in the future if we wanted to convert our `plugins/redis.js` plugin to a "package style" plugin that was installed via NPM. One potential solution to avoid namespace collisions is to have a required nomenclature such as `haraka-plugin-$name`.

**Comment by baudehlo** [comment]
Yes, so for core plugins we probably need to make node_modules last in the list. That won't stop custom plugin collisions but I guess that's the best we can do.

**Comment by baudehlo** [comment]
> Whatever solution we adopt now should work properly now with plugins/redis.js, but also in the future if we wanted to convert our plugins/redis.js plugin to a "package style" plugin that was installed via NPM. One potential solution to avoid namespace collisions is to have a required nomenclature such as haraka-plugin-$name.

If we move it to npm we couldn't call it "redis" anyway. We'd *have* to adopt the nomenclature of haraka-*

**Comment by baudehlo** [comment]
I've moved node_modules to last for all cases. In the local install case it makes more sense to check plugins/* first too.

**Comment by baudehlo** [comment]
## Test plan

* Need to test installed mode + core mode
* Need to test each variation of loading plugins
* Need to test conflict on name in various forms
* Need to test plugin not existing
* Need to test plugin existing and failing to compile
* Need to test plugin inheritance
* Plugin timeouts (already tested)

Anything else?

**Comment by msimerson** [comment]
> I've moved node_modules to last for all cases. In the local install case it makes more sense to check plugins/* first too.

Looks good. You might want to update your excellent comments to that affect as well.

Your list of test cases looks quite thorough.

**Comment by baudehlo** [comment]
One more question here: The current code tests and looks for HARAKA_PLUGIN_PATH - which isn't documented anywhere. Do we need to keep supporting that? My current PR doesn't.

**Comment by msimerson** [comment]
HARAKA_PLUGIN_PATH was added [here](https://github.com/haraka/Haraka/commit/6fd3fda344b667659fa5060b6dc30513742535a5#diff-779ba29ca31786b5ba28f320e69586b2). Whether it should be supported? I'd fall on the side of no. By making plugins npm modules, we're providing a better way to install plugins.

**Comment by baudehlo** [comment]
OK, I'm +1 on removing it and left a comment on that commit.

**Comment by baudehlo** [comment]
So I'm *finally* getting around to writing tests for this. Sorry it's taking so long.

@msimerson - Need feedback on what to do about `watch` doing: `require('./utils.js');`. Once it is loaded, the require() used is the core version of require(), which doesn't have any knowledge of `./utils.js`.

My suggestion is maybe a `var utils = plugin.haraka_core_require(...)` which you have to call at runtime. This would make it easier for us to:

- Get rid of the `./name` hack in the long run
- Move haraka core modules into a subfolder

Thoughts?

**Comment by msimerson** [comment]
> My suggestion is maybe a var utils = plugin.haraka_core_require(...) which you have to call at runtime.

+1

But I'd try to make the name less clumsy than plugin.haraka_core_require. Potential names:

Since `plugin` is a global inherited by all plugins:

```js
var utils = plugin.haraka_core_require('utils');  // re-redundant
var utils = plugin.haraka_require('utils');   // redundant
var utils = plugin.require('utils');   // fully qualified
var utils = haraka_require('utils');   // unique
```

It seems obvious enough to me that `plugin.require` is going to do something a little different than a bare `require`.

>  This would make it easier for us to:

* Get rid of the ./name hack in the long run
* Move haraka core modules into a subfolder

+10

msimerson:
> Move haraka core modules into a subfolder

Moving all the core modules into a 'lib' folder is something I'm also keen to see.

baudehlo:
Note that by "plugin" in the previous comment I mean a plugin's "this". We have no other way of easily injecting methods.

msimerson:
>> which you have to call at runtime.

> Note that by "plugin" in the previous comment I mean a plugin's "this".

Sorry, I'm using the term 'global' poorly there. `plugin` or `this` inside a plugin is an instance of Haraka's plugins.js, and thus can have the functions we choose injected, such as the logging functions, and then a new function that loads Haraka "core" libs / modules. In the case of the watch plugin, I was thinking the code would change like this:

```diff
diff --git a/plugins/watch/index.js b/plugins/watch/index.js
index d8f5971..adb576f 100644
--- a/plugins/watch/index.js
+++ b/plugins/watch/index.js
@@ -3,14 +3,13 @@
 
 var util    = require('util');
 
-var utils   = require('./utils');
-
 var wss = { broadcast: function () {} };
 var watchers = 0;
 
 exports.register = function () {
     var plugin = this;
     plugin.inherits('redis');
+    plugin.utils = plugin.require('utils');
 
     plugin.load_watch_ini();
 
@@ -217,7 +216,7 @@ exports.get_connection_results = function (connection) {
         relay      : get_relay(connection),
         helo       : get_helo(connection),
         early      : get_early,
-        queue      : { newval: utils.elapsed(connection.start_time) },
+        queue      : { newval: plugin.utils.elapsed(connection.start_time) },
     };
 
     // see if changed since we last sent
 ```

It's just a matter of choosing a good `require` nomenclature to specify "I want the Haraka version of this library / file / module". This seems sufficiently distinct to me:

```js
var util    = require('util');    // standard node.js / npm require
var utils   = this.require('utils');   // haraka's "plugin" require
var utils   = plugin.require('utils');  // haraka's "plugin" require
```

Perhaps plugin.require is *too* subtle, and plugin.haraka_core('utils') would be better.

baudehlo:
OK, think this is ready to merge. @msimerson ?

msimerson:
Rebased. Squashed. Looks improved. Code is cleaner now. I'll test it today.

msimerson:
Tested. Looks good to me.

I'd really like to cut a 2.7.3 release straight away and then merge this, #1301, and another PR that removes watch to its own repo. Then all of those changes would roll up together in a 2.8.0 version, that also updates the UPGRADING file, noting that the require semantics in plugins has changed. This is a change that **may** harm some third party plugins.

smfreegard:
> I'd really like to cut a 2.7.3 release straight away and then merge this, #1301, and another PR that removes watch to its own repo. Then all of those changes would roll up together in a 2.8.0 version, that also updates the UPGRADING file, noting that the require semantics in plugins has changed. This is a change that may harm some third party plugins.

+1

baudehlo:
The require changes only happen in folder plugins (node_modules or
plugins/foo/package.json) btw. We only have 1 of those currently - watch. I
doubt anyone else is using that feature yet.

baudehlo:
Also note that moving watch out will break the tests - we need at least one
core plugins/folder/package.json plugin to test with :)

baudehlo:
I need to write docs first. We never properly documented where things get loaded from.

baudehlo:
Docs done. I also simplified the code for loading plugins as modules.

msimerson:
Rebased again and squashed. LGTM. The local branch history is kinda messy. If you continue working on this branch, you can refresh against this cleaned up branch with:

```sh
git checkout master
git branch -D fix_npm_installs
git remote update
git checkout origin/fix_npm_installs
git checkout -b fix_npm_installs
# continue changes
```

or, rebase and squash your local branch:

```sh
git rebase origin/master
# fix the conflicts
git rebase --continue
# and then squash
git rebase -i origin/master
git push -f origin/fix_npm_installs
```

baudehlo:
I'm done with it now - no more changes :)

PR #1295, opened 2016-01-10
Allow pass notes to send_email

### LINKED ISSUE #1293: Transaction.notes not passing to Hmail.todo.notes
We are creating a plugin that changes the SMTP server based in some info from an API. So we plan to get the info in the `queue` event, and store the SMTP server in the transaction.notes and then use it in the `get_mex` hook.

Look up this plugin code
````js
var outbound = require('./outbound');

exports.register = function() {
  this.register_hook('queue', 'send_email');
  this.register_hook('get_mx', 'change_smtp_server');
};


exports.send_email = function (next, connection) {

  var contents = [
    "From: " + "ricardo.polo@tulpep.com",
    "To: " + "ricardo.polo@tulpep.com",
    "MIME-Version: 1.0",
    "Content-type: text/plain; charset=us-ascii",
    "Subject: Some subject here",
    "",
    "Some email body here",
    ""].join("\n");

  outbound.send_email(connection.transaction.mail_from, connection.transaction.rcpt_to, contents, function (code, msg) {
    connection.transaction.notes.send_server = "server.com";
    connection.transaction.notes.send_server_port = 25;
    switch (code) {
      case OK:    return next(OK, msg);
      case DENY:  return next(DENY, msg);
      default:    return next(DENYSOFT, msg);
    }
  });

};

exports.change_smtp_server =  function (next, hmail, domain) {
  var mx = {
      priority: 0,
      exchange: hmail.todo.notes.send_server,
      port: hmail.todo.notes.send_server_port
  };
  return next(OK, mx);
};
````

It is supposed acording to docs that everything in `transaction.notes` will be in `hmail.todo.notes`, but in this case `todo.notes.send_server` is empty. Why could this be happening? 

Any idea, maybe @baudehlo @smfreegard @msimerson ?
Thank you so much for you help!!

ricardopolo:
When you create a new email using `send_email()` it creates a brand new transaction.
If you need some information in that transaction like notes filled in another hooks you cannot access to them.

This PR allows to pass a new option the `send_email()` to preserve the notes. It just adds another option to the options object that `send_email()` already receives.

Includes updated docs.
Fixes #1293

### DIFF
--- a/docs/Outbound.md
+++ b/docs/Outbound.md
@@ -283,3 +283,9 @@ see https://tools.ietf.org/html/rfc2821#section-4.5.2), you should pass the
 ```dot_stuffed: true``` option, like so:
     
     outbound.send_email(from, to, contents, outnext, { dot_stuffed: true });
+
+
+In case you need notes in the new transaction that `send_email()` creates, you should pass the
+```notes``` option, like so:
+
+    outbound.send_email(from, to, contents, outnext, { notes: transaction.notes });
--- a/outbound.js
+++ b/outbound.js
@@ -357,12 +357,18 @@ exports.send_email = function () {
     var options = arguments[4];
 
     var dot_stuffed = ((options && options.dot_stuffed) ? options.dot_stuffed : false);
+    var notes = ((options && options.notes) ? options.notes : null);
 
     this.loginfo("Sending email via params");
 
     var transaction = trans.createTransaction();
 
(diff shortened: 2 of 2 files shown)


msimerson:
+1

Absent any further discussion (here or in #1293), I'll merge this after 2015-01-21.

smfreegard:
+1 - seems sensible.  I meant to merge it the other day, but forgot.

PR #1296, opened 2016-01-13
Feature/improved exit logging

pvandijk:
As someone completely new to Haraka, i noticed that the console logging was partially broken on my environment (centos/node 5.4), such that error messages from the logger were being discarded as process.exit would simply dump the buffers. 
Instead of seeing an error for "Failed to setup listeners: listen EACCES" for port 25, i was left with debug info about connect hooks.

Hopefully this change will make the first time experience for people trying Haraka out a little easier.

Similarly if anything is out of line here please let me know as this is the first PR i've made for Haraka.

This PR changes a few small things:
- Cut over the existing process.exit calls to use exit() from the exit module, which addresses the buffering issue with the logger
- Remove the exit parameter from logger.dump_logs - i found it a bit strange that a logging module would be responsible for process termination

### DIFF
--- a/haraka.js
+++ b/haraka.js
@@ -3,6 +3,7 @@
 'use strict';
 
 var path = require('path');
+var exit = require('exit');
 
 // this must be set before "server.js" is loaded
 process.env.HARAKA = process.env.HARAKA || path.resolve('.');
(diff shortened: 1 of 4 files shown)


Dexus:
I am opposed to this amendment because it is an unnecessary dependency.
This change has absolutely no improvement.

msimerson:
Hi @pvandijk, thanks for this PR. 

In general, this PR is an improvement. My first reaction was the same as @Dexus, that I didn't like the introduction of another dependency. I'm in the habit of putting all my `process.exit` calls inside a `cleanupAndExit` function. However, after looking over the `exit` module, the single function it introduces, and the cast of characters that also depend on it, I've changed my mind and I'm in favor of adding the dependency and merging this PR as-is.

smfreegard:
+1 from me too.  It fixes a problem that has bitten me numerous times before.

msimerson:
Hey @pvandijk, if you'll [rebase](https://github.com/haraka/Haraka/wiki/Contributing) this I'll go ahead and merge.

pvandijk:
Thanks for the feedback - it's now squashed.

PR #1299, opened 2016-01-15
logger: ignore empty/null/false arguments

### LINKED ISSUE #1298: logger broken on null parameter since 2.7.0
The following crashes:
```
var data = null:
this.logdebug('Log my data: ', data);
```

`[CRIT] [-] [core] TypeError: Cannot read property 'name' of null`

Issue is here:
https://github.com/haraka/Haraka/blob/a522bfc08a9bead6def21e3154f48bab1f252862/logger.js#L190

Introduced by https://github.com/haraka/Haraka/commit/d66974b711969448bebcce95eb603a8138ea47ce

... because `typeof null` returns `'object'`.

`if (data && data.name)` should be enough.

msimerson:
* skip processing if typeof data !== object && !data
    * handles `null` case
    * guards against future false-like values that happen to be an object
* remove redundant typeof check

fixes #1298

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -171,10 +171,11 @@ logger.log_if_level = function (level, key, plugin) {
         var pluginstr = '[' + (plugin || 'core') + ']';
         for (var i=0; i < arguments.length; i++) {
             var data = arguments[i];
-            if (typeof(data) !== 'object') {
+            if (typeof data !== 'object') {
                 str += data;
                 continue;
             }
+            if (!data) continue;
 
             // if the object is a connection, add the connection id
             if (data instanceof connection.Connection) {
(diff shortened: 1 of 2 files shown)


logger.js
```diff
@@ -171,7 +171,8 @@ logger.log_if_level = function (level, key, plugin) {
         var pluginstr = '[' + (plugin || 'core') + ']';
         for (var i=0; i < arguments.length; i++) {
             var data = arguments[i];
-            if (typeof(data) !== 'object') {
+            if (!data) continue;
```
baudehlo: Probably wants to be if (data == null) in case someone wants to log a plain 0.

msimerson:
And now I've added test cases too. :-)

baudehlo:
+1

PR #1301, opened 2016-01-16
remove TLD stuff to haraka-tld

### LINKED ISSUE #1269: Create an --upgrade/-u flag to bin/haraka
We need to add a --upgrade/-u flag to bin/haraka that copies in the new:
```
 config/extra-tlds
 config/public-suffix-list
 config/three-level-tlds
 config/top-level-tlds
 config/two-level-tlds
```
files.

It would require the -c flag for the config directory to be used.

msimerson:
# Changes

* removes net_utils.* TLD functions to [haraka-tld](https://github.com/haraka/haraka-tld)

# Benefits

* removes [maintenance of TLD files](https://github.com/haraka/Haraka/wiki/Release%20Process) to external package
* removes TLD files from haraka/config, so that 'npm install' or 'npm upgrade' automatically brings in the latest versions
* reduces the size of Haraka (8 additions and 21,272 deletions)
* haraka-tld has excellent test coverage (> 90%)
* tests for this package pass in 15 seconds.
* as we remove stuff like this from Haraka core, our main test suite will complete faster.

fixes #1269


baudehlo:
Depends on #1278 I assume?

msimerson:
Maybe. The plugins that require it try finding it Haraka/etc instead of Haraka/node_modules/haraka-tld/etc, likely because of our custom require.

baudehlo:
Try a full install. It won't even load. That's why I started working on it.

msimerson:
> Try a full install. It won't even load. That's why I started working on it.

I'm not sure what you mean. A full install of what won't load what?

baudehlo:
As in using bin/haraka -i and installing it there. I tried that with another module I wrote and it just doesn't work currently on master.

msimerson:
Oh, I see. I haven't tried that. I now have [a script](https://github.com/msimerson/Mail-Toaster-6/blob/master/provision-haraka.sh) that provisions a fresh OS in a container and then does a clean Haraka install, applies all my config settings, and then runs some `swaks` tests to make sure it works.

msimerson:
This PR is updated and now passes tests. The change I had to make so that these functions could find their config files is in haraka/haraka-tld#3. Basically, check to see where the install dir is and load the files from there.

This solution works for core module like this package. This obviously isn't a good enough solution for #1331. I plan to test this soon and absent any further discussion/objections, get this merged in anticipation of 2.8.0.

msimerson:
Finally got around to testing. Works as expected.

Issue #1307, opened 2016-01-19
smtp_forward/smtp_proxy - hook_queue and hook_queue_outbound


smfreegard:
smtp_forward and smtp_proxy work differently regarding outbound mail.   The difference is my fault looking at 'Blame'.   smtp_forward sets `exports.hook_queue_outbound = exports.hook_queue` whereas smtp_proxy does not.

My reasoning for smtp_forward was because I wanted all mail to go to the forwarder - I didn't want outbound to handle it at all.

We've had a number of people hit the same issue with smtp_proxy and I just dealt with another one today on IRC.   They expect queue/smtp_proxy to proxy irrespective of queue .vs. queue_outbound.

Either we need to add a common configuration option to enable the plugins to handle both hooks or we should add a config item to smtp.ini to do it globally across all queue/* plugins.

Thoughts?

msimerson:
>  They expect queue/smtp_proxy to proxy irrespective of queue .vs. queue_outbound.

I agree with them, I think that *should* be the default.

I'd suggest having them both hook queue_outbound by default (that whole, POLA thing) and then having an option in smtp_forward.ini and smtp_proxy.ini that..... blah, it took less time to create a PR (#1308) than to explain with English.

PR #1308, opened 2016-01-19
add queue_outbound config option

msimerson:
This is my answer to #1307, reflecting the way I think it ought to be.

Since this would change the default behavior of the smtp_proxy plugin, it may not be proper to make the code do this right away, and until a future version bump, perhaps smtp_proxy.enable_outbound should still default default to false?

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,7 +1,11 @@
 
 2.N.N - Feb NN, 2016
 
-2.7.3 - Jan 16, 2016
+2.7.3 - Jan 22, 2016
+
+* Changes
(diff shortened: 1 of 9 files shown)


smfreegard:
LGTM - what about qmail-queue?  (I don't use it or know anything about it - but should it have similar behaviour?).

baudehlo:
Probably.

msimerson:
> what about qmail-queue? (I don't use it or know anything about it - but should it have similar behaviour?).

Hrmm, likely so. If one was using the qmail_queue plugin, then they'd almost certainly expect that. I just added another commit doing that.

PR #1310, opened 2016-01-21
New configuration option for not removing carriage return

ricardopolo:
Currently Haraka removes the `\r` in the lines processed.

This PR allows to have an option to disable this feature, so the lines remain as they come in the SMTP message.

The default behavior was not changed. 
The PR includes the docs updated.


msimerson:
> Currently Haraka removes the \r in the lines processed.

Those \r are actually a \r\n and have a special name `CRLF` in SMTP parlance. This variable should probably be named `strip_crlf` or something to that effect.

transaction.js
```diff
@@ -10,6 +10,7 @@ var MessageStream = require('./messagestream');
 var trans = exports;
 
 var MAX_HEADER_LINES = config.get('max_header_lines') || 1000;
+var REMOVE_CARRIAGE_RETURN = config.get('remove_carriage_return') || 'true';
```
msimerson: What happens here if the file `remove_carriage_return` contains `false`?

Some test coverage of this might be a good idea.

msimerson: Also, you're setting REMOVE_CARRIAGE_RETURN='true' (a string that contains 'true'), which really really really ought to be the boolean `true` instead.

ricardopolo: I don't use a boolean, because I read the `false` and `true` from a config file, so it comes as string.

If I use  `if (REMOVE_CARRIAGE_RETURN )` it will pass if the string has any content, not just `true`. Thats why I use  `if (REMOVE_CARRIAGE_RETURN === 'true')`

So, any adivice about how to convert that value from the config to a boolean? 
I tried  `var REMOVE_CARRIAGE_RETURN  = Boolean(config.get('remove_carriage_return');)`
 with not success, it always returns `true`.

I'll add tests
I'll rename the variable.

baudehlo: This should be in smtp.ini (where you can define what is a Boolean)

msimerson: This is what the code in transaction.js might look like, with the boolean declaration in smtp.ini:

```diff
diff --git a/transaction.js b/transaction.js
index 75cd382..9933480 100644
--- a/transaction.js
+++ b/transaction.js
@@ -46,7 +46,11 @@ exports.createTransaction = function(uuid) {
     t.uuid = uuid || utils.uuid();
     // Initialize MessageStream here to pass in the UUID
     t.message_stream = new MessageStream(
-            config.get('smtp.ini'), t.uuid, t.header.header_list);
+        config.get('smtp.ini', {
+            booleans: [
+                '+main.strip_crlf',
+            ]
+        }), t.uuid, t.header.header_list);
     return t;
 };
 
```

The leading + is what defines the boolean as true, unless it has been declared false in the config file.

ricardopolo:
I found that using `line_endings: \n` option of `message_stream` you can fullfill this requirement.

In fact you even have a code specifically to remove the `\r` (strip the CR) from the stream.
https://github.com/haraka/Haraka/blob/master/messagestream.js#L285-L293

So this is a basically a unnecessary thing. I closing the PR.

PR #1316, opened 2016-01-23
Now passing options parameter to MessageStream get_data

ricardopolo:
There is a bug in `get-data` of `messagestram`.

Currently if you pass the options parameter they are ignored. 

You can confirm it here, and test the pruposed fix https://repl.it/BgLr/0

So this PR fixes it. Tested for passing just the callback and the options and callback.

### DIFF
--- a/messagestream.js
+++ b/messagestream.js
@@ -403,11 +403,10 @@ MessageStream.prototype.destroy = function () {
     }
 };
 
-MessageStream.prototype.get_data = function (cb) { // Or: (options, cb)
-    var options = {};
-    if (arguments.length === 2) {
-        cb = arguments[1];
-        options = arguments[0];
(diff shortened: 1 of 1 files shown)


ricardopolo:
@msimerson any comments about this?

msimerson:
Your change looks good. I have to admit I don't understand *why* the current code there doesn't work, and before reviewing your proposed code, I couldn't see the bug. Now I can, but I still don't know why, and that annoys me. I'd much rather see test coverage added to cover this, but if you get this rebased (post a comment after), I'll merge it, even without tests.

msimerson:
@ricardopolo, still waiting for you to [rebase](https://github.com/haraka/Haraka/wiki/Contributing#rebase) this so it can get merged.

ricardopolo:
@msimerson Sorry for the delay, tomorrow I'll create the test

PR #1319, opened 2016-01-24
Added RabbitMQ Queue

thisboyiscrazy:
Modified rabbitmq_ampqlib to allow for different rabbitmq queues based to regex replaces in rctp_to

### DIFF
--- a/config/rabbitmq_amqplib_queue.json
+++ b/config/rabbitmq_amqplib_queue.json
@@ -5,7 +5,7 @@
 	"password":"guest",
 	"exchangeName":"email_messages",
 	"exchangeType":"direct",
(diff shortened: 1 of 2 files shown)


msimerson:
How does this plugin compare to the other two RabbitMQ plugins in the plugins/queue directory? Maybe the docs for all 3 of the RabbitMQ plugins should have a section that explains how the plugins differ and how they're alike?

thisboyiscrazy:
I will clearify that. The other onse will only put the messages in one
queue in rabbitmq. This one can put the messages in different ones based on
rctp.

thisboyiscrazy:
I found some coding issue i will rix and add documentation.

PR #1321, opened 2016-01-25
rabbitmq_ampqlib_queue

thisboyiscrazy:
Allows messages to be delivered to multiple queues based on rctp_to


plugins/queue/rabbitmq_amqplib_queue.js
```diff
@@ -0,0 +1,102 @@
+// queue/rabbitmq_amqplib
+
+var amqp = require("amqplib/callback_api");
+
+var channel;
+var deliveryMode;
+var logger = require('./logger');
```
msimerson: logger is unused and should be removed.

msimerson:
My gut feel is that instead of being a new plugin, this would be better as a feature addition of the existing `queue/rabbitmq_ampqlib` plugin. Add a config setting to rabbirmq_amqplib.ini and when enabled, it registers the *data* and *queue_outbound* hooks, enabling this feature.

msimerson:
If you have objections to integrating this with the existing plugin, then this PR really should include a paragraph in both plugins explaining why they exist separately, with references to each other and guidance for choosing the most appropriate one.

thisboyiscrazy:
That's is what i was originally going, I will do that.

msimerson:
closing, in expectation of a future PR that merges this with the existing rabbitmq plugin.

PR #1330, opened 2016-02-05
Call next only after message_stream ended

### LINKED ISSUE #1163: "Error: Cannot pipe while currently piping" with lmtp
I have a plugin to forward mail to [archiveopteryx](archiveopteryx.org) via lmtp:

```javascript
// queue through lmtp if the message is marked to_aox
"use strict";

var outbound = require('./outbound');

exports.hook_queue = function(next, connection) {
    if (connection.transaction.notes.to_aox) {
        this.logdebug('to_aox is true, sending output with using_lmtp=true');
        connection.transaction.notes.using_lmtp = true;
        outbound.send_email(connection.transaction, next);
    } else {
        this.logdebug('to_aox is false, ignoring');
        next();
    }
}

exports.hook_get_mx = function (next, hmail, domain) {
    if (!hmail.todo.notes.using_lmtp) return next();
    var config = this.config.get('archiveopteryx.ini');
    var section = config['lmtp'] || config.main;
    this.loginfo('lmtp port: ' + section.port);
    var mx = {
        priority: 0,
        exchange: section.host || '127.0.0.1',
        port: section.port || 24,
        using_lmtp: true
    };
    return next(OK, mx);
}
```

debug log: https://gist.github.com/tino/675a2e0b5dd1cf71e647  (replaced the domain with mydomain.io)
The list of enabled plugins is also shown there.

Any way I can debug this?

chazomaticus:
Fixes #1163.  Patch from @baudehlo.

With the way the plugin was written before, it was possible that a later plugin would also call connection.transaction.message_stream.pipe() before dkim_verify's pipe had ended, leading to an error.  This moves the next() call out of the DKIM callback and into the end handler on the stream to ensure that'll never happen.

I've been running this patch for a while and it does seem to fix the crash, but I haven't had time to dig in and verify it's behaving correctly everywhere.

### DIFF
--- a/plugins/dkim_verify.js
+++ b/plugins/dkim_verify.js
@@ -19,7 +19,7 @@ exports.hook_data_post = function(next, connection) {
         if (err) {
             connection.logerror(self, 'error=' + err);
         }
-        if (!results) return next();
+        if (!results) return;
         results.forEach(function (res) {
             connection.auth_results(
               'dkim=' + res.result +
@@ -49,7 +49,7 @@ exports.hook_data_post = function(next, connection) {
         connection.logdebug(self, JSON.stringify(results));
         // Store results for other plugins
         txn.notes.dkim_results = results;
-        return next();
     }, ((plugin.timeout) ? plugin.timeout - 1 : 0));
+    txn.message_stream.once('end', next);
     txn.message_stream.pipe(verifier, { line_endings: '\r\n' });
 };


baudehlo:
+1

Please bring it up to date with master (you can do this in the Github UI now).

msimerson:
> Please bring it up to date with master 

never mind, I'll do it, after I land a couple other PRs in progress...

baudehlo:
Shame that didn't get into 2.7.3. Guess we should move to 2.8 asap

chazomaticus:
Thanks y'all!  <3

PR #1333, opened 2016-02-10
Haraka now takes a list of backend hosts to forward to

kgoess:
The original config for Haraka took a single host and port in the
smtp_forward.ini config to connect to. If that host was a CNAME with a
bunch of possible hosts it could resolve to, Haraka would have no way of
knowing if say just one of the hosts was down.

Now we can have the pool listed explicitly in the config like this:

    forwarding_host_pool=1.1.1.1:1111, 2.2.2.2:2222, 3.3.3.3.:3333....

If that's not in the config, then smtp_client will continue to use the
old host and port settings.

When a host in the pool goes down, we remove it from the pool of available
hosts and we set a timer to check it periodically to see if it looks like
it's come back online.

The first attempted delivery to a failed backend will result in the mail
being rejected back to the originating MTA, who is responsible for
resending it to Haraka.

### DIFF
--- a/host_pool.js
+++ b/host_pool.js
@@ -1,0 +1,174 @@
+"use strict";
+
+var net = require('net');
+
+/* HostPool:
+ *
+ * Starts with a pool of backend hosts from a "forwarding_host_pool"
+ * configuration that looks like this:
+ *
+ *      1.1.1.1:11, 2.2.2.2:22, 3.3.3.3:33
+ *
+ * It randomizes the list and then gives then out sequentially (for
+ * predictability).
+ *
+ * If failed() is called with one of the hosts, we mark it down for retry_secs
+ * and don't give it out again until that period has passed.
+ *
+ * If *all* the hosts have been marked down, we ignore the marks and just give
+ * out the next host. That's too keep some random short-lived but widespread
+ * network failure from taking the whole system down.
+ */
+
+var logger = require('./logger');
+var utils  = require('./utils');
+
+// takes a comma/space-separated list if ip:ports
+//  1.1.1.1:22,  3.3.3.3:44
+function HostPool(hostports_str, retry_secs){
+    var self = this;
+
+    var hosts = (hostports_str || '')
+            .trim()
+            .split(/[\s,]+/)
+            .map(function(hostport){
+                var splithost = hostport.split(/:/);
+                return { host: splithost[0],
+                          port: splithost[1]
+                        };
+            });
+    self.hostports_str = hostports_str;
+    self.hosts = utils.shuffle(hosts);
+    self.dead_hosts = {};  // hostport => true/false
+    self.last_i = 0;  // the last one we checked
+    self.retry_secs = retry_secs || 10;
+}
+
+
+/* failed
+ *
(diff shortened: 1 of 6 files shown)


msimerson:
Hi @kgoess , this looks very good. If you rebase against haraka/master and then force push this branch, it should now pass the integration tests.

One thing that might be worth changing is the way the host pool is configured. A method that already has widespread use is to specify an array for the host value instead of a string. This would necessitate extending our ini parser to recognize the array syntax. 

Using the array syntax for the host pool would permit it to also be used with the domain specific forward routes.

host_pool.js
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * If *all* the hosts have been marked down, we ignore the marks and just give
+ * out the next host. That's too keep some random short-lived but widespread
+ * network failure from taking the whole system down.
+ */
+
+var logger = require('./logger');
+var utils  = require('./utils');
+
+// takes a comma/space-separated list if ip:ports
```
msimerson: s/if/of/

host_pool.js
```diff
@@ -0,0 +1,174 @@
...
+    var self = this;
+
+    var hosts = (hostports_str || '')
+            .trim()
+            .split(/[\s,]+/)
+            .map(function(hostport){
+                var splithost = hostport.split(/:/);
+                return { host: splithost[0],
+                          port: splithost[1]
+                        };
```
msimerson: Since Haraka is a MTA, having a default port 25 (when unspecified) makes the config syntax a little more friendly.

kgoess:
I'm working on integrating the .ini arrays into this pull request, and defaulting the port to 25.

msimerson:
Pinging Kevin. This is a nice change and I'd like to get this merged.

kgoess:
Oops, sorry, I've done the work, need to merge it here, will get back on that...

baudehlo:
@kgoess ? Holding up 2.8 now...

baudehlo:
I suggest if we don't hear from Kevin on this before the end of this weekend (2016-04-10) we move forwards with 2.8 without this PR. Any objections?

msimerson:
> Any objections?

We don't run out of version numbers at 2.8.0. The only reason **I** haven't pushed 2.8 hard yet is that I haven't finished the first "haraka module as an npm plugin" yet. I got 90% done and then ran splat into a time sink (buying a house) and haven't had enough spare cycles to get back to it.

Buuuuuttttt, now that [npm changed](https://medium.com/@mbostock/prereleases-and-npm-e778fc5e2420#.v590k8m55) and our alpha version is getting installed anyway, and we've gotten quite a bit of accidental testing, we may as well just release it.

baudehlo:
Let's give Kevin until Sunday. If no movement, I say release.

kgoess:
All righty, suggestions taken (and appreciated). CI tests are passing. Take a look at that. But I'm getting on a plane tonight to leave town for a week, if there's any problems or question I won't see them until next weekend. Cheers!

baudehlo:
The one thing this now lacks is docs.

msimerson:
Merging. Will get testing in 2.8.0.beta.1

PR #1334, opened 2016-02-10
fix out-of-range errors from banner insertion

kgoess:
We're seeing these occasionally in production logs:

    RangeError: out of range index
        at RangeError (native)
        at insert_banner (mailbody.js:189:13)

It looks like an empty html part results in an insert-at index of -1, which
won't work with buf.copy. So fix it by returning 0 in that case.

That means that an empty html part will now be just the banner. That's the
current behavior for the text part though. And if the only thing we have is an
html part, I think we want to show the banner.

### DIFF
--- a/mailbody.js
+++ b/mailbody.js
@@ -129,6 +129,13 @@ Body.prototype.parse_start = function (line) {
 };
 
 function _get_html_insert_position (buf) {
+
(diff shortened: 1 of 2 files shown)


msimerson:
Looks good. Thanks for this, and the test coverage.

PR #1335, opened 2016-02-10
Load plugin config from own folder and merge

baudehlo:
Let me know what you think of these changes.

### DIFF
--- a/config.js
+++ b/config.js
@@ -9,10 +9,17 @@ module.exports = new Config();
 function Config (path) {
     this.root_path = path || cfreader.config_path;
     this.module_config = function (defaults_path, overrides_path) {
+        var path = require('path'); // This can be called somehow before "path" at file top is loaded??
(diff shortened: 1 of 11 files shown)


baudehlo:
Tests added. Code fixed. Pretty much ready for this to be peer reviewed.

One weakness is if the plugin does `require('./config')` and uses that.

We can probably fix that, because that version of `require()` is already shimmed. What do people think?

baudehlo:
Added docs.

baudehlo:
Made require(`./config`) work the same as plugin.config does.

baudehlo:
Objections to merging this so we can all start playing with it?

smfreegard:
+1 - Can't see anything obviously wrong.  Sweet solution!

Dexus:
:+1: looks good to me. 

> Sweet solution!
Agree :)

PR #1337, opened 2016-02-13
Allow original email's Subject to be included in bounce message

davebeyer:
Add the original email's Subject to the set of values that can be used in the config/outbound.bounce_message template, for more informative bounce messages.

### DIFF
--- a/outbound.js
+++ b/outbound.js
@@ -1709,6 +1709,7 @@ HMailItem.prototype.populate_bounce_message_with_headers = function(from, to, re
         me:   config.get('me'),
         from: from,
         to:   to,
+        subject: header.get_decoded('Subject').trim(),
         recipients: this.todo.rcpt_to.join(', '),
         reason: reason,
         extended_reason: this.todo.rcpt_to.map(function (recip) {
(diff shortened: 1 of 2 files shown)


outbound.js
```diff
@@ -1709,6 +1709,7 @@ HMailItem.prototype.populate_bounce_message_with_headers = function(from, to, re
         me:   config.get('me'),
         from: from,
         to:   to,
+        subject: header.get('Subject').trim(),
```
baudehlo: This should use `.get_decoded()` otherwise it's in the original format.

davebeyer:
@baudehlo, thanks for the tip.  I made the change but unfortunately it now seems to be failing one of the integration checks.  (Is there a way for me to see what check is failing?)    After this is resolved, if the change seems reasonable to you, I'm happy to rebase to clean it up.

baudehlo:
Seems to be a simple deprecation problem in our eslint settings. @msimerson can you figure out what the issue is?

msimerson:
Okay, HEAD is updated for compatibility with eslint 2. If you rebase against haraka/master, the integration tests should now pass.

davebeyer:
Thanks @msimerson - I've rebased and integration tests now pass.

msimerson:
+1

Dexus:
:+1:  thanks

Issue #1324, opened 2016-02-01
Don't require TLS certificate pair when acting as an SMTP client.


smfreegard:
Currently we require a tls_[key|cert].pem before we'll send STARTTLS and switch to a secure socket. This is unnecessary when we're acting as a client.

If we have a tls_[key|cert].pem, then we can present it as a client certificate, but it shouldn't be required for TLS to be attempted.

baudehlo:
I'm reading the docs for tls.createSecureContext and it says 'key' is required. I don't know how we achieve this. Any ideas?

Dexus:
We can use PEM and create one... Or some other module...

Von meinem iPhone gesendet

smfreegard:
@baudehlo: it would seem that you can call it without any arguments based on the source:
https://github.com/nodejs/node/blob/master/lib/_tls_legacy.js#L688

As a side note - we could completely do away with our TLS cludges by dropping node 0.10.x support completely as we can then do:

var sock = net.connect(....);
tls.connect({socket: sock}, .... {
   ....
});

Which will update the SMTP client socket to TLS.   On the server end you would use new tls.TLSSocket() to upgrade.

celesteking:
yes dropping stale node versions is a good idea.
Also, what is outbound_load_tls_ini branch about?

msimerson:
Just yesterday I was thinking it's about time to drop 0.10, so I'm +1 on that.

baudehlo:
My only problem is moving up on boxes I control has proved hard due to
requiring a newer compiler than ships. e.g. debian 7.7 won't even compile
node 0.12.

baudehlo:
Also kludges aside - they work and are stable. I'd hate to change code just for the sake of it.

msimerson:
Who is changing code for the sake of it? Switching to `tls.TLSSocket` (#1099) was a response to past issues with our current TLS. It's a better API. The main reason we haven't done it yet (besides a shortage of round tuits), is that it requires dropping node 0.10. But dropping 0.10 also enables #1323, which is also a very nice improvement.

What's preventing you from upgrading your boxes to Debian 8?

baudehlo:
Laziness and fear. I don't have a second box to test an upgrade on, so fear
of downtime while I emergency fix things. Emailitin is now processing over
140k attachments a month for people, so it has become a pretty important
service for them.

msimerson:
>  I don't have a second box to test an upgrade on

[...]

PR #1340, opened 2016-02-16
Outbound load tls.ini

### LINKED ISSUE #1054: TLS in outbound: make it configurable
```
outbound	Looking up A records for: j.mx.utoronto.ca
outbound	Ongoing connection failed to 128.100.132.12:25 : Error: 140665633523680:error:14082174:SSL routines:SSL3_CHECK_CERT_AND_ALGORITHM:dh key too small:s3_clnt.c:3331:
```

We need to:
* Have configurable list of TLS disabled hosts/subnets (like in tls.ini?)
* Maybe don't screw the session if host didn't negotiate TLS, but fallback to plain text (should be configurable)
* add DH and cipher params, possibly with per-host/subnet option

baudehlo:
This fixes #1054 (mostly)

### DIFF
--- a/outbound.js
+++ b/outbound.js
@@ -14,6 +14,7 @@ var config      = require('./config');
 var constants   = require('./constants');
 var trans       = require('./transaction');
 var plugins     = require('./plugins');
+var tls_socket  = require('./tls_socket');
 var async       = require('async');
 var Address     = require('./address').Address;
 var TimerQueue  = require('./timer_queue');
@@ -41,7 +42,7 @@ exports.load_config = function () {
         booleans: [
             '-disabled',
             '-always_split',
-            '-enable_tls',    // TODO: default to enabled in Haraka 3.0
+            '+enable_tls',
             '-ipv6_enabled',
         ],
     }, function () {
@@ -1096,6 +1097,8 @@ HMailItem.prototype.try_deliver_host_on_socket = function (mx, host, port, socke
         "auth": [],
     };
 
+    var tls_config = tls_socket.load_tls_ini();
+
     var send_command = function (cmd, data) {
         if (!socket.writable) {
             self.logerror("Socket writability went away");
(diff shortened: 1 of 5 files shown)


msimerson:
I haven't tested but this *looks* okay. Is it ready to merge @baudehlo ?

baudehlo:
My only concern was it doesn't address the fact that outbound doesn't need
a tls cert, but it mandates one. That's a bug IMHO.

baudehlo:
This doesn't fix #1324 but I don't really see how that would work, since `tls.createSecureContext()` requires a `key` param. Not sure how we'd fix that while supporting all the other params - would love to know.

Perhaps additional params would only work if you provided a key?

msimerson:
Not sure, will look later. But, a workaround that springs to mind is that if the crt/key isn't provided, auto-generate a self-signed one at startup.

baudehlo:
OK, fixed this issue too. Tested the code. No key works fine. Should probably do the same thing with smtp_client.js (including loading tls.ini). But I don't use that code so someone else might want to patch it.

baudehlo:
Just realized this ignores no_tls_hosts currently. Working on that.

msimerson:
Rebased, squashed, and merging. I'm going to publish a 2.8.0-alpha1 release (easier to install & test) and test deploy.

outbound.js
```diff
@@ -1144,7 +1147,12 @@ HMailItem.prototype.try_deliver_host_on_socket = function (mx, host, port, socke
         }
 
         // TLS
-        if (smtp_properties.tls && cfg.enable_tls && !secured) {
+        if (!(self.todo.domain in tls_config.no_tls_hosts) &&
+            !(host in tls_config.no_tls_hosts) &&
+            smtp_properties.tls && cfg.enable_tls && !secured)
+        {
+            console.log(tls_config.no_tls_hosts);
```
celesteking: how production-ready

msimerson: Please *read* the messages. It sure **looks** like @baudehlo is still working on it. You'll also notice that it's in an -alpha release version, for early testing. Who was silly enough to claim production ready?

baudehlo: Yeah it works, but obviously no console.logs should be there :)

baudehlo: One more important feature I need to add to this: inbound and outbound need to be able to specify separate cypher suites. You might want inbound to be much more strict than outbound.

celesteking:
https://github.com/haraka/Haraka/blame/master/smtp_client.js#L127
```
[CRIT] [-] [core] TypeError: client.socket.address is not a function
[CRIT] [-] [core]     at pluggableStream.<anonymous> (.. Haraka/smtp_client.js:145:57)
```
Let's get this sorted.

PR #1341, opened 2016-02-17
Copy only minimal config on install

### LINKED ISSUE #1339: Stop copying config on haraka -i
Now that config correctly loads from the core dir and merges in any changes in your local install, we can stop copying all the config files when installing Haraka.

I propose we do the following:

* on install, create only config/me, config/loglevel and config/smtp.ini
* don't copy any other config files
* possibly: provide a `haraka -c /path --copy-defaults <config-file-name>`

This will ensure that upgrading Haraka is significantly simpler and more robust.

baudehlo:
This implements and closes #1339

### DIFF
--- a/bin/haraka
+++ b/bin/haraka
@@ -191,15 +191,6 @@ function copyFile (srcFile, dstFile) {
 
 function setupHostname(confPath) {
     var hostname = os.hostname() + "\n";
-    
-    try {
-        var fd = fs.openSync(path.join(confPath, 'me'), 'w');
-        fs.writeSync(fd, hostname, null);
-    }
-    catch (e) {
-        warning("Unable to write config/me file: " + e);
-    }
-    
     try {
         var fd = fs.openSync(path.join(confPath, 'host_list'), 'w');
         fs.writeSync(fd, hostname, null);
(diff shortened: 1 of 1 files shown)


msimerson:
+1

baudehlo:
We should have a few people test this to make sure it does the right thing but it seems sound to me.

msimerson:
> We should have a few people test this

I'm about to roll 2.8.0-alpha1 with this, and will be testing this and tls PR.

msimerson:
config/plugins might also be a good one to have installed by default.

baudehlo:
Oh crap yeah. That needs to go in there.

smfreegard:
config/plugins should just be a placeholder that points to the docs IMO, the default we had made a whole load of assumptions IIRC.

We should probably create some example use cases and configurations in the docs and people can grab the appropriate config/plugins files from there.

msimerson:
config/plugins is just a file with lots of entries with most of them commented out, making it easier for folks to:

a. find the plugins that exist
b. enable them
c. have the listed in config/plugins in a reasonable location

Easy solution: put config/plugins back in. If we want to develop some documentation and perhaps a collection of config/plugins entries for particular uses, great, but I'd *still* have a default config/plugins in there, and comments pointing to those new docs.

Also, config/dkim should be installed too, as well as the enclosed dkim_key_gen.sh.

(Yeah, I'm testing.)

msimerson:
I've got a test instance of 2.8.0-alpha.2 deployed with this change. Other than having to copy in the plugins file and the dkim/dkim_key_gen.sh script, I don't see any significant issues.

PR #1342, opened 2016-02-18
handle arrays in .ini files

kgoess:
msimerson commented in a pull request https://github.com/haraka/Haraka/pull/1333#issuecomment-184313920

    One thing that might be worth changing is the way the host pool is
    configured. A method that already has widespread use is to specify an array
    for the host value instead of a string. This would necessitate extending our
    ini parser to recognize the array syntax.

The syntax looks like this. There is no "official" .ini syntax for arrays, but
this seems to be common.

    hosts[] = first_host
    hosts[] = second_host
    hosts[] = third_host

### DIFF
--- a/configfile.js
+++ b/configfile.js
@@ -8,7 +8,7 @@ var yaml = require('js-yaml');
 // for "ini" type files
 var regex = exports.regex = {
     section:        /^\s*\[\s*([^\]]*?)\s*\]\s*$/,
-    param:          /^\s*([\w@:\._\-\/]+)\s*(?:=\s*(.*?)\s*)?$/,
+    param:          /^\s*([\w@:\._\-\/\[\]]+)\s*(?:=\s*(.*?)\s*)?$/,
     comment:        /^\s*[;#].*$/,
     line:           /^\s*(.*?)\s*$/,
     blank:          /^\s*$/,
@@ -16,6 +16,7 @@ var regex = exports.regex = {
     is_integer:     /^-?\d+$/,
     is_float:       /^-?\d+\.\d+$/,
     is_truth:       /^(?:true|yes|ok|enabled|on|1)$/i,
+    is_array:       /(.+)\[\]$/,
 };
 
 var cfreader = exports;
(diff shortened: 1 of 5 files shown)


msimerson:
@kgoess , I have [commit here](https://github.com/haraka/Haraka/commit/86e69a316cd387619108bc9c39c08994d9bbc1c9) that fixes the test failure.

@baudehlo , @smfreegard , @Dexus , opinions on this implementation?

What I'm wondering about is syntax like this PR has:

```ini
[section_name]
hosts[] = first_host
hosts[] = second_host
hosts[] = third_host
```

### -vs- 

```ini
[section_name]
hosts = [ first_host, second_host, third_host ]
```

### -vs-

```ini
[section_name]
hosts = first_host
hosts = second_host
hosts = third_host
```

All would produce exactly, the same data structure:

```js
{section_name: { hosts: ['first_host', 'second_host', 'third_host'] }}
````

The question then is, which is the best syntax to using in the config files?

A couple immediate advantages to the first example is that the bracket syntax seems very clear that an array is expected. Also, values with spaces or commas in them won't require any special handling. Robustness and long-term maintainability seems to favor the first syntax.

On the minus side, it's kinda ugly syntax. I'm also wondering about discoverability. Having *seen* it's usage, it's quite obvious and explicit (good things), but if I were editing a config file and changing a setting from 'host=a.b.c' to point at multiple hosts, I'd have expected to use a syntax more like the 2nd or 3rd example. Users will need to see an example, or read the docs to get that syntax right.

The last example is an extension of the first. Basically, any time you find a key with more than one value, transform it into an array. While that's more elegant ini syntax, it could also have the side effect of creating arrays accidentally.

I'm okay merging this as-is (after my test fix is merged) but I'd like some additional feedback.

baudehlo:
I'm happy with the first syntax, but it needs documented in docs/Config.md

msimerson:
> but it needs documented in docs/Config.md

Unless my eyes are deceiving me, that's already been done.

baudehlo:
Looks like it. Sorry.

msimerson:
Merged. Thanks @kgoess !

PR #1355, opened 2016-02-22
update to uribl due to #1301

msimerson:
* the top_level_tlds object was moved to haraka-tld in PR #1301. These calls were missed.


msimerson:
Since the uribl plugin is currently broken in HEAD (and 2.8.0-alpha.2), I intend to wait until tomorrow after everyone has had a chance to review, and then merge this and publish alpha.3

PR #1357, opened 2016-02-23
Allow for override for inbound/outbound

baudehlo:
- Let's tls.ini have separate [inbound] and [outbound] sections for 'ciphers', 'requestCert', 'rejectUnauthorized' options

### DIFF
--- a/docs/Outbound.md
+++ b/docs/Outbound.md
@@ -40,9 +40,18 @@ Default: false. Switch to true to enable TLS for outbound mail when the
 remote end is capable.
 
 This uses the same `tls_key.pem` and `tls_cert.pem` files that the `tls`
-plugin uses. See the [tls plugin
+plugin uses, along with other values in `tls.ini`. See the [tls plugin
 docs](http://haraka.github.io/manual/plugins/tls.html) for information on generating those
 files.
+
+Within `tls.ini` you can specify global options for the values `ciphers`,
+`requestCert` and `rejectUnauthorized`, alternatively you can provide
+separate values by putting them under a key: `[outbound]`, such as:
+
+```
+[outbound]
+ciphers=!DES
+```
 
 * `ipv6_enabled`
 
--- a/docs/plugins/queue/smtp_forward.md
+++ b/docs/plugins/queue/smtp_forward.md
@@ -48,7 +48,8 @@ Configuration
 
   * enable\_tls=[true]
 
-    Enable TLS with the forward host (if supported)
+    Enable TLS with the forward host (if supported). TLS uses options
+    from the tls plugin.
 
   * auth\_type=[plain|login]
 
--- a/docs/plugins/queue/smtp_proxy.md
+++ b/docs/plugins/queue/smtp_proxy.md
@@ -49,7 +49,8 @@ Configuration
 
   * enable\_tls=[true|yes|1]
  
-    Enable TLS with the forward host (if supported)
+    Enable TLS with the forward host (if supported). TLS uses options from
+    the tls plugin.
 
   * auth\_type=[plain|login]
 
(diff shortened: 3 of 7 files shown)


docs/Outbound.md
```diff
@@ -40,10 +40,19 @@ Default: false. Switch to true to enable TLS for outbound mail when the
 remote end is capable.
 
 This uses the same `tls_key.pem` and `tls_cert.pem` files that the `tls`
-plugin uses. See the [tls plugin
+plugin uses, along with other values in `til.ini`. See the [tls plugin
```
msimerson: typo? tls.ini

baudehlo: Yup

baudehlo: Clearly too much Reddit r/til today ;)

baudehlo: Fixed.

Only question about this I really have is, is `smtp_forward` and proxy
really "outbound". Do they need split off into a `[local]` section?

msimerson:
> is `smtp_forward` and proxy really "outbound"

Hrmm. I don't usually think of them as such, since they're typically delivering mail to servers "within" my sphere of control (ie, my own server, or a directly client).

baudehlo:
Yeah agreed. Would a key of "local" be acceptable?

msimerson:
> Would a key of "local" be acceptable

Heh, whatever you choose is acceptable. ;)  In the case of smtp_forward and proxy, neither local, inbound, nor outbound are really good terms. DJB chose the term `remote` for qmail which fits, but is too broad as outbound would also fit in that bucket.  I can see a few choices here:

1. don't define anything for forward/proxy and let it use a tls default options. When a use case for those materializes, let the use case define it.
2. let each plugin have its own [smtp_forward], [smtp_client] section in tls.ini
3. have each plugin define a [tls_options] section in it's own config file

baudehlo:
I feel like "local" is a well used term for local networks. Another might be "internal". But "remote" seems entirely wrong.

msimerson:
> But "remote" seems entirely wrong.

Agreed. My premise is that none of those terms really "fit" well. By trying to make plugins fit into one or the other, it feels like we're going to end up stretching the terms even further in the future as new plugins arrive. I lean towards the "let each plugin/component have it's own overrides" section.

The way I anticipate this being used is that everyone uses the [tls_options : default] block upon install, until they have an issue. Like perhaps, they're using smtp_forward to an old Exchange server that advertises STARTTLS but can't negotiate DH params > 512.  When they discover that, they can add a [smtp_forward] block to tls.ini that adjusts the settings for just that plugin, without them having to determine what magic term [inbound | outbound | internal | remote] is the right incantation. If they're doing something else with smtp_proxy or smtp_bridge, it won't get inadvertently affected because smtp_* got lumped into the same "category" as smtp_proxy.

baudehlo:
The problem is those plugins use smtp_client.js for TLS support, so the
code doesn't know if it's smtp_forward or smtp_proxy.

msimerson:
> The problem is those plugins use smtp_client.js for TLS support, so the code doesn't know if it's smtp_forward or smtp_proxy.

In looking at smtp_forward, it loads smtp_client and the *only* call being made is smtp_client.get_client_plugin(), and the first argument there is a 'plugin,' which has a `name` property right? Isn't that plugin object part of the smtp_client's closure?  (it appears so to me)

It looks like it would be just a matter of passing that plugin name along, and maybe changing the big long list of positional args to an object and passing that instead, to clean up that functions interface.

msimerson:
Also, just because it's possible to be plugin specific doesn't necessarily mean it's the right design choice. I've only thought of how I'm *guessing* others will use it. There's surely other use cases that deserve consideration.

baudehlo:
OK, done it using plugin.name now.

msimerson:
> Document use of plugin name for tls config

Great timing, I was just about to suggest... ;)

msimerson:
This looks done. Is it ready?

baudehlo:
Yup.

Though I have only tested outbound, not smtp_(forward|proxy) as I don't use
them. Hopefully the alphas can iron anything out there.

msimerson:
Rebased & squashed.

PR #1359, opened 2016-02-23
remove address.js to address-rfc2821

msimerson:
* address-rfc2821 was packaged up separately last year
* finally replacing ./address with external module
* as you can see in this PR, quite a few tests use it, including a fixture
    * making this a dependency for haraka-test-fixtures


baudehlo:
Is all this stuff going into 2.8? I kind of hope so - it's been a long time
coming to make these big changes.

msimerson:
> Is all this stuff going into 2.8?

That is every bit my intention. That's part of why I'm doing this silly alpha releases, so I can get these external module dependencies all ironed out **before** the 2.8.0 release.

msimerson:
Also, due to the `haraka-tld` PR landing, everyone UPDATING has to do a "npm upgrade" anyway, to get the new modules installed. Therefore, it makes sense to try and get *all* the modules that need to be packaged separately done before our next release.

baudehlo:
Well that just happens as part of npm install anyway, since the
package.json requires new stuff.

PR #1363, opened 2016-02-26
Fix caching of STUN results

smfreegard:
Caching of the public IP was only implemented if the value was manually set in smtp.ini (which has limited value being as it is cached itself by the configuration loader).

This caches the STUN result (or lack of result), so that the STUN server is not endlessly queried for the same result.

I noticed the issue when I ran this up on a host which STUN failed to traverse the firewall - it caused connect.geoip to get a STUN timeout on every connection.

### DIFF
--- a/net_utils.js
+++ b/net_utils.js
@@ -199,7 +199,7 @@ exports.same_ipv4_network = function (ip, ipList) {
 
 exports.get_public_ip = function (cb) {
     var nu = this;
-    if (nu.public_ip) {
+    if (nu.public_ip !== undefined) {
         return cb(null, nu.public_ip);  // cache
     }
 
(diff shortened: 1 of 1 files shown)


msimerson:
I agree caching this is a good idea, but never retrying after an initial failure doesn't seem like the right choice. When setting the value to null, maybe also set a timer that tries again after N minutes?

smfreegard:
I thought about it - but this either fails or it works.  I can't see the point of endlessly flogging a dead horse when it doesn't and this isn't exactly a critical function.

Quite honestly - I think this is better moved into core (e.g. so it's always run by Haraka in init_master_respond) and set into server.notes.public_ip instead.

baudehlo:
Lint failed btw.

msimerson:
> Lint failed btw.

Trailing whitespace at the end of a comment.

baudehlo:
Assuming no more discussion about this - can this be merged?

PR #1368, opened 2016-02-29
plugins.js: small refactoring

msimerson:
* I imported swaths of plugins.js into haraka-test-fixtures
    * required to improve the "find the plugin path" logic
    * in the course of doing that I ..
* wrapped some long lines
* removed some unnecessary indent
* refactored _get_code out of _compile
    * optimization: don't read the file when hasPackageJson

### DIFF
--- a/plugins.js
+++ b/plugins.js
@@ -1,18 +1,24 @@
 'use strict';
 // load all defined plugins
 
-var logger      = require('./logger');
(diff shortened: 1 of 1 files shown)


msimerson:
Tested locally, rolling into 2.8.0-alpha.6

PR #1375, opened 2016-03-03
if rcpt_to returns DSN in msg, log it properly.

celesteking:
Otherwise we end up with DSN object in msg and not with a string.


celesteking:
And the code is repeated 2 times now in connection.js, worth DRYing.

```javascript
if (typeof msg === 'object' && msg.constructor.name === 'DSN') {
// blah
```

msimerson:
LGTM

PR #1383, opened 2016-03-10
Clarify usage of check.private_ip

### LINKED ISSUE #1382: Bug in rspamd plugin with private_ip?
### Haraka version

Haraka.js — Version: 2.7.2

### Expected behavior
According to the docs:

    check.private_ip

    Default: false

    If true, messages from private IPs will not be scanned by Rspamd.

https://github.com/haraka/Haraka/blob/master/plugins/rspamd.js#L104
should be 

    if (cfg.check.private_ip &&

instead of

    if (!cfg.check.private_ip &&

Or the docs should have the "not" removed.

swerter:
Here you go: fixes #1382

### DIFF
--- a/docs/plugins/rspamd.md
+++ b/docs/plugins/rspamd.md
@@ -49,7 +49,8 @@ rspamd.ini
 
     Default: false
 
-    If true, messages from private IPs will not be scanned by Rspamd.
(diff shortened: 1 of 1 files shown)


msimerson:
thanks @swerter , merged!

Issue #1387, opened 2016-03-16
STARTTLS queue/smtp_forward though maybe smtp_client issue Not Working


BradRuderman:
### Haraka version
Haraka.js — Version: 2.8.0-alpha.7
### Expected behavior

smtp_client will issue a "STARTTLS" command when talking on port 587, and therefore work (i think) with amazon SES

### Observed behavior

I noticed that the smtp_forward plugin doesn't work with SES. I get the following error:

```
[NOTICE] [A380A9AE-4A27-49D8-944A-6A54C94C9D05.1] [core] message mid="<abcd@mess.com>" size=1458 rcpts=1/0/0 delay=0.002 code=CONT msg=""
[DEBUG] [A380A9AE-4A27-49D8-944A-6A54C94C9D05.1] [core] running queue_outbound hooks
[DEBUG] [A380A9AE-4A27-49D8-944A-6A54C94C9D05.1] [core] running queue_outbound hook in queue/smtp_forward plugin
[DEBUG] [A380A9AE-4A27-49D8-944A-6A54C94C9D05.1] [queue/smtp_forward]
[INFO] [A380A9AE-4A27-49D8-944A-6A54C94C9D05.1] [queue/smtp_forward] forwarding to email-smtp.us-west-2.amazonaws.com:587
[INFO] [-] [core] [smtp_client_pool] [587:email-smtp.us-west-2.amazonaws.com:300] dispense() clients=1 available=0
[DEBUG] [-] [core] [smtp_client_pool] [587:email-smtp.us-west-2.amazonaws.com:300] createResource() - creating obj - count=1 min=0 max=10
[DEBUG] [-] [core] [smtp_client_pool] uuid=4545934E-47CC-46AD-91A3-A5B79FCC3AD4 host=email-smtp.us-west-2.amazonaws.com port=587 pool_timeout=300 created
[DEBUG] [A380A9AE-4A27-49D8-944A-6A54C94C9D05.1] [queue/smtp_forward] Got smtp_client: 4545934E-47CC-46AD-91A3-A5B79FCC3AD4
[INFO] [A380A9AE-4A27-49D8-944A-6A54C94C9D05.1] [queue/smtp_forward] Configuring authentication for SMTP server email-smtp.us-west-2.amazonaws.com:587
[ERROR] [-] [core] client.socket.remoteAddress undefined!
[...]

PR #1388, opened 2016-03-18
Fix outbound requesting cert and rejecting unauth

baudehlo:
Fixes outbound tls not requesting a the certificate or rejecting unauthorized tls.


baudehlo:
FWIW this needs to make it into 2.8. It's a big missing piece of the TLS changes I made, so please review @msimerson @smfreegard

BradRuderman:
@baudehlo Could this be related to my issue here?

https://github.com/haraka/Haraka/issues/1387

msimerson:
Merging, to be tested in forthcoming alpha.8.

PR #1394, opened 2016-03-29
Add rspamd results into connection.results.

swerter:
- Add rspamd results into connection results store


msimerson:
Results are already stored. Have a look and in transaction results instead.

swerter:
You're right. Sorry.

PR #1395, opened 2016-03-29
Disable SSLv2 (for DROWN)

chazomaticus:
This prevents clients from using SSLv2 (for the DROWN attack) *and*
SSLv3 (for the POODLE attack) when they issue STARTTLS.  See
<https://poodle.io/>, <https://drownattack.com/>, and
<http://disablessl3.com/#nodejs>.

The only way I know to test is a bit odd--you need an openssl version compiled with SSLv2 support.  This can be hard to find on modern systems; for example: the openssl that comes with Ubuntu 14.04 will say `unknown option -ssl2` even though later in the usage docs it has `-ssl2 - just use SSLv2`).  Run haraka, then:

    openssl s_client -ssl2 -starttls smtp -connect localhost:25

(Replace `localhost`/`25` with the hostname and port of haraka.)  If you get an error that includes a line like:

    140690032785224:error:1406D0B8:SSL routines:GET_SERVER_HELLO:no cipher list:s2_clnt.c:452:

then it successfully initiated the SSLv2 handshake and just got stymied by a lack of cipher suites, meaning *your install is vulnerable to DROWN attacks*.

If, however, you get an error like:

    140695119636296:error:1407F0E5:SSL routines:SSL2_WRITE:ssl handshake failure:s2_pkt.c:429:

then it could *not* initiate the SSLv2 handshake, and your install is *not* vulnerable to DROWN.

We've been running this a while now without issue, and 3rd parties have tested and agree that our harakas aren't susceptible to DROWN.

### DIFF
--- a/tls_socket.js
+++ b/tls_socket.js
@@ -181,12 +181,25 @@ function createServer(cb) {
             // options.secureOptions = constants.SSL_OP_ALL;
 
             // Setting secureProtocol to 'SSLv23_method' and secureOptions to
-            // constants.SSL_OP_NO_SSLv3 are used to disable SSLv2 and SSLv3
-            // protcol support.
+            // constants.SSL_OP_NO_SSLv2/3 are used to disable SSLv2 and SSLv3
+            // protocol support, to prevent DROWN and POODLE attacks at least.
+            // Node's docs here are super unhelpful, e.g.
+            // <https://nodejs.org/api/tls.html#tls_tls_createserver_options_secureconnectionlistener>
+            // doesn't even mention secureOptions.  Some digging reveals the
+            // relevant openssl docs: secureProtocol is documented in
+            // <https://www.openssl.org/docs/manmaster/ssl/ssl.html#DEALING-WITH-PROTOCOL-METHODS>,
+            // (note: you'll want to select the correct openssl version that
+            // node was compiled against, instead of master), and secureOptions
+            // are documented in
+            // <https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_set_options.html>
+            // (again, select the appropriate openssl version, not master).
+            // One caveat: it doesn't seem like all options are actually
+            // compiled into node.  To see which ones are, fire up node and
+            // examine the object returned by require('constants').
 
             options.secureProtocol = options.secureProtocol || 'SSLv23_method';
-            options.secureOptions  = options.secureOptions  ||
-                constants.SSL_OP_NO_SSLv3;
+            options.secureOptions = options.secureOptions |
+                    constants.SSL_OP_NO_SSLv2 | constants.SSL_OP_NO_SSLv3;
 
             var requestCert = true;
             var rejectUnauthorized = false;
@@ -283,8 +296,11 @@ function connect (port, host, cb) {
         // TODO: bug in Node means we can't do this until it's fixed
         // options.secureOptions = constants.SSL_OP_ALL;
 
+        // See comments around similar code in createServer above for what's
+        // going on here.
         options.secureProtocol = options.secureProtocol || 'SSLv23_method';
-        options.secureOptions  = options.secureOptions  || constants.SSL_OP_NO_SSLv3;
+        options.secureOptions = options.secureOptions |
+                    constants.SSL_OP_NO_SSLv2 | constants.SSL_OP_NO_SSLv3;
 
         var sslcontext = (tls.createSecureContext || crypto.createCredentials)(options);
 


tls_socket.js
```diff
@@ -181,12 +181,25 @@ function createServer(cb) {
...
+            // (again, select the appropriate openssl version, not master).
+            // One caveat: it doesn't seem like all options are actually
+            // compiled into node.  To see which ones are, fire up node and
+            // examine the object returned by require('constants').
 
             options.secureProtocol = options.secureProtocol || 'SSLv23_method';
-            options.secureOptions  = options.secureOptions  ||
-                constants.SSL_OP_NO_SSLv3;
+            options.secureOptions = options.secureOptions |
+                    constants.SSL_OP_NO_SSLv2 | constants.SSL_OP_NO_SSLv3;
```
chazomaticus: Note that I'm changing the first `|` to `||`.  This means that regardless of what the caller passes in `secureOptions`, haraka will prevent them from shooting themselves in the foot by allowing SSLv2/v3 connections.

I'm not 100% sure this is always desired, but it seems like good security to me.

msimerson:
I'd prefer to [deal with DROWN](https://nodejs.org/en/blog/) by bumping our minimum version of node.js to 0.12.12.

msimerson:
Even better, 0.10.43 was patched to remove SSLv2 as well, so simply requiring at least 0.10.43 would also take care of it.

chazomaticus:
That's a great idea too.  Might not hurt to do both, at least for now, just in case someone's stuck on an old node version.

msimerson:
> just in case someone's stuck on an old node version.

It might be painting with too broad of a brush but in my experience, people who don't upgrade their node are not going to upgrade their Haraka, and vice versa.  I think we're better off just removing all of the vestiges of SSLv2, which makes this patch much smaller.

smfreegard:
> It might be painting with too broad of a brush but in my experience, people who don't upgrade their node are not going to upgrade their Haraka, and vice versa

+1

chazomaticus:
I can't speak for anyone else, but I will say that in our deployment, it's far easier to upgrade just Haraka than node itself.

msimerson:
> it's far easier to upgrade just Haraka

That may be, but that only solves one security issue. Everyone needs to upgrade node for a number of other security (mostly openssl) issues as well. Are you using some "slowly if at all" platform like Debian where you have to roll your own node releases to get them them in a timely manner? (if so, sorry, I've felt that pain, and that's a cost of using that distro).

It won't be very far into the future when everyone (node.js upstream, us) drops support for 0.10 and then 0.12. Upgrading node might be more work but security-wise, it's really not optional. Patching it locally for your deployment until you upgrade node? Sure.

I'm having a hard time seeing any upside to introducing this code into Haraka 2.8 and I'm seeing a risk of constants.SSL_OP_NO_SSLv2 no longer existing in the future, causing a new maintenance issue.

msimerson:
I looked into this some more. SSLv2 has been disabled-by-default in node since 0.10.29, and was removed entirely in 0.10.43. @chazomaticus , did you test with a version of node > 0.10.29 and discover that Haraka was vulnerable?

Bumping to 0.10.43 or 0.12.12 would not resolve *drown* for node 4 or 5, if that truly is an issue. Based on my reading of [nodejs/node#5433], I don't think that downgrade is possible in node 4 or 5.

chazomaticus:
> did you test with a version of node > 0.10.29 and discover that Haraka was vulnerable?

Yes, our haraka (2.7.2-ish plus a few local mods) on node v0.12.9 was vulnerable.  Not sure what flags that node was compiled with.

I'd just like to reiterate my support for this patch.  In my opinion, it's not great security if it can be defeated with a compile flag on the language runtime, which many people who would deploy haraka might not have control over or care about.  This patch would keep everyone secure from this threat regardless of the whims of administrators of the machines it's running on.

> I'm seeing a risk of constants.SSL_OP_NO_SSLv2 no longer existing in the future, causing a new maintenance issue.

I'd also like to point out that I don't see a maintenance issue here.  I believe js coerces an undefined value, e.g. if SSL_OP_NO_SSLv2 were no longer defined by node, into a 0 in this expression, without warning or error.

In other words, I think this code would ensure haraka is doing the right thing regardless of node version or compile flags, and that sounds like a good idea to me.

msimerson:
> Yes, our haraka ... on node v0.12.9 was vulnerable.

I hereby withdraw my objections. I further recommend that if the @haraka/core to wander by here also agrees, they click that green "Merge pull request" button.

baudehlo:
Done.

chazomaticus:
Thanks @msimerson, @baudehlo, and @smfreegard!

msimerson:
Thank *you* @chazomaticus!

PR #1399, opened 2016-03-31
require node 0.10.43+

msimerson:


baudehlo:
+1

I'm going to try a binary node build soon so I can see if it's straightforward to get my older Debian boxes up to date with node 4 or 5. Hopefully that can end the shame of me being the stick in the mud on node 0.10.

PR #1404, opened 2016-04-01
standardize on oxford spelling

msimerson:


baudehlo:
*twitch* But fair enough.

PR #1418, opened 2016-04-19
Fix broken encoding on _char_to_qp

### LINKED ISSUE #1417: Encoding error with add_body_filter
### Haraka version
Git pull from 11 april 2016

### Expected behavior
Mail arriving with correct encoding like
test rréé èè

### Observed behavior
Instead its arriving with 
test rrÃ©Ã© Ã¨Ã¨

### Steps to reproduce
I sent a mail with outlook and a smtp tool (SendSMTP) with a body like "test rréé èè" and "tést" as subject.
The subject isn't broken when it arrives on my gmail box, the same code worked in the previous haraka version i had (from december / january). I tried this simple 

This is the code i just tried

```
var plugin = exports
//var fs = require('fs');


        plugin.register_hook('data', 'add_hfcontent');
}

exports.add_hfcontent = function (next, connection) {
        connection.transaction.notes.banner = false;
        connection.logcrit("inside the filter beginning");
        connection.transaction.add_body_filter('text/html', function (ct, enc, buf) {
            return(buf);
        });
        return next();
}
```

DarkSorrow:
Fixes #1417 .

- Fix for the issue #1417, the encoding was broken on some frenchy characteres when add_filter was used, which seems to call this function.

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -133,7 +133,7 @@ exports.decode_qp = function (line) {
 };
 
 function _char_to_qp (ch) {
-    var b = new Buffer(ch);
+    var b = new Buffer(ch, "binary");
     var r = '';
     for (var i=0; i<b.length; i++) {
         r = r + '=' + _pad(b[i].toString(16).toUpperCase(), 2);


DarkSorrow:
When i test to send a "test rr éé bøø" in a mail i get a correct encoding on the message i receive back
Here is the mail i received with my test above

> From: "xxxxxx" <xxxx@xxxxxxx>
> Subject: test
> To: xxxxxxxxxxx
> MIME-Version: 1.0
> Date: Tue, 19 Apr 2016 11:26:52 +0200
> X-Mailer: SendSMTP Version 2.16 Release 1 Build 1
> Content-Type: text/plain; charset="UTF-8"
> Content-Transfer-Encoding: quoted-printable
> Content-Disposition: inline
> 
> test rr =C3=A9=C3=A9 b=C3=B8=C3=B8

PR #1419, opened 2016-04-19
Use utf8 encoding now for QP not binary

baudehlo:
Replaces #1418 .


baudehlo:
Tests are failing because spameatingmonkey appears to be down.

baudehlo:
Better fix for #1417

DarkSorrow:
Yes this work on the mail i've sent with the encoding, i will try to make some more test later too.
I closed my other request.

baudehlo:
Merged

PR #1422, opened 2016-04-26
Remove stray log error from list detection code

davebeyer:
Fixes 1.

- Removal of incorrect log error from list detection code 

Checklist:
- [N/A] docs updated
- [N/A] tests updated

(Unless I'm overlooking something! ...) This removes an incorrect 
connection.logerror() from the (currently rudimentary) email
list detection code in the data.headers plugin.  As currently coded,
it would record an error to the log file for any email with a "Sender"
header that didn't start with "owner-".

### DIFF
--- a/plugins/data.headers.js
+++ b/plugins/data.headers.js
@@ -367,7 +367,9 @@ exports.mailing_list = function (next, connection) {
                     found_mlm++;
                     continue;
                 }
-                connection.logerror(plugin, "mlm start miss: " + name + ': ' + header);
+                // NOTE: Unlike the next "j.match" code block, this condition alone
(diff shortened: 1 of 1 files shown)


davebeyer:
Looks like it failed due to a trailing space... I'll remove and rebase later.

msimerson:
The reason I put that error message there was that I wasn't certain if any other mailing lists added a sender header. Instead of deleting it, changing it debug priority would be better.

davebeyer:
@msimerson, thanks for the comment.  I've changed it to a logdebug(), removed the trailing space, and rebased to a single commit.

Issue #1098, opened 2015-07-24
Normalize the local and remote connection data


msimerson:
Instead of:

```js
connection = {
    local_ip: '1.2.3.4',
    local_port: 25,
    local_host: 'mail.example.com',
    haproxy_ip: undefined,
    remote_ip: '5.4.3.2',
    remote_port: 23432,
    remote_host: 'mail.sender.com',
    hello_host: 'mail.sender.com',
    greeting: 'EHLO',
}
```

Nicely organize the data into this format:

```js
connection = {
    local = {
        ip:   '1.2.3.4',
        proxy: undefined,
        port: 25,
        host: 'mail.example.com',
    },
    remote = {
        ip:   '5.4.3.2',
        port: 23432,
        host: 'mail.sender.com',
    },
    hello: {
        host: 'mail.sender.com',
        verb: 'EHLO',
    },
    tls: {
        using: true,
        cipher: 'AES256-GCM-SHA384',
        authorized: ...
    }
    ....
}
```

smfreegard:
TLS attributes could also live under connection.remote.tls as currently this is stored in connection.notes.tls by the TLS plugin:

`````
    /* Upgrade the connection to TLS. */
    connection.client.upgrade(plugin.tls_opts, function (authorized,
            verifyError, cert, cipher) {
        clearTimeout(timer);
        connection.reset_transaction(function () {
            connection.hello_host = undefined;
            connection.using_tls = true;
            connection.notes.tls = {
                authorized: authorized,
                authorizationError: verifyError,
                peerCertificate: cert,
                cipher: cipher
            };
            connection.loginfo(plugin, 'secured:' +
                ((cipher) ? ' cipher=' + cipher.name + ' version=' + cipher.version : '') +
                ' verified=' + authorized +
                ((verifyError) ? ' error="' + verifyError + '"' : '' ) +
                ((cert && cert.subject) ? ' cn="' + cert.subject.CN + '"' +
                ' organization="' + cert.subject.O + '"' : '') +
                ((cert && cert.issuer) ? ' issuer="' + cert.issuer.O + '"' : '') +
                ((cert && cert.valid_to) ? ' expires="' + cert.valid_to + '"' : '') +
                ((cert && cert.fingerprint) ? ' fingerprint=' + cert.fingerprint : ''));
            return next(OK);  // Return OK as we responded to the client
        });
    });
`````

This can be done as part of #1099

baudehlo:
Maybe we should have an IP address class?

[...]

PR #1423, opened 2016-04-26
Add IP whitelisting to early_talker

Synchro:
The early_talker plugin currently causes delays even when submitting from localhost, which is quite unnecessary. The additional delay at DATA makes things even worse.

This change adds an ip_whitelist to the early talker plugin, based on the pattern used by the greylisting and xclient plugins. By default it includes 127.0.0.1 and ::1.

Includes tests and docs.

### DIFF
--- a/plugins/early_talker.js
+++ b/plugins/early_talker.js
@@ -48,7 +48,7 @@ exports.early_talker = function(next, connection) {
     }
 
     // Don't delay whitelisted IPs
-    if (plugin.whitelist && plugin.ip_in_list(plugin.whitelist, connection.remote_ip)) { // check connecting IP
+    if (plugin.ip_in_list(connection.remote_ip)) { // check connecting IP
         connection.results.add(plugin, { skip: 'whitelist' });
         return next();
     }
@@ -79,21 +79,25 @@ exports.early_talker = function(next, connection) {
 
 
 /**
- * Check if the ip is whitelisted
+ * Check if an ip is whitelisted
  *
- * @param  {Array} whitelist A list of ipaddr objects
  * @param  {String} ip       The remote IP to verify
  * @return {Boolean}         True if is whitelisted
  */
-exports.ip_in_list = function (whitelist, ip) {
+exports.ip_in_list = function (ip) {
+    var plugin = this;
     var ipobj = ipaddr.parse(ip);
 
-    for (var i = 0; i < whitelist.length; i++) {
+    if (!plugin.whitelist) {
+        return false;
+    }
+
+    for (var i = 0; i < plugin.whitelist.length; i++) {
         try {
-            if (ipobj.match(whitelist[i])) {
+            if (ipobj.match(plugin.whitelist[i])) {
                 return true;
             }
-        } catch (e) {
+        } catch (ignore) {
         }
     }
     return false;
@@ -120,8 +124,8 @@ exports.load_ip_list = function(list) {
             }
 
             whitelist.push(addr);
-        } catch (e) {
+        } catch (ignore) {
         }
     }
     return whitelist;
-}
+};
--- a/plugins/early_talker.js
+++ b/plugins/early_talker.js
@@ -1,5 +1,8 @@
 // This plugin checks for clients that talk before we sent a response
 
+var ipaddr = require('ipaddr.js');
+var isIPv6 = require('net').isIPv6;
+
 exports.register = function() {
     var plugin = this;
     plugin.load_config();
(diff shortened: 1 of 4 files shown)


Synchro:
Tests are passing now!

plugins/early_talker.js
```diff
@@ -41,6 +47,12 @@ exports.early_talker = function(next, connection) {
         return next();
     }
 
+    // Don't delay whitelisted IPs
+    if (plugin.whitelist && plugin.ip_in_list(plugin.whitelist, connection.remote_ip)) { // check connecting IP
```
msimerson: There's no need to pass plugin.whitelist here as an argument. Just pass in the remote_ip. In exports.ip_in_list, you can add a `var plugin = this` at the beginning and you'll have access to plugin.whitelist within the function.

msimerson: You can also move the `if (plugin.whitelist)` test into plugin.ip_in_list, simplifying the request dispatch.

plugins/early_talker.js
```diff
@@ -64,3 +76,52 @@ exports.early_talker = function(next, connection) {
...
+ */
+exports.ip_in_list = function (whitelist, ip) {
+    var ipobj = ipaddr.parse(ip);
+
+    for (var i = 0; i < whitelist.length; i++) {
+        try {
+            if (ipobj.match(whitelist[i])) {
+                return true;
+            }
+        } catch (e) {
```
msimerson: in cases where you're ignoring the caught error, rename the error from 'e' to 'ignore', making your intentions explicit.

plugins/early_talker.js
```diff
@@ -64,3 +76,52 @@ exports.early_talker = function(next, connection) {
...
+            var addr = list[i];
+            if (addr.match(/\/\d+$/)) {
+                addr = ipaddr.parseCIDR(addr);
+            }
+            else {
+                addr = ipaddr.parseCIDR(addr + ((isIPv6(addr)) ? '/128' : '/32'));
+            }
+
+            whitelist.push(addr);
+        } catch (e) {
```
msimerson: same as above, change to `catch (ignore)`

msimerson:
Hey @Synchro , this looks very good. Thanks for the PR.  I've made a couple minor suggestions that would improve it.

Also, I added a comment to #1098 suggesting that we add a `connection.remote.is_private` boolean, which would match for all localhost and rfc1918 IP space. There's a number of plugins (including this one) that would be simplified (and speed overall performance) if is_private was calculated just once per connection. I'm noting that here so that when we implement #1098, this plugin can take advantage of that.

Synchro:
Thanks for the tips, I've just pushed those, but getting this failure. Not sure why - it means that `connection.remote_ip` is not an IP - why would that be? Did I miss something?

plugins/early_talker.js
```diff
@@ -64,3 +76,56 @@ exports.early_talker = function(next, connection) {
...
+
+/**
+ * Check if an ip is whitelisted
+ *
+ * @param  {String} ip       The remote IP to verify
+ * @return {Boolean}         True if is whitelisted
+ */
+exports.ip_in_list = function (ip) {
+    var plugin = this;
+    var ipobj = ipaddr.parse(ip);
```
msimerson: Why parse the IP if plugin.whitelist doesn't exist?   (shouldn't that check come before this one?)

baudehlo: This is a guess, but the connection in this case is from
haraka-test-fixtures, and doesn't implement remote_ip

https://github.com/haraka/haraka-test-fixtures/blob/master/lib/connection.js

msimerson: > This is a guess, but the connection in this case is from haraka-test-fixtures....and doesn't implement remote_ip

That's unlikely, as he's setting remote_ip in the test.

msimerson:
Ahah, it's not **your** test that's failing, its the previous two tests that are failing now:

* `early_talker - is an early talker`
* `early_talker - is an early talker, reject=false`

Looking now to see why...

Synchro:
The other tests don't set the remote_ip, so they fail the whitelist checks. Shoudl I fix the other tests, or check for remote_ip in the code?

baudehlo:
Ah never mind - ignore my recent PR then - I was looking at the wrong
branch. :)

msimerson:
The problem is that you're parsing connection.remote_ip before the plugin.whitelist check.

Synchro:
That's better! Are you OK with this now, or should I squash commits?

msimerson:
merged. thanks @Synchro

PR #1425, opened 2016-04-28
Bump iconv module for use in node v6.x

EyePulp:
Haraka can't do an npm install with the older version of iconv - update to the newer one.


baudehlo:
Merged. Thanks.

PR #1428, opened 2016-04-30
Fix dot stuffing for a single dot

### LINKED ISSUE #1411: Dot stuffing not applied to single `.` lines
### Haraka version
2.7.2 (though I see the relevant code is the same in HEAD)

### Expected behavior
A message that contains a `.` by itself on a line is dot stuffed during delivery.

### Observed behavior
The line is not dot-stuffed and so the receiving end misinterprets the solo `.` as the end-of-DATA marker, and then complains about non-SMTP command content after it (i.e. the remainder of the message). In the receiver's mailbox, the message is truncated. Longer lines that start with a `.` are dot stuffed correctly.

### Steps to reproduce
Send a message body containing a `.` on a line by itself, for example:
```
hello
.
hello
```

This code in outbound.js:

    if (dot_stuffed === false && line.length > 3 && line.substr(0,1) === '.') {

A `.` by itself on a line is 3 chars including the `CRLF`, and so does not match this test and thus will not be dot-stuffed.

Changing the `> 3` to `>= 3` would match it correctly, and I would have put that in a PR, however, I can't see where in the code it sends the *real* end of message marker and so didn't want to risk breaking that if it relies on it being part of the message before trying to send it.

baudehlo:
Fixes #1411 .

- This fixes dot-stuffing when sending email via outbound using new contents

I've reviewed any problems that are raised in the original ticket and can't find any.

I thought about writing tests, but it's a lot of work - requires too much monkey patching for my current laziness.

### DIFF
--- a/outbound.js
+++ b/outbound.js
@@ -417,7 +417,7 @@ exports.send_email = function () {
     while (match = re.exec(contents)) {
         var line = match[1];
         line = line.replace(/\r?\n?$/, '\r\n'); // make sure it ends in \r\n
-        if (dot_stuffed === false && line.length > 3 && line.substr(0,1) === '.') {
+        if (dot_stuffed === false && line.length >= 3 && line.substr(0,1) === '.') {
             line = "." + line;
         }
         transaction.add_data(new Buffer(line));


Synchro:
Thanks!

PR #1437, opened 2016-05-02
Change net_utils.get_ips_by_host error handling.

### LINKED ISSUE #1381: net_utils.get_ips_by_host error codes
```js
// in net_utils
                dns.resolve4(hostname, function resolve_cb (err, res) {
                    if (err) {
                        errors.push(err.message);
                        return iter_done();
                    }

// in fcrdns plugin
                        switch (err[e]) {
                            case 'queryAaaa ENODATA':
                            case 'queryAaaa ENOTFOUND':
                                break;
```
Recent node sets various [DNS error props](https://nodejs.org/api/dns.html#dns_error_codes). We should push entire error above and not just the message, as recent node returns it as "queryAaaa ENOTFOUND domain.name" and who knows how it's gonna change.

Dexus:
Fixes #1381

- change ERROR strings to dns error constants

### DIFF
--- a/dkim.js
+++ b/dkim.js
@@ -306,8 +306,8 @@ DKIMObject.prototype.end = function () {
         clearTimeout(timer);
         if (err) {
             switch (err.code) {
-                case 'ENOTFOUND':
-                case 'ENODATA':
+                case dns.NOTFOUND:
+                case dns.NODATA:
                 case dns.NXDOMAIN:
                     return self.result('no key for signature', 'invalid');
                 default:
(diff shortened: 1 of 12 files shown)


baudehlo:
What about `plugins/lookup_rdns.strict.js` and `plugins/mail_from.is_resolvable.js` which both use this function?

Dexus:
Fixed now, had pushed the wrong directory.

baudehlo:
Merged.

PR #1440, opened 2016-05-05
auth_ldap fixes

gramakri:
Fixes # .

- Advertise AUTH without TLS for private ips in auth_ldap . This is similar to auth_flat.
- Handle LDAP client connection failure. When the LDAP server is down, haraka crashes because of missing error handler.

Checklist:
- [X ] docs updated
- [X ] tests updated


baudehlo:
I don't see why private_ip ranges should be exempt - the use of TLS is there to prevent password leakage over the network. SMTP AUTH isn't secure.

gramakri:
@baudehlo Do you think we can add an option for this? I can add a `require_tls` option which defaults to true.

That said, what I actually need is a require_tls_for_private_ips. So, I might drop this patch. The first patch for ldap client connection error is still useful.

gramakri:
I will close this PR and create a new one with just the client crash fix for now.

PR #1441, opened 2016-05-05
set max_unrecognized_commands to 10

gramakri:
Fixes # .

- max_unrecognized_commands is documented to have a default value of 10

Without this fix, all haraka AUTH fails. From the logs, this appears because max_recognized_commands hits the limit 2 and the auth plugins use the unrecognized command hook.

Checklist:
- [X ] docs updated
- [X ] tests updated


smfreegard:
Rather than changing the default to 10, why don't you simply put this plugin *after* the auth/* plugins in your `config/plugins` file (or preferably at the bottom of it), that way it only counts unrecognized commands not caught by other plugins.

gramakri:
@smfreegard it is documented to have the default as 10. https://haraka.github.io/manual/plugins/max_unrecognized_commands.html

msimerson:
merging, since this PR matches the docs. Thanks @gramakri

PR #1443, opened 2016-05-05
auth_ldap: handle ldap client connection error

gramakri:
- Fix crash when ldap client cannot connect to ldap server


gramakri:
@baudehlo I separated this commit from the other ldap commit

PR #1447, opened 2016-05-07
pass connection object to get_plain_passwd

pstadler:
Make the `connection` object accessible when `get_plain_passwd()` is being called. To avoid backwards incompatibilities, this has been added as an additional 3rd parameter. The reasoning behind this change is that one might wants to push data to the `notes` object, in order to persist information about the user.

Checklist:
- [ ] docs updated
- [x] tests updated

### DIFF
--- a/plugins/auth/auth_base.js
+++ b/plugins/auth/auth_base.js
@@ -22,7 +22,7 @@ exports.hook_capabilities = function (next, connection) {
 };
 
 // Override this at a minimum. Run cb(passwd) to provide a password.
-exports.get_plain_passwd = function (user, cb) {
+exports.get_plain_passwd = function (user, cb, connection) {
     return cb();
 };
 
@@ -52,7 +52,7 @@ exports.check_plain_passwd = function (connection, user, passwd, cb) {
         if (plain_pw === null  ) { return cb(false); }
         if (plain_pw !== passwd) { return cb(false); }
         return cb(true);
-    });
+    }, connection);
 };
 
 exports.check_cram_md5_passwd = function (connection, user, passwd, cb) {
@@ -68,7 +68,7 @@ exports.check_cram_md5_passwd = function (connection, user, passwd, cb) {
             return cb(true);
         }
         return cb(false);
-    });
+    }, connection);
 };
 
 exports.check_user = function (next, connection, credentials, method) {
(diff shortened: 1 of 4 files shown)


pstadler:
I just wanted to mention that I'm running Haraka with Node v6 w/o any problems.

pstadler:
Example usage:

```js
exports.get_plain_passwd = function getPlainPasswd(user, cb, connection) {
  const plugin = this;

  api.getAuthUser(user)
    .then((response) => response.body)
    .then((result) => {
      if (!result.valid) {
        plugin.loginfo(`No password found for unknown user "${user}"`);

        return cb();
      }

      connection.notes.auth_email_address = result.email;

      return cb(result.password);
    })
    .catch((err) => {
      plugin.logerror(`Error fetching password for user "${user}": ${err}`);

      return cb();
    });
};
```

msimerson:
Passing in the connection object is a perfectly reasonable thing to do. Changing the calling signature by passing it as a positional argument violates the nodejs convention of passing the callback as the last positional argument.

I'm torn on this. I'd like to suggest switching to calling the functions with a single object, so that passing in more stuff in the future doesn't change the call signature and won't pose backwards compatibility problems.  But doing that now would be a potentially breaking change (i.e., has to wait for 3.0) for anyone with 3rd party auth plugins.

Makes me want to suggest we create a 2.8 branch and start making all our 3.0 changes now in master, with back ports to 2.8 as necessary.

pstadler:
I have the same feeling about introducing another argument, but then again, breaking backwards-compatibility for this very function is a bad thing.

Let me know if I can do anything here.

pstadler:
Is there any chance that this lands somehow in the 2.8 releases?

baudehlo:
Can you just check arguments.length and do the right thing if it's 3 vs 2?

pstadler:
If I understand you correctly, this won't solve the problem of breaking custom auth plugins written with the current signature of `get_plain_passwd()`.

baudehlo:
Sorry I meant function.length, not arguments.length:

```
> var f = function (a, b) { }
undefined
> f.length
2
```

So you can check how many params get_plain_password takes, and pass in the connection object as second param if it expects 3, or just user, cb if it expects 2.

baudehlo:
Please see https://github.com/haraka/Haraka/tree/pstadler-get_plain_passwd-pass-connection for an implementation that works (tests pass, with both two arg and 3 arg forms).

pstadler:
This seems rather hacky to me, but would still solve my problems of course. 👍

Issue #1448, opened 2016-05-08
Remove Haraka word from Receive headers


ricardopolo:
Currently in the email headers is added the Haraka word and its version.

You see something like
`Received: from na01-by2-obe.outbound.protection.outlook.com
 (mail-by2lp0236.outbound.protection.outlook.com [207.46.163.236]) by
 ServerName (Haraka/2.7.3) with ESMTPS id`

So, any way we can disable the showing of the Haraka version? 
We prefer to not show exactly which server or version are using for security requirements.

Thanks!

darkpixel:
As a quick fix, you could tweak connection.js here: https://github.com/haraka/Haraka/blob/master/connection.js#L1356

prasayer:
How do I go further and remove the client ip and rDNS name?
I can edit that connection.js but am reluctant to do so. Any configuration file is available for this?

baudehlo:
Nope but you could write a plugin that monkey-patches it rather than editing the file.

prasayer:
Thanks Matt. Great work with Haraka. Can you give me some more pointers on
this approach or point me to a tutorial?

msimerson:
The other idea we've suggested, which would be the ideal solution here, is to create a config option (maybe header.received in smtp.ini) that permitted one to define the exact received string, along with a few tokens (maybe %ip, etc..) that get replaced when the header is added.

prasayer:
Agreed. If it is in the cards for near future I can wait for it to be
released. Would you know?

msimerson:
It's waiting for someone to care enough to write it.

PR #1449, opened 2016-05-08
Add ability to hide (Haraka/x.x.x) in headers.  Closes GH-1448.

darkpixel:
Fixes # 1448

- If a file exists in the config directory 'header_hide_version', do not put the Haraka version number into the received headers.

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -1339,6 +1339,8 @@ Connection.prototype.received_line = function() {
     if (this.authheader) smtp = smtp + 'A';
     // sslheader only populated with node.js >= 0.8
     var sslheader;
+    var header_hide_version = config.get('header_hide_version') ? true : false;
+
     if (this.notes.tls && this.notes.tls.cipher) {
         sslheader = '(version=' + this.notes.tls.cipher.version +
             ' cipher=' + this.notes.tls.cipher.name +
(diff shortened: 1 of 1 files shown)


msimerson:
I'm wondering if the better way to address this may be to create a header_string file, or a header_string property in smtp.ini and have template variables in it that get replaced. Then the received header can be customized much further my editing a config file, and it's less likely to be affected by future Haraka upgrades.

darkpixel:
```
Received: from some.server.tld (some.server.tls
        [1.2.3.4])       by my.mailserver.tls (Haraka/2.7.3) with ESMTP id
        4C0675D5-240D-4015-A0AC-01B2C15E2A35.1  envelope-from
        <joe@server.tls>;       Thu, 05 May 2016 12:47:54 -0700
```

I can understand wanting to remove the version string, but what else would you want to remove out of there?  The header format is pretty standard in debugging and tracking down problems.

Note: I'm not familiar enough with the RFCs to know whether there is a standard header format or not.

msimerson:
> what else would you want to remove out of there?

Off the top of my head, things I've seen other MTAs put there, and other things some don't want included:

* MTA
* MTA version
* TLS info
* TLS protocol
* AUTH info (user)

There's been plenty of conversations about this in the QP mailing list, with a diversity of opinions about what should and should not be included, for various reasons. Perhaps a reasonable way to describe it is that including more data is beneficial for debugging and less is beneficial for security. Each organization has their priorities and making the received header so customizable would avoid future bike shed discussions on the matter.

ricardopolo:
Do agree with @msimerson. Every user should be able to decide what put in here :)

darkpixel:
I agree.  Unfortunately I'm still a n00b when it comes to nodejs/javascript.  No guarantees on when it'll be ready.  ;)

baudehlo:
Any objections from the team for merging this?

ricardopolo:
In my opinion could be better if you can customize what do you want to show... but... meanwhile If it was my decision I would approved this PR :)

And maybe later work in the full customization

msimerson:
+1 to what @ricardopolo said.

baudehlo:
Merge away. I want to get 2.8.1 out asap.

PR #1452, opened 2016-05-10
Make qlist work better

baudehlo:
Makes bin/haraka --qlist output number of recipients at which domain

### DIFF
--- a/outbound.js
+++ b/outbound.js
@@ -31,7 +31,9 @@ var WRITE_EXCL  = core_consts.O_CREAT | core_consts.O_TRUNC | core_consts.O_WRON
 
 var MAX_UNIQ = 10000;
 var my_hostname = require('os').hostname().replace(/\\/, '\\057').replace(/:/, '\\072');
-var fn_re = /^(\d+)_(\d+)_/; // I like how this looks like a person
+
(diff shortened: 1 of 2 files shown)


baudehlo:
@smfreegard made it work like you wanted it. This is a bit cleaner now.

outbound.js
```diff
@@ -312,13 +320,41 @@ exports._add_file = function (hmail) {
...
+                todo_struct.rcpt_to = todo_struct.rcpt_to.map(function (a) { return new Address (a); });
+                todo_struct.mail_from = new Address (todo_struct.mail_from);
+                todo_struct.file = file;
+                todo_struct.full_path = path.join(queue_dir, file);
+                cb(null, todo_struct);
+            }
+        });
+        td_reader.on('end', function () {
+            if (Buffer.byteLength(todo) !== todo_len) {
+                console.error("Didn't find right amount of data in todo for file:", file);
```
smfreegard: Shouldn't this be `return cb(new Error("Didn't find right amount of data in todo for file: ", file)` instead?

baudehlo: If I did that the whole thing would stop (but I guess now it would hang).

baudehlo:
While SMF has improvements, I am going to merge this because I'm working on other code this depends on.

PR #1453, opened 2016-05-10
Restore old config

baudehlo:
I broke the config with the SpamAssassin plugin fix. This restores it.


baudehlo:
This alone is probably worth 2.8.1

PR #1456, opened 2016-05-12
rcpt_to_ldap fixes

gramakri:
- rcpt_to.ldap: handle connection error:  without this haraka crashes if ldap server is down
- ~~ldapjs is a required dependacy: ldapjs is in package.json~~ (removed since it is an optional dep)
- pick up domain specific config correctly: code incorrectly uses setting based on host_list file
- soft deny on ldap client and search errors: code incorrectly allows rcpt to's if the ldap server is down or generated an error.

Checklist:
- [X ] docs updated
- [X ] tests updated

### DIFF
--- a/plugins/rcpt_to.ldap.js
+++ b/plugins/rcpt_to.ldap.js
@@ -66,6 +66,11 @@ exports.ldap_rcpt = function(next, connection, params) {
         return next();
     }
 
+    client.on('error', function (err) {
+        connection.loginfo(plugin, 'client error ' + err.message);
+        next();
+    });
+
     client.bind(cfg.binddn, cfg.bindpw, function(err) {
         connection.logerror(plugin, 'error: ' + err);
     });
--- a/plugins/rcpt_to.ldap.js
+++ b/plugins/rcpt_to.ldap.js
@@ -53,7 +53,7 @@ exports.ldap_rcpt = function(next, connection, params) {
 
     txn.results.add(plugin, { msg: 'connecting' });
 
-    var cfg = plugin.in_host_list(domain) ? plugin.cfg.main : plugin.cfg[domain];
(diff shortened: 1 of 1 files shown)


msimerson:
> ldapjs is a required dependacy: ldapjs is in package.json

Only partly correct, it's in the optional dependencies section. Leave the try / catch block please.

gramakri:
@msimerson do you think we should fix auth_ldap  ? Or is the logic for that different?

msimerson:
> do you think we should fix auth_ldap

I really have no idea what's broken.  I don't use the LDAP plugin but I know it's working for at least a couple orgs (just look at commit history for the plugin). My only concern would be making sure the fix doesn't break it for their use cases.

gramakri:
@msimerson sorry, i wasn't clear. I meant the ldap require in https://github.com/haraka/Haraka/blob/master/plugins/auth/auth_ldap.js#L3

gramakri:
can someone restart the build? i suspect it is a spurious failure

msimerson:
> I meant the ldap require in...

Oh, I see. It wouldn't hurt. In the case of these LDAP plugins, it's not a huge problem if they require ldapjs, as it's not shocking for Haraka to crash if ldapjs didn't install correctly and/or can't load. That said, preventing Haraka from crashing and instead emitting a useful error message is always going to the better choice.

msimerson:
> pick up domain specific config correctly: code incorrectly uses setting based on host_list file

I wonder if you happened to notice this line of code: https://github.com/haraka/Haraka/blob/master/plugins/rcpt_to.ldap.js#L7

Are you sure it was incorrectly using that setting?

baudehlo:
Merged, thanks.

PR #1458, opened 2016-05-13
Allow SSH for dev when using Docker

erickpece:

Allow ssh to Docker container for dev testing.  

> rm -f /etc/service/sshd/down

 is required per https://github.com/phusion/baseimage-docker#enabling-ssh

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get -y -q install nodejs && \
 
 # Optional, useful for development
 # See https://github.com/phusion/baseimage-docker#login_ssh
+#RUN rm -f /etc/service/sshd/down
 #RUN /usr/sbin/enable_insecure_key
 
 # Install Haraka


Dexus:
is not necessary you can also use `docker exec -t -i container bash`

erickpece:
In that case, maybe it makes more sense to remove all of these lines:

```
 # Optional, useful for development
 # See https://github.com/phusion/baseimage-docker#login_ssh
 #RUN rm -f /etc/service/sshd/down
 #RUN /usr/sbin/enable_insecure_key
```

Dexus:
I don't know why there is a Dockerfile. But if it will be used for Production Environments it is okay how it is. For Developer they can edit the Dockerfile :smile: 

my personal mind of this dockerfile is: It has a to big overhead for a container... it where better with a other distro like "Alpine Linux". But anyway.

erickpece:
If the Dockerfile shouldn't be there, then why keep it at all?  If it should be there, why leave it with incomplete instructions?

I do agree that the overhead is pretty big though.

msimerson:
merging this, because the line is commented out, so it doesn't change anything, just provides a developer expedience.

PR #1466, opened 2016-05-18
Outbound pool support

baudehlo:
I'd probably suggest holding off merging this until after 2.8.1, but I'd also like it to be well examined, since it has a huge impact on anyone using outbound.

Please read the code and comment.

- Uses generic-pool much like smtp_client does

Checklist:
- [x] docs updated
- [x] tests updated


baudehlo:
I'm going to merge this in the hopes it inspires further testing.

PR #1468, opened 2016-05-18
Fix config overriding for core modules

### LINKED ISSUE #1467: Config override feature works only for plugins
Not sure if this is by design or a bug.

With the latest release of haraka, we get the config override feature where I can pass a config directory using `-i` and haraka will pick up configs from that directory and fallback to the default configs.

This feature only works for plugins from what I can tell. The core modules like `outbound.js` do not read the config from the haraka defaults and always read from the `-i` directory.

I debugged a little and reason is that plugins.js gets a config using `config.module_config`. The core modules, however, simple use config.js. The default instance of config.js has a `undefined` root_path and thus uses the cfreader.config_path which is the `-i` directory.

baudehlo:
Fixes #1467

### DIFF
--- a/config.js
+++ b/config.js
@@ -15,6 +15,9 @@ function Config (root_path) {
         }
         return cfg;
     };
+    if (process.env.HARAKA) {
+        this.overrides_path = path.join(__dirname, 'config');
+    }
 }
 
 Config.prototype.get = function(name, type, cb, options) {


baudehlo:
Well, 2.8.2 - you released 2.8.1 but changed it to 2.8.2 in changes for
some reason :)

msimerson:
2.8.1 was released to npm as 2.8.2, because an unnamed person got a step ahead of himself in the release process checklist. So next up is 2.8.3.

baudehlo:
Oh boy. :)

msimerson:
Yeah, it's nearly time someone writes haraka-release.sh, so we can automatically screw up the releases. ;-)

baudehlo:
I used to use shipit for perl modules. I expect there's something similar.

msimerson:
I've tried a number of them with perl, everything from cpan-release (very minimal) to dist::zilla (very maximal) and I'd much prefer something like:

```sh
sed -i .bak -e 's/Mon DD/May 18/' \
    && cd ../haraka-tld && ./update_tld_files && ./bump_version && npm release \
    && cd ../haraka && ./bump_version \
    && grunt versioncheck \
    && ./update_changes_from_git_history \
    && git tag -vN.N.N && git push --tags \
    && npm test \
    && git commit -m 'release version N.N.N' \
    && npm publish
```

Issue #603, opened 2014-06-11
feature request: graceful restart


echolet:
```kill -HUP `cat var/haraka.pid````

msimerson:
> Note that HUP is already used in Haraka for flushing the queue.
> But we could go with SIGUSR2.

If Haraka were running in cluster mode and we had safe disconnects, HUP could flush the queue **and** restart all the workers.

baudehlo:
Fixed now.

PR #1470, opened 2016-05-19
Graceful shutdown and restart

baudehlo:
Fixes (probably a number of tickets somewhere deep in the unknown)

- Adds bin/haraka --graceful to reload child processes gracefully
- On SIGINT, will shutdown gracefully
- Adds plugin.shutdown() method so plugins can clearInterval() any timers they have setup

Will add docs in another commit.

Checklist:
- [ ] docs updated
- [ ] tests updated


baudehlo:
Closes #603

PR #1482, opened 2016-05-26
Just use crypto for reseed random

### LINKED ISSUE #1356: Update reseed_rng to use seedrandom 2.4.2?
A [much newer](https://github.com/davidbau/seedrandom) version of seedrandom is available upstream. Should we update?

baudehlo:
Fixes #1356 .

- Makes reseed_rng use crypto.randomBytes
- Assumes openssl random generator is solid
- Probably more likely than some random JS being good


msimerson:
👍

baudehlo:
I actually think this plugin can probably be removed.

It was a bug in node v0.4, but I'm pretty sure it doesn't happen any more:

```
var cluster = require('cluster');

var numCPUs = require('os').cpus().length;

if (cluster.isMaster) {
  // Fork workers.
  for (var i = 0; i < numCPUs; i++) {
    cluster.fork();
  }

  cluster.on('exit', (worker, code, signal) => {
    console.log(`worker ${worker.process.pid} died`);
  });
} else {
  // Workers can share any TCP connection
  // In this case it is an HTTP server
  console.log(Math.random());
}
```
output
```
0.17349092005356148
0.030112427956487053
```

baudehlo:
But for now I'll merge it anyway. It seems harmless.

PR #1495, opened 2016-06-04
Fixing DKIM 

zllovesuki:
- Uses "UTF-8" in all toString
- Remove whitespaces where Sig is broken into multi lines
- Add CRLF to the end of body regardless of bodycanon (as described in http://dkimcore.org/specification.html)


zllovesuki:
In case people look back and question why CRLF is added regardless:

https://tools.ietf.org/html/rfc6376

For simple body:
```
The "simple" body canonicalization algorithm ignores all empty lines
   at the end of the message body.  An empty line is a line of zero
   length after removal of the line terminator.  If there is no body or
   no trailing CRLF on the message body, a CRLF is added.  It makes no
   other changes to the message body.  In more formal terms, the
   "simple" body canonicalization algorithm converts "*CRLF" at the end
   of the body to a single "CRLF".

   Note that a completely empty or missing body is canonicalized as a
   single "CRLF"; that is, the canonicalized length will be 2 octets.
```

For relaxed body:
```
Ignore all empty lines at the end of the message body.  "Empty
       line" is defined in Section 3.4.3.  If the body is non-empty but
       does not end with a CRLF, a CRLF is added.  (For email, this is
       only possible when using extensions to SMTP or non-SMTP transport
       mechanisms.)
```

PR #1499, opened 2016-06-06
Merge master into v3devel

Dexus:


baudehlo:
I'm happy to merge this, but this branch seems almost redundant right now. Thoughts?

Dexus:
Yeah the most is redundant, the graylisting plugin can i also merge into the master or I create a own external plugin for it. What your Thoughts?

baudehlo:
Is that different to our current plugins/greylist.js ?

Dexus:
The current greylist is only in v3devel a other is not in master if i'm correct remember. And i would build it as a external plugin for optional install.

Von meinem iPhone gesendet

baudehlo:
https://github.com/haraka/Haraka/blob/master/plugins/greylist.js ?

Dexus:
Ah yes that is the plugin, okay than we can delete v3devel.

Von meinem iPhone gesendet

baudehlo:
Closing.

PR #1501, opened 2016-06-07
Make rcpt_to.ldap work with aliases plugin

darkpixel:
- Instead of performing an LDAP lookup on the e-mail address specified by the client in the RCPT TO command, look at the value in the rcpt_to array which may have been modified by the aliases plugin.


darkpixel:
That last check appears to have failed due to something un-related to my PR.  Can someone check and let me know if I need to do anything?

baudehlo:
It's that damn host check timeout. The other Matt thinks there is likely a bug there so I haven't dived into it yet.

Issue #1498, opened 2016-06-05
CTRL+C doesn't kill Haraka


darkpixel:
### Haraka version
2.8.5

### Expected behavior
CTRL+C should stop new connections, drain existing connections, then exit.

### Observed behavior
CTRL+C prints 'SIGINT received', hangs until CTRL+C pressed again with no connections.

baudehlo:
Do you have a custom plugin that holds something open?

If nothing custom, please list your config/plugins here.

darkpixel:
This occurs with a bare minimum set of plugins:
```
root@collective:/opt/uitmail# haraka -c . -o
Returning boolean false for main.disabled=false
Returning boolean true for main.enable_tls=true
Returning boolean false for main.ipv6_enabled=false
Returning boolean true for main.always_split=true

Hook: init_master --------------------------------------------------------------
Plugin                              Method                              Prio T/O
--------------------------------------------------------------------------------
redis                               init_redis_connection                  0  30
process_title                       hook_init_master                       0  30

Hook: init_child ---------------------------------------------------------------
Plugin                              Method                              Prio T/O
--------------------------------------------------------------------------------
redis                               init_redis_connection                  0  30
process_title                       hook_init_child                        0  30

Hook: connect_init -------------------------------------------------------------
Plugin                              Method                              Prio T/O
--------------------------------------------------------------------------------
process_title                       hook_connect_init                      0  30

Hook: disconnect ---------------------------------------------------------------
Plugin                              Method                              Prio T/O
--------------------------------------------------------------------------------
process_title                       hook_disconnect                        0  30

Hook: data ---------------------------------------------------------------------
Plugin                              Method                              Prio T/O
--------------------------------------------------------------------------------
process_title                       hook_data                              0  30

Hook: log ----------------------------------------------------------------------
Plugin                              Method                              Prio T/O
--------------------------------------------------------------------------------
log.syslog                          syslog                                 0  30

Hook: lookup_rdns --------------------------------------------------------------
Plugin                              Method                              Prio T/O
--------------------------------------------------------------------------------
toobusy                             check_busy                             0  30

Hook: rcpt ---------------------------------------------------------------------
Plugin                              Method                              Prio T/O
--------------------------------------------------------------------------------
rcpt_to.in_host_list                hook_rcpt                              0  30

Hook: mail ---------------------------------------------------------------------
Plugin                              Method                              Prio T/O
--------------------------------------------------------------------------------
[...]

PR #1502, opened 2016-06-07
Shut down redis properly.  Closes GH-1498

darkpixel:
Fixes # 1498

- Properly shut down redis by calling quit() instead of end()
- Also quit server.notes.redis


msimerson:
Thanks @darkpixel

PR #1504, opened 2016-06-07
s/end/quit to properly shutdown redis client

darkpixel:
- Fix a few additional places where the redis client was not being properly shut down


baudehlo:
Thanks, merged.

PR #1506, opened 2016-06-08
DKIM: DKIMObject needs to have a copy of header_idx

zllovesuki:
Pass in a copy of header_idx to DKIMObject instead of header_idx directly.

Since DKIMObject mutates header_idx by using pop(), therefore, if the email has more than 1 DKIM-Signature, then the first DKIMObject will verify but then the later DKIMObject will not verify since there are no header_idx.

### DIFF
--- a/dkim.js
+++ b/dkim.js
@@ -41,6 +41,9 @@ function Buf() {
     };
 }
 
+function cloneHack(object) {
+    return JSON.parse(JSON.stringify(object));
+}
 
 ////////////////
 // DKIMObject //
@@ -53,7 +56,7 @@ function DKIMObject (header, header_idx, cb, timeout) {
     this.sig = header;
     this.sig_md5 = md5(header);
     this.run_cb = false;
-    this.header_idx = header_idx;
+    this.header_idx = cloneHack(header_idx);
     this.timeout = timeout;
     this.fields = {};
     this.headercanon = this.bodycanon = 'simple';
--- a/dkim.js
+++ b/dkim.js
@@ -52,7 +52,7 @@ function DKIMObject (header, header_idx, cb, timeout) {
     this.sig = header;
     this.sig_md5 = md5(header);
     this.run_cb = false;
-    this.header_idx = JSON.parse(JSON.stringify((header_idx));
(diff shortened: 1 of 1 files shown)


msimerson:
Would it be better to not use pop()?

zllovesuki:
@msimerson It is possible not to use pop(). The code in question that uses pop() is:
```javascript
// Now we canonicalize the specified headers
for (var h = 0; h < this.signed_headers.length; h++) {
    var header = this.signed_headers[h];
    this.debug(this.identity + ': canonicalize header: ' + header);
    if (this.header_idx[header]) {
        // RFC 6376 section 5.4.2, read headers from bottom to top
        var this_header = this.header_idx[header].pop();
        if (this_header) {
            // Skip this signature if dkim-signature is specified
            if (header === 'dkim-signature') {
                var h_md5 = md5(this_header);
                if (h_md5 === this.sig_md5) {
                    this.debug(this.identity + ': skipped our own DKIM-Signature');
                    continue;
                }
            }
            if (this.headercanon === 'simple') {
                this.verifier.update(this_header);
            } else if (this.headercanon === 'relaxed') {
                var hc = this.header_canon_relaxed(this_header);
                this.verifier.update(hc);
            }
        }
    }
}
```

Though pop() is more intuitive. One solution I can think of is to use a for-loop. Do you have a suggestion?

zllovesuki:
Actually, a for-loop is going to clutter the code a lot.

dkim.js
```diff
@@ -53,7 +56,7 @@ function DKIMObject (header, header_idx, cb, timeout) {
     this.sig = header;
     this.sig_md5 = md5(header);
     this.run_cb = false;
-    this.header_idx = header_idx;
+    this.header_idx = cloneHack(header_idx);
```
baudehlo: I'd honestly just inline the copy - it's a well understood idiom in JS now for copying.

zllovesuki:
Inlining is the way to go

msimerson:
> Actually, a for-loop is going to clutter the code a lot.

Then never mind me. :-). I wasn't looking at the code, just thinking that "if pop() has undesirable side effects, don't use pop".

PR #1511, opened 2016-06-10
Ssl shutdown bug

baudehlo:
I'm only PR'ing this to wonder why it never got merged, and see if it's potentially still an issue. I was just going through branches on GH and saw this and figured it deserved a look at.

### DIFF
--- a/tls_socket.js
+++ b/tls_socket.js
@@ -184,9 +184,11 @@ function createServer(cb) {
 
             var pair = tls.createSecurePair(sslcontext, true, true, false);
 
-            setInterval(function () {
-                if (pair.ssl.receivedShutdown()) {
+            var interval_timer = setInterval(function () {
+                if (pair.ssl.receivedShutdown) {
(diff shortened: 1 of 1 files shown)


Dexus:
This bug is 2 years old i have seen your report on node.js 0.10.2x release.

I cant find any doc about the shutdown...

Von meinem iPhone gesendet

baudehlo:
So it's fixed in newer versions of node?

Dexus:
I will tomorrow take a look, never have seen there a bug, so i don't know details now.

Von meinem iPhone gesendet

> Am 11.06.2016 um 01:12 schrieb Matt Sergeant <notifications@github.com>:
> 
> So it's fixed in newer versions of node?
>

Dexus:
So after my research, i can't find any open bugs that since your report on the old node repo. So it looks fixed.

baudehlo:
OK, closing and deleting branch.

PR #1512, opened 2016-06-11
Support dkim_sign with outbound.send_email()

### LINKED ISSUE #336: DKIM won't sign messages sent with outbound.send_email(...)
DKIM plugin listen to queue_outbound hook, so it won't sign messages that are sent through internal API: 

```
outbound.send_email(from, to, contents, outnext);
```

baudehlo:
Fixes #336

- Implements dkim_sign with mail sent via outbound.send_email()
- Adds new hook: `hook_pre_send_trans_email`.

Checklist:
- [X] docs updated
- [ ] tests updated

### DIFF
--- a/docs/Plugins.md
+++ b/docs/Plugins.md
@@ -157,6 +157,7 @@ These are the hook and their parameters (next excluded):
 * bounce (hmail, err) - called when an outbound message bounces
 * delivered (hmail, [host, ip, response, delay, port, mode, ok_recips, secured, authenticated]) - called when outbound mail is delivered
 * send\_email (hmail) - called when outbound is about to be sent
+* pre\_send\_trans\_email (fake_connection) - called just before an email is queued to disk with a faked connection object
 
 ### rcpt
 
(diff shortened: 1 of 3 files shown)


baudehlo:
I'd appreciate comments on the way I fixed this. It's a bit of a hack. Adds a new hook, though that hook could be useful for other features too, and we can probably improve on the hack over time.

@msimerson @smfreegard

msimerson:
I briefly looked it over. It's clever to simulate the connection so that dkim_sign needed no alternations to  to sign.

baudehlo:
Yeah it seems to work. I don't want people to assume it's a real connection object though - but I assume docs should take care of that enough.

msimerson:
> Yeah it seems to work.

If I used outbound, then I'd deploy and test this. Since you already have, I would just merge this and hope if it also works for the wider community. If not, we've got plenty more version numbers to get it fixed with.

(I still fall on the "release early, release often" side of the fence. I agree that bug-free releases are a great and noble goal, but shipping code **is** a form of testing.)

PR #1514, opened 2016-06-12
read file contents of plugin when plugin is a package

### LINKED ISSUE #1513: plugins as package don't get their hooks registered
### Haraka version

2.8.6

### Expected behavior

All plugins will have their hooks registered.

### Observed behavior

Plugins that are npm modules and/or packages have no hooks registered.

### Steps to reproduce

1. install haraka
2. start haraka with a plugin that's a package (I'm using `watch`)
3. enable log level logdebug so these messages are output:
````
[INFO] [-] [core] registered hook helo to limit.max_errors priority 0
[INFO] [-] [core] registered hook ehlo to limit.max_errors priority 0
[INFO] [-] [core] registered hook mail to limit.max_errors priority 0
[INFO] [-] [core] registered hook rcpt to limit.max_errors priority 0
[INFO] [-] [core] registered hook data to limit.max_errors priority 0
````
notice that the module-as-package has no registered hooks

### PR DESCRIPTION
**Pull request by msimerson**
Fixes #1513 

- load the file contents of plugins-as-npm-package same as plugin.js file

### DIFF
--- a/plugins.js
+++ b/plugins.js
@@ -172,8 +172,9 @@ Plugin.prototype._make_custom_require = function () {
 
 Plugin.prototype._get_code = function (pp) {
     var plugin = this;
+
     if (plugin.hasPackageJson) {
-        return 'exports = require("' + path.dirname(pp) + '");';
+        return 'var _p = require("' + path.dirname(pp) + '"); for (var k in _p) { exports[k] = _p[k] }';
     }
 
     try {

### CODE REVIEW

**Comment by baudehlo** [comment]
Won't this break some aspect of loading the plugin via require instead of the vm loader?

**Comment by msimerson** [comment]
> Won't this break some aspect of loading the plugin via require instead of the vm loader?

Right now, loading plugins as a package (at least in the watch plugin case) is effectively broken because the hooks are registered. With this patch, the hooks get registered and they actually work. 

If this change breaks something else, I'm not sure what that something else would be. In empirical testing, this fixes the problem I observed.

**Comment by baudehlo** [comment]
The way I see it is package plugins need to explicitly register hooks. The same way they need to explicitly load constants etc.

The aim was to remove magic. Doesn't this just put it back again?

**Comment by msimerson** [comment]
> The way I see it is package plugins need to explicitly register hooks. 

Is this explicitly registering a hook?

```js
exports.hook_init_http = function () {
}
````

Is this explicitly registering a hook?

```js
    this.register_hook('init_http', 'init_http');
}
exports.init_http = function () {
    ....
}
```

I think they are. But right now, neither way works.  If there's some other way to explicitly register them, please do tell. And put it in the `Plugins as Modules` section of Plugins.md.

baudehlo:
My assumption was that the latter should work. At a minimum. We should probably support the former too at least though - otherwise it would be too hard to move plugins to modules.

baudehlo:
Ah, here's a better fix which sticks to using require():

```
        return 'var _p = require("' + path.dirname(pp) + '"); for (var k in _p) { exports[k] = _p[k] }';
```

msimerson:
PR updated.

PR #1529, opened 2016-06-30
Make rcpt_to.access case insensitive

darkpixel:
- Compares entries on the whitelist and blacklist in a case-insensitive matter

### DIFF
--- a/plugins/rcpt_to.access.js
+++ b/plugins/rcpt_to.access.js
@@ -63,7 +63,7 @@ function _in_whitelist(connection, plugin, address) {
         connection.logdebug(plugin, 'checking ' + address + ' against ' +
             plugin.wl[i]);
 
-        if (plugin.wl[i] === address) {
+        if (plugin.wl[i].toLowerCase() === address.toLowerCase()) {
             return true;
         }
     }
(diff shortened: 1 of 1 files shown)


msimerson:
Returning OK is wrong. The access plugin determines if a recipient is black/whitelisted, not whether the address is valid.

plugins/rcpt_to.access.js
```diff
@@ -63,7 +63,7 @@ function _in_whitelist(connection, plugin, address) {
         connection.logdebug(plugin, 'checking ' + address + ' against ' +
             plugin.wl[i]);
 
-        if (plugin.wl[i] === address) {
+        if (plugin.wl[i].toLowerCase() === address.toLowerCase()) {
             return true;
```
msimerson: Instead of lower casing every entry in the whitelist on every connection, performance can be improved by lower casing all the file entries when the file is loaded. That's saves millions and millions of toLowerCase() operations at run time, and you only have to lower the user supplied username.

darkpixel:
Oops.  That was supposed to go into a branch I was playing with.  Give me a bit to get things cleaned up.  I'm slammed at work today because of the 3-day weekend.

msimerson:
Oh hey, I just noticed your editing the plugins/rcpt_to.access plugin, which is deprecated by the plugins/access plugin (which, BTW, reads the same config files and already handles the lower-case matching in a much much faster way).

darkpixel:
I've been dragging me feet on switching a few mail servers... ;)

PR #1535, opened 2016-07-12
outbound: Add results property to transaction

gramakri:
transaction.results is used widely by many plugins to store results.
The code in connection.js creates this property in init_transaction
but similar logic is missing in outbound code.

In the future, results property should move to the transaction constructor.
But this is complicated by the fact that ResultStore requres a connection
object and doing this properly requires a (risky) refactor.

(The specific crash here was that the dkim plugin tries to sign a bounce
mail and crashes when doing a txn.results.add)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 node_js:
 #   - "0.6"     # no longer supported by async
 #   - "0.8"     # no longer supported by iconv
-    - "0.10"
+#   - "0.10"
     - "4"
     - "5"
     - "6"
--- a/outbound.js
+++ b/outbound.js
@@ -27,6 +27,7 @@ var existsSync  = utils.existsSync;
 var FsyncWriteStream = require('./fsync_writestream');
 var generic_pool = require('generic-pool');
 var server      = require('./server');
+var ResultStore = require('./result_store');
 
 var core_consts = require('constants');
 var WRITE_EXCL  = core_consts.O_CREAT | core_consts.O_TRUNC | core_consts.O_WRONLY | core_consts.O_EXCL;
@@ -532,6 +533,7 @@ exports.send_trans_email = function (transaction, next) {
     };
 
     logger.add_log_methods(connection);
+    transaction.results = new ResultStore(connection);
 
     connection.pre_send_trans_email_respond = function (retval) {
         var deliveries = [];


gramakri:
eslint requires node 4 atleast it seems (per package.json).

msimerson:
> In the future, results property should move to the transaction constructor.

That would not work. There are plenty of SMTP connections that never have transactions and those have results. Not to mention, what would happen to all the results that were accumulated before the transaction started?

I'd have no problem dropping node 0.10 from our test suite.

gramakri:
> That would not work. There are plenty of SMTP connections that never have transactions and those have results. Not to mention, what would happen to all the results that were accumulated before the transaction started?

Ah no, my intent was not to move the existing `connection.results` that already stores non-transactional results. My comment was to add an *extra* results property to transaction. So, basically we will have _two_ results - one in transaction and one in connection. Note that the code is already this way and I am not suggesting anything new.

msimerson:
> one in transaction and one in connection. Note that the code is already this way and I am not suggesting anything new.

Ah, I misunderstood you. You are correct that there's one result store per connection, and there's also one result store per transaction.

msimerson:
Please also make this change so that the tests pass and I can merge this:

```diff
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 node_js:
 #   - "0.6"     # no longer supported by async
 #   - "0.8"     # no longer supported by iconv
-    - "0.10"
+#   - "0.10"
     - "4"
     - "5"
     - "6"
```

msimerson:
Upstream Node.js support is being dropped for 5.x soon too, I'd be more than half tempted to drop version 5 from our automated testing as well.

msimerson:
thanks @gramakri

swerter:
Hi
I'm having troubles with this change. I try to check all outgoing emails by rspamd and save the results in Elasticsearch. With this change, the results of outgoing emails are getting reset, which means the rspamd results are not saved in Elasticsearch. Any suggestions what to do?

msimerson:
Ohhhhh, getting reset...not good. Does wrapping this line:

```js
transaction.results = new ResultStore(connection);
```

with an if conditional like this fix it?

```js
if (!transaction.results) {
    transaction.results = new ResultStore(connection);
}
```

gramakri:
@baudehlo @msimerson Wondering how this can ever happen? Can a transaction object be reused?

swerter:
@msimerson Yes, adding the if condition works, the results are now correctly stored into elasticsearch.

PR #1545, opened 2016-07-20
update async to v2

msimerson:
# CAUTION

- update async to version 2
- see [caolan/async/CHANGELOG](https://github.com/caolan/async/blob/master/CHANGELOG.md)

### Todo

- [x] review async.* usage
- [ ] be confident upgrading doesn't break anything
- [x] merge

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   "dependencies": {
     "address-rfc2821"       : "*",
     "address-rfc2822"       : "*",
-    "async"                 : "~1.5.2",
+    "async"                 : "~2.0.0",
     "daemon"                : "~1.1.0",
     "generic-pool"          : "~2.4.1",
     "iconv"                 : "~2.2.0",
--- a/plugins/dkim_sign.js
+++ b/plugins/dkim_sign.js
@@ -170,7 +170,11 @@ exports.hook_queue_outbound = exports.hook_pre_send_trans_email = function (next
         return next();
     }
 
-    plugin.get_key_dir(connection, function(keydir) {
+    plugin.get_key_dir(connection, function (err, keydir) {
+        if (err) {
+            connection.logerror(plugin, err);
+            return next(DENYSOFT, "Error getting key_dir in dkim_sign");
+        }
         var domain;
         var selector;
         var private_key;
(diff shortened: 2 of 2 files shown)


baudehlo:
This should mostly be a great win. Especially for outbound performance on large queues. Nice.

baudehlo:
as far as I can tell from a quick grep, the one place that needs fixed is plugins/dkim_sign.js#231 which needs changed to:

```javascript
    async.filter(dom_hier, fs.exists, function(err, results) {
```

baudehlo:
Here's a diff to fix dkim_sign (untested):

```
diff --git a/plugins/dkim_sign.js b/plugins/dkim_sign.js
index 4739214..7b9a0bb 100644
--- a/plugins/dkim_sign.js
+++ b/plugins/dkim_sign.js
@@ -170,7 +170,11 @@ exports.hook_queue_outbound = exports.hook_pre_send_trans_email = function (next
         return next();
     }
 
-    plugin.get_key_dir(connection, function(keydir) {
+    plugin.get_key_dir(connection, function(err, keydir) {
+        if (err) {
+            connection.logerror(plugin, err);
+            return next(DENYSOFT, "Error in getting key_dir in dkim_sign");
+        }
         var domain;
         var selector;
         var private_key;
@@ -228,9 +232,17 @@ exports.get_key_dir = function (connection, cb) {
     }
     connection.logdebug(plugin, dom_hier);
 
-    async.filter(dom_hier, fs.exists, function(results) {
+    async.filter(dom_hier, function (file, cb) {
+        var result;
+        try {
+            cb(null, fs.exists(file));
+        }
+        catch (e) {
+            return cb(e);
+        }
+    }, function(err, results) {
         connection.logdebug(plugin, results);
-        cb(results[0]);
+        cb(err, results ? results[0] : null);
     });
 };
```

baudehlo:
Oops, nuke the `var results` line.

msimerson:
patch added to this PR

msimerson:
WARNING: merging this PR will require that anyone upgrading Haraka and using dkim_sign will have to do a 'npm install' in the Haraka install dir to upgrade async. Really, the `npm install` step should be done always, but in this case, it's a requirement.

PR #1547, opened 2016-07-21
normalized connection properties

### LINKED ISSUE #1098: Normalize the local and remote connection data

msimerson:
### work in progress

Fixes #1098

### connection.js changes *only* in #1577 

- converts connection.remote_* to connection.remote.*
- converts connection.local_* to connection.local.*
- converts connection.using_tls -> connection.tls.enabled
- converts connect.greeting -> connection.hello.verb
- maintains previous names, providing backwards compatibility

### Changes proposed in this pull request:

* updates all code to use the new normalized locations
* Updates all the connection.set invocations to use the new connection.set()

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -24,6 +24,7 @@ var outbound    = require('./outbound');
 var date_to_str = require('./utils').date_to_str;
 var indexOfLF   = require('./utils').indexOfLF;
 var ResultStore = require('./result_store');
+// var net_utils   = require('./net_utils');
 
 var hostname    = (os.hostname().split(/\./))[0];
 var version     = JSON.parse(
--- a/connection.js
+++ b/connection.js
@@ -24,7 +24,6 @@ var outbound    = require('./outbound');
 var date_to_str = require('./utils').date_to_str;
 var indexOfLF   = require('./utils').indexOfLF;
 var ResultStore = require('./result_store');
-// var net_utils   = require('./net_utils');
 
 var hostname    = (os.hostname().split(/\./))[0];
 var version     = JSON.parse(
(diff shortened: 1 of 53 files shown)


baudehlo:
This is going to break basically every application built on Haraka, so it needs some careful consideration. I know it'll break the Haraka mailing list, and emailitin, for example.

One way to mitigate some of that might be to do it in a backwards compatible way (have both sets of data at first). Ugly but doable.

msimerson:
Look at the patch, it *already* does it in a backwards compatible way. :-)  at least mostly. I ran out of time. It needs some more review before I can confidently say 100%.

baudehlo:
+1

connection.js
```diff
@@ -1140,14 +1173,14 @@ Connection.prototype.cmd_proxy = function (line) {
...
-        self.remote_ip = src_ip;
-        self.remote_port = parseInt(src_port, 10);
-        self.remote_host = undefined;
-        self.hello_host = undefined;
+        self.local.ip = dst_ip;
+        self.local.port = parseInt(dst_port, 10);
+        self.remote.ip = src_ip;
+        self.remote.port = parseInt(src_port, 10);
+        self.remote.host = null;
+        self.hello.host = null;
```
baudehlo: Surely these should use `.set()`?

connection.js
```diff
@@ -153,20 +154,48 @@ function setupClient(self) {
...
+        host: null,
+        verb: null,
+    };
+    this.tls = {
+        enabled: server.has_tls ? true : false,
+        verified: false,
+        cipher: {},
+        authorized: null,
+    }
+
```
baudehlo: Wouldn't it be better to use `.set()` here too?

msimerson: Better for whom? Surely having a bunch of function calls is not going to be faster than just declaring the object and a few properties. I don't think that using .set() makes the code more readable for humans. In what way do you think it'd be better?

PR #1550, opened 2016-07-26
AUTH PLAIN splitted into two lines

### LINKED ISSUE #1366: Response 334 after AUTH PLAIN
Hello,

Using Opera Mail (possibly affecting other clients too) I could not authenticate. After reviewing the logs, I noticed that this client does not send the username in the AUTH PLAIN request but afterwards. Haraka is supposed to [handle this case](https://github.com/haraka/Haraka/blob/master/plugins/auth/auth_base.js#L169). But the 334 response is never actually sent to the client because it contains only a whitespace. It looks like [this test](https://github.com/haraka/Haraka/blob/master/connection.js#L466) is to blame.

swerter:
Fixes #1366 .

- AUTH PLAIN divided in two lines: Server should respond with 334 code.

Checklist:
- [x] tests updated

Now also additional email clients can connect to Haraka, such as Gnome geary.

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -460,17 +460,18 @@ Connection.prototype.respond = function(code, msg, func) {
         code = msg.code;
         msg = msg.reply;
     }
-    if (!(Array.isArray(msg))) {
-        // msg not an array, make it so:
-        messages = msg.toString().split(/\n/).filter(function (msg2) {
-            return /\S/.test(msg2);
-        });
+    if (!Array.isArray(msg)) {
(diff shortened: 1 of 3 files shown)


baudehlo:
For some reason Travis-CI doesn't acknowledge this PR. So I'm going to do this to try and fix it:

http://stackoverflow.com/questions/19589260/force-travis-to-build-a-specific-pull-request-or-commit-sha-that-it-missed

baudehlo:
LGTM. +1

PR #1559, opened 2016-08-05
Rspamd: support for soft rejections, adding/removing headers & DKIM

fatalbanana:
Checklist:
- docs updated [YES]
- tests updated [NO: no tests for this plugin so far]

Feedback/incongruous demands/etc welcome.


smfreegard:
LGTM +1

msimerson:
Nice, I like this. But, instead of having foo.disabled toggles, invert their logic to use foo.enabled = true | false. It reads as much much better english and it's a best practice.

msimerson:
thanks @fatalbanana

PR #1562, opened 2016-08-10
Fix adding/removing headers in rspamd plugin; add never_add_headers

fatalbanana:
- Oops! We had some [bad documentation](https://github.com/vstakhov/rspamd/pull/836/files) RE: use of `task:set_rmilter_reply` and consequently Haraka plugin was expecting data in a wrong format. :(
- Also add `never_add_headers` option, which could be useful if you want to use custom headers from Rspamd, or just don't want to add headers for whatever reason.

Checklist:
- [x] docs updated
- [ ] ~~tests updated~~

### DIFF
--- a/config/rspamd.ini
+++ b/config/rspamd.ini
@@ -1,6 +1,9 @@
 ;host = localhost
 ;port = 11333
-;always_add_headers = false
+;add_headers = sometimes
+
+[dkim]
+;enabled = true
 
 [header]
 bar = X-Rspamd-Bar
@@ -15,6 +18,9 @@ score = X-Rspamd-Score
 ;message = Detected as spam
 ;authenticated=false
 ;spam = true
+
+[rmilter_headers]
+;enabled = true
 
 [spambar]
 ;positive = +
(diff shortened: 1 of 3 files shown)


Dexus:
LGTM

Von meinem iPhone gesendet

msimerson:
### header settings
* main.always_add_headers
* main.never_add_headers
* headers.enabled

It seems that getting the desired header behavior is...excessively complex. Maybe adding never_add_headers isn't the best choice here, and instead there should be an add_headers setting that accepts arguments: always, sometimes, never. 

That should simplify both the settings and the documentation.

fatalbanana:
If you don't mind the backwards-incompatible change we could definitely do better. Possible values for add_headers could be something like always|sometimes (default)|never|rspamd (where last case would replace headers.enabled).

msimerson:
If the new config option has a different name, it should be pretty easy to write a little config setter that maps old config settings to that new option. Then the change is backwards compatible. :-)

fatalbanana:
Right. :) Ok. I will try cook something up. ;)

fatalbanana:
Well- maybe someone would like to retain Haraka's headers while adding additional headers provided by rspamd- so I did not merge governance of rspamd-provided headers into `add_headers` setting. Renamed `headers.enabled` to `rmilter_headers.enabled` to try create less confusion. How do you feel about it now? ;)

msimerson:
👍

Issue #1558, opened 2016-08-03
enable plugins only for specific ports


gramakri:
I use a single haraka for inbound and outbound. This causes a bit of trouble with DNSBL which  blacklists IPs even when the incoming port is 587.

Wondering if any of the following two approaches sound useful:
* DNSBL plugin can have configuration to enable itself for specific incoming ports.
* A more generic plugin level approach which we can use in plugins file. Something like:
```
incoming: dnsbl # other values are 'relay', 'msa'

plugin2 # applies to all modes, like now

relay: plugin3
```

I haven't completely thought it through but I just opening this issue for suggestions/comments.

The other option, of course, is to say this is unsupported and run two instances of haraka.

smfreegard:
See the delay_deny plugin.  This is exactly what it is for.  Skip rejections until the recipients so that things like DNSBL can be skipped for AUTH/Relays.

gramakri:
@smfreegard ah! thanks a lot, looks like exactly what i want.

PR #1564, opened 2016-08-12
delay_deny: add included mode

gramakri:
Currently, this plugin delay denies everything unless listed in the
config file. This can be a little risky when experimenting this plugin
when a lot of other plugins are already in use.

In such cases, it's easier to use an included mode (a whitelist mode)
where one can add plugins to bypass one by one.


gramakri:
Made this patch when investigating #1558 based on @smfreegard 's suggestion.

@smfreegard @msimerson @baudehlo  comments?

PR #1566, opened 2016-08-16
Don't call next() at stream end

### LINKED ISSUE #1553: dkim_verify calls next before results are verified
### Haraka version
2.8.8

### Expected behavior
by the time the queue plugin is called the transaction notes should have the dkims results when a dkim signature is present.

### Observed behavior
when the queue plugin is run the dkim results may or may not be on the transaction notes. Based on the callback flow in the `dkim_verify.js` it appears the next hook is being called to soon. `next` should be called in the callback of `DKIMVerifyStream` instead of on `tx.message_stream.once('end')`

### Steps to reproduce
Log the presence of the DKIM-signature header and the tx.notes.dkim_results in a custom queue plugin and you'll see that some messages are queued before the tx.notes.dkim_results has been populated.

We've patched this issue in our systems by removing the `message_stream.once('end', next)` line of code and added calls to `next` into the DKIMVerifyStream callback ensuring queuing will not be called until dkim verification is complete.

baudehlo:
Fixes #1553 

- Call `next()` in the callback, not at stream end.

### DIFF
--- a/plugins/dkim_verify.js
+++ b/plugins/dkim_verify.js
@@ -17,8 +17,12 @@ exports.hook_data_post = function(next, connection) {
     var verifier = new DKIMVerifyStream(function (err, result, results) {
         if (err) {
             connection.logerror(self, 'error=' + err);
+            return next();
         }
-        if (!results) return;
+        if (!results) {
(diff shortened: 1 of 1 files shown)


msimerson:
LGTM.

PR #1568, opened 2016-08-17
rcpt_to.routes URI format w/ LMTP support

darkpixel:
- Allow rcpt_to.routes to use URI-formatted destinations so ```lmtp://``` vs ```smtp://``` can be specified

- [x] docs updated
- [ ] tests updated

Existing tests pass, but I'm not sure they are thorough enough.  I'm a bit lost in the rcpt_to.routes.js test file, so I haven't written specific tests for my changes yet.

### DIFF
--- a/docs/plugins/rcpt_to.routes.md
+++ b/docs/plugins/rcpt_to.routes.md
@@ -60,6 +60,11 @@ The [routes] section can include routes for domains and email addresses:
     matt@example.com=some.where.com
     spam@example.com=honeybucket.where.com:26
 
+You may also use URI format to specify SMTP vs LMTP:
+    [routes]
+    aaron@example.com=lmtp://mail.example.com:2525
+    matt@example.com=smtp://127.0.0.1:4242
+
 # Performance
 
 ## File based
--- a/docs/plugins/rcpt_to.routes.md
+++ b/docs/plugins/rcpt_to.routes.md
@@ -1,4 +1,4 @@
-# `rcpt_to.routes`
+# rcpt_to.routes
 
 Recipient Routes does two things: recipient validation and MX routing.
 
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "js-yaml"               : "~3.6.1",
     "nopt"                  : "~3.0.4",
     "npid"                  : "~0.4.0",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -25,6 +25,7 @@
     "js-yaml"               : "~3.6.1",
     "nopt"                  : "~3.0.4",
     "npid"                  : "~0.4.0",
+    "url-parse"             : "~1.1.3",
```
baudehlo: Just use node's `url` module, not a new npm module. It should work just fine.

docs/plugins/rcpt_to.routes.md
```diff
@@ -1,4 +1,4 @@
-# `rcpt_to.routes`
+# rcpt_to.routes
```
msimerson: @baudehlo, will this work now with the new html page rendering?

baudehlo: Maybe? I think so.

darkpixel: I didn't test it.  I looked at several of the other docs and assumed that was why it showed up in the plugins list as 'Haraka' instead of 'rcpt_to.routes'.  If that's not where the website generates the name, let me know and I'll change it.

baudehlo: Yeah the regexp is horribly buggy in the site builder:

```
    my ($title) = ($output =~ /<h1>([^<]*)/);
    $title ||= "Haraka";
```

So "# `plugin.name`" would get turned into <h1><code>plugin.name</code></h1>
and thus break the regexp. Gross.

I've asked for help on the site builder before but nobody comes forwards :(

Matt.

baudehlo:
FWIW I fixed the site builder now so all the Titles should be correct.

msimerson:
Besides @baudehlo's comment on using `uri` to avoid a dependency (a sentiment I agree with), this LGTM.

baudehlo:
Yep absolutely. Fix that and I'll merge.

darkpixel:
Don't merge this yet.  I stubbed out the replacement for url-parse, but haven't tested it in production yet.  I'll hit it tomorrow.

darkpixel:
Tested in production.  Works for me.

PR #1571, opened 2016-08-19
Allow streams in outbound.send_email contents

### LINKED ISSUE #1549: outbound.send_email accept stream
Hi.

We use Haraka for processing the email, make some changes, and send it again.

For doing that we use the `connection.transaction.message_stream.pipe`.
When we ended the modifications we send using `outbound.send_email`. However `outbound.send_email` only accepts a string, so we do something like
````js
outbound.send_email(connection.transaction.mail_from, connection.transaction.rcpt_to, message.toString(), function (code, msg) {});
````

As you see, we conver the stream to string, and you know that could be very memory intensive with large attachments.

Would you please allow send a stream?

Thanks!!

### PR DESCRIPTION
**Pull request by baudehlo**
Fixes #1549  .

- Allow send_email to use a stream in the `contents` object

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/line_socket.js
+++ b/line_socket.js
@@ -4,7 +4,7 @@
 var net  = require('net');
 var tls  = require('./tls_socket');
 var util = require('util');
-var line_regexp = /^([^\n]*\n)/;
+var line_regexp = require('./utils').line_regexp;
 
 function Socket(options) {
     if (!(this instanceof Socket)) return new Socket(options);
--- a/outbound.js
+++ b/outbound.js
@@ -431,6 +431,7 @@ function _fname () {
     return time + '_0_' + process.pid + "_" + _next_uniq() + '.' + my_hostname;
 }
 
+var line_regexp = /^([^\n]*\n?)/;
 exports.send_email = function () {
 
     if (arguments.length === 2) {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by ricardopolo** [comment]
Does not work :disappointed: 

image


The testing code was like this:
````js
mailcomposer(mail_options).build(function(err, messageStream){
   outbound.send_email(connection.transaction.mail_from, connection.transaction.rcpt_to, messageStream, function (code, msg) {
````

**Comment by baudehlo** [comment]
From the docs of mailcomposer, that's not how you get a stream. Your code should be:
```javascript
var stream = mailcomposer(mail_options).createReadStream();
outbound.send_email(connection.transaction.mail_from, connection.transaction.rcpt_to, stream, function (code, msg) {
```

ricardopolo:
@baudehlo not working yet.

Using this code:

````js
connection.loginfo("Before Stream");
var stream = mailcomposer(mail_options).createReadStream();
connection.loginfo("After Stream");
outbound.send_email(connection.transaction.mail_from, connection.transaction.rcpt_to, stream, function (code, msg) {
              connection.loginfo("After Send Email");
````


This is the output (After sending with the stream Haraka Stops, and never shows "After Send Email"")
image

baudehlo:
Can you try adding a stream.resume() to the end of the stream_line_reader
function?

ricardopolo:
Excuse me @baudehlo but I did not find a `stream_line_reader` function... any help?

baudehlo:
It's here in outbound.js:
https://github.com/haraka/Haraka/blob/c22548e7a784a7818388c7457e8cef39b89f5d69/outbound.js#L533

You are using the right branch I hope?

ricardopolo:
I did it. Same behavior

image

ricardopolo:
@ImLostInSpace the issue persists...

screen shot 2016-09-02 at 5 15 27 pm

baudehlo:
Yes the commit wasn't designed to fix anything. It was just lying around on
my disk waiting to be pushed.

I need to write some code to test this to figure out what's going on. Maybe
this weekend.

baudehlo:
Should work now.

ricardopolo:
Great!!! It works!!! Thank you so much!
:shipit:

PR #1572, opened 2016-08-19
Attempt to fix outbound shutdown error

### LINKED ISSUE #1570: Socket in pool got FIN
@baudehlo Removing the line configuration of  `pool_timeout=0`, in the latest version from master, just like the default config is I am getting this error.

"fin" in spanish means end, but I am not sure if that means the same in SMTP context :smile: 

image

baudehlo:
This is just a belt and braces attempt to stop Haraka shutting down from the error @ricardopolo reported here: https://github.com/haraka/Haraka/issues/1570

If it works: Fixes #1570

### DIFF
--- a/docs/Outbound.md
+++ b/docs/Outbound.md
@@ -78,7 +78,7 @@ Timeout for connecting to remote servers. Default: 30s
 * `pool_timeout`
 
 Outbound mail uses "pooled" connections. An unused pool connection will send
-a QUIT after this time. Default: 300s
+a QUIT after this time. Default: 50s
 
 Pooled connections means that a mail to a particular IP address will hold that
 connection open and use it the next time it is requested. This helps with
--- a/outbound.js
+++ b/outbound.js
@@ -73,7 +73,7 @@ exports.load_config = function () {
         cfg.connect_timeout = 30;
     }
     if (cfg.pool_timeout === undefined) {
-        cfg.pool_timeout = 300;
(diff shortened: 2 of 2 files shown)


ricardopolo:
In my testing it works :+1: :shipit:

ricardopolo:
I have some bad news... after more intensive testing I am seeing this exception.

I think the PR did something, the old error does not appear. But we have this new one
image

baudehlo:
Right, but does the server shut down after that?

ricardopolo:
Not @baudehlo the server did not shutdown, just show the error but continues working

baudehlo:
Yes that's expected.

ricardopolo:
Nice. But if this is a expected behavior I would change the log from Warn to Info. 

I dont think is nice to have lots of warns in the logs when are normal operations behaviors.

baudehlo:
Well it's a warning because the remote end really isn't supposed to do that. It's not very nice of them. But I can lower it to NOTICE perhaps.

baudehlo:
Following discussion on IRC, I've lowered the default timeout to 50s. It seems most servers have a longer timeout than that, so this should eliminate this warning.

ricardopolo:
Sadly the issue continues, I configure pool_timeout to 30 and we sill have this error.


screen shot 2016-09-22 at 2 56 02 pm

So would you change the log level please?
Thanks!

PR #1582, opened 2016-09-04
Check user-part aliases first

### LINKED ISSUE #1581: Aliases plugin not working as expected: order of rules
### Haraka version

 2.8.8

### Expected behavior

Given the rules

{
"@example.com" : { "action" : "alias", "to" : "bar@example.com" },
"foo" : { "action" : "alias", "to" : "foo@example.com" }
}

And the recipient foo-test@example.com

The expected replacement recipient is foo@example.com

### Observed behavior

The replacement recipient is bar@example.com

### Steps to reproduce

Send mail to foo-test@example.com

The reason is that the two if statements "if (config['@'+host]) {" and "if (config[user] || config[match[0]] || config[match[0] + '@' + host]) {" are in the wrong order, they should be reversed.

darkpixel:
Fixes #1581

- Test the user-part of the alias first instead of the domain part.  (i.e. most specific match to least specific match)

### DIFF
--- a/plugins/aliases.js
+++ b/plugins/aliases.js
@@ -34,23 +34,6 @@ exports.aliases = function (next, connection, params) {
         }
     }
 
-    if (config['@'+host]) {
-
-        action = config['@'+host].action || action;
-        match  = '@'+host;
-
-        switch (action.toLowerCase()) {
-            case 'drop':
-                _drop(plugin, connection, '@'+host);
-                break;
-            case 'alias':
-                _alias(plugin, connection, match, config[match], host);
-                break;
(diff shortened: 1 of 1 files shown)


baudehlo:
This PR needs more explanation as it seems like a big change to how things work (or I'm just interpreting that from the diff).

The best explanation could come in the form of a test.

msimerson:
Closing, as this currently won't merge and the suggested test additions haven't been made. It can be reopened in the future if these issues are addressed.

msimerson:
A change like this (with tests) has been merged.

Issue #1586, opened 2016-09-06
hook_deferred don't change delay with next(DENYSOFT, '3600'), throw error


streamtOtO:
### Haraka version
 2.8.8
### Expected behavior
 Call in  **hook_deferred** `next(DENYSOFT, '3600')`. Must delay email delivery for 3600 seconds.
### Observed behavior
Catch error `[CRIT] [687CEB36-7069-494E-B597-5A69DFFCA095.1.1] [outbound] Plugin bounce failed: TypeError: Cannot read property 'delay' of undefined
    at HMailItem.deferred_respond (/usr/lib/node_modules/Haraka/outbound.js:2072:23)
    at HMailItem.deny_respond (/usr/lib/node_modules/Haraka/plugins.js:558:39)
    at callback (/usr/lib/node_modules/Haraka/plugins.js:459:35)
    at Object.plugins.run_next_hook (/usr/lib/node_modules/Haraka/plugins.js:463:45)
    at Object.plugins.run_hooks (/usr/lib/node_modules/Haraka/plugins.js:402:13)
    at callback (/usr/lib/node_modules/Haraka/plugins.js:454:21)
    at Plugin.exports.hook_deferred (/opt/prodata-email-validation-2/haraka/plugins/bounce.js:13:5)
    at Object.plugins.run_next_hook (/usr/lib/node_modules/Haraka/plugins.js:486:28)
    at Object.plugins.run_hooks (/usr/lib/node_modules/Haraka/plugins.js:402:13)
    at HMailItem.temp_fail (/usr/lib/node_modules/Haraka/outbound.js:2063:13)
`
### Steps to reproduce
Call in  **hook_deferred** `next(DENYSOFT, '3600')`

### Maybe here error:
File Haraka/plugins.js, line [558](https://github.com/haraka/Haraka/blob/master/plugins.js#L558):
`object[respond_method](retval, msg)` call function with two parameters `(retval, msg)`, and in the file Haraka/outbound.js, line [2145](https://github.com/haraka/Haraka/blob/master/outbound.js#L2145), expect three parameters `HMailItem.prototype.deferred_respond = function (retval, msg, params)`, where `params` expected is ` object`  `{delay:'delay_in_seconds',err:'some message'}`

smfreegard:
> HMailItem.prototype.deferred_respond = function (retval, msg, params), **where params expected is object** {delay:'delay_in_seconds',err:'some message'}

next(DENYSOFT, { delay: 3600 });

[...]

PR #1591, opened 2016-09-08
Add third arguments params

streamtOtO:
Fixes [#1586](https://github.com/haraka/Haraka/issues/1586)

- Add third argument `params`.


baudehlo:
I kind of think the `constants.ok` (`else` part) probably should be passed the `params` object too. Probably not useful at all, but it keeps things consistent.

msimerson:
LGTM

PR #1592, opened 2016-09-12
Fix TypeError with null mail_from.host

soncodi:
Fixes intermittent TypeError in console, with no meaningful debug info.


msimerson:
Thanks @soncodi

PR #1610, opened 2016-09-26
Returning undefined instead of piping destination

thihara:
Fixes message stream piping issue.

When trying to pipe the message stream into a transforming stream (gzip in my case) the returning value is undefined.

This is because the short circuiting return statement lacks the necessary return value.


msimerson:
thanks @thihara

PR #1614, opened 2016-09-27
Keep results if it exists

baudehlo:
Fixes something discussed on #1535


gramakri:
@baudehlo just fyi, in case you didn't notice, there appear to be some changes unrelated to the issue.

baudehlo:
Yeah my bad. I'll fix that maybe in a separate PR.

PR #1618, opened 2016-09-29
Enforce line endings in streams

### LINKED ISSUE #1583: Blank spaces error
After using the new streaming option I am having delivery issues.

Some customers are reporting issues like this:

"Parece que su programa de correo electrónico agregó caracteres no válidos (caracteres de avance de línea vacíos) al mensaje cuando lo envió. Pruebe a usar un programa de correo diferente para enviar el mensaje. Si el problema persiste, envíe este mensaje de informe de no entrega (NDR) a su administrador de correo para obtener ayuda."

I would translate that as the program adds blank breaklines.
The code error is 550 5.6.11 .

Remeber I create the stream using mailcomposer (from nodemailer). When sending with `toString()` the issue does not happen.

baudehlo:
Fixes #1583 (maybe).

- Enforce line endings in streams on send_trans_email

### DIFF
--- a/outbound.js
+++ b/outbound.js
@@ -544,13 +544,13 @@ function stream_line_reader (stream, transaction, cb) {
             if (!(current_data.length || this_line.length)) {
                 return;
             }
-            transaction.add_data(new Buffer(this_line));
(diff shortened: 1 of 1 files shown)


baudehlo:
Closed because I don't think this fix is correct.

PR #1619, opened 2016-09-29
Loop protect plugin

Bramzor:
Fixes # .

- Added new plugin for loop protection

Checklist:
- [x] docs updated

### DIFF
--- a/docs/plugins/loop_protect.md
+++ b/docs/plugins/loop_protect.md
@@ -1,0 +1,10 @@
+# loop_protect
+
+This plugin places an header and protects against possible loops in case of bad configuration..
+
+
+## Usage
+
+*Optional* Hostname can be specified by using:
+
+echo '{"hostname": "server.yourdomain.com"}' > config/loop_protect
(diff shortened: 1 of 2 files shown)


baudehlo:
New plugins should go on npm. There's a couple of plugins up there you can copy the format of.

Bramzor:
Don't get it. A completely different git for a plugin that is only a few lines of code? Makes no sense. To much overhead for such a thing don't you think?

msimerson:
Hi @Bramzor . Although it's "just a few lines of code" to you, to us it's "yet another plugin in the big fat not-well-organized and lengthy-to-scroll-through list of plugins that we intend to eventually whittle away, moving each to their own npm module.

Seriously. It's become a big enough maintenance problem that we put a lot of effort into creating [haraka-test-fixtures](https://www.npmjs.com/package/haraka-test-fixtures) and [haraka-config](https://www.npmjs.com/package/haraka-config) so that plugins could become npm modules with config files and test suites of their very own.

PR #1621, opened 2016-09-30
Node 0.10 LTS support drops 10/1

msimerson:
- drop support for node 0.10

Checklist:
- [x] docs updated
- [x] tests updated


msimerson:
Also note, this requires the newly patched version of node.js 0.12 released yesterday, with the latest of the known openssl vulnerabilities patched.

PR #1622, opened 2016-09-30
update geoip for compat with newer ES

msimerson:
- adds a geo property to the result store for geoip plugin
- send the more explicit geo property to ES

Checklist:
- [x] docs updated

see caution note here: [https://www.elastic.co/guide/en/elasticsearch/guide/current/lat-lon-formats.html](https://www.elastic.co/guide/en/elasticsearch/guide/current/lat-lon-formats.html)

### DIFF
--- a/docs/plugins/log.elasticsearch.md
+++ b/docs/plugins/log.elasticsearch.md
@@ -132,7 +132,7 @@ curl -XPUT localhost:9200/_template/haraka_results -d '
                         "geoip" : {
                             "properties" : {
                                 "org"      : { "type" : "string", "index" : "not_analyzed" },
-                                "ll"       : { "type" : "geo_point" },
+                                "geo"      : { "type" : "geo_point" },
                                 "distance" : { "type" : "float" }
                             }
                         },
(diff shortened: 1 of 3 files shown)


msimerson:
Background info (for Google and posterity). Without this change, when saving to ES 2.0+, this error message will often be logged:

````
[ERROR] [$UUID] [log.elasticsearch] [mapper_parsing_exception] failed to parse
````

If one digs into it further and adds some appropriate logging, they can turn up this little nugget:

````
[INFO] [$UUID] [log.elasticsearch] { error: 
   { root_cause: [ [Object] ],
     type: 'mapper_parsing_exception',
     reason: 'failed to parse',
     caused_by: 
      { type: 'illegal_argument_exception',
        reason: 'illegal latitude value [120.97210000000001] for p.geoip.ll' } },
  status: 400 }
````

And that can be explained by a change in how ES parses an array of geo values (inverting lat/long, as explained at the URL above).

Anyone logging to Elasticsearch, who upgrades both Haraka and ES, will need to re-apply the index map template included in the docs for this plugin, so that ES gets the new explicitly defined `geo` hash.

PR #1630, opened 2016-10-03
Rspamd plugin: Fix spambar for negative scores

fatalbanana:
Fixes a warning about an unused variable that points at a bug: negative scores weren't processed correctly.

Also avoid adding an empty spambar when score is > -1 and < 1 (make this neutral instead).

### DIFF
--- a/plugins/rspamd.js
+++ b/plugins/rspamd.js
@@ -265,20 +265,18 @@ exports.add_headers = function (connection, data) {
 
     if (cfg.header && cfg.header.bar) {
         var spamBar = '';
-        if (data.score === 0) {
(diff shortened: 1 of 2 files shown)


msimerson:
Hey @fatalbanana , I have created a PR against your branch that adds some basic tests for rspamd. Right now, they'll fail on Haraka/master@HEAD, but they should pass on this branch of yours. Please merge them into your branch (which I think will update this PR).

Also, feel free to add more tests and validate that other portions of the rspamd plugin are working as expected.

PS: you may need to run `npm install haraka-test-fixtures` to install the latest version, which has added support for message header manipulation.

PR #1644, opened 2016-10-06
remove ./net_utils to haraka-net-utils

msimerson:

- removes ./net_utils to haraka-net-utils ([github](https://github.com/haraka/haraka-net-utils), [npm](https://www.npmjs.com/package/haraka-net-utils))
- updates plugins that require it
- replaces (5x) instances of logger.logerror with console.error (as haraka-net-utils can't load ./logger)
    - in these cases, the log messages were pretty low-value, but it'd be nice to detangle logger so it can be loaded by external libs

Checklist:
- [x] docs updated
- [x] tests updated


msimerson:
In addition to the test suite, I've now tested this in production for a few hours with no ill effect.

PR #1647, opened 2016-10-08
permit specifying haraka plugins w/o haraka-plugin- prefix

msimerson:

- permit specifying haraka plugins w/o `haraka-plugin-` prefix
- in config/plugins
- in resultstore

### DIFF
--- a/plugins.js
+++ b/plugins.js
@@ -77,14 +77,16 @@ Plugin.prototype._get_plugin_path = function () {
         paths.push(
             path.resolve(process.env.HARAKA, 'plugins', name + '.js'),
             path.resolve(process.env.HARAKA, 'plugins', name, 'package.json'),
-            path.resolve(process.env.HARAKA, 'node_modules', name, 'package.json')
+            path.resolve(process.env.HARAKA, 'node_modules', name, 'package.json'),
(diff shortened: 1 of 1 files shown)


msimerson:
There's an implicit question in this PR, mostly for @baudehlo . Which is, how are you loading [haraka-plugin-outbound-rate-limit](https://github.com/haraka/haraka-plugin-outbound-rate-limit)?  I presume the only way it can work (without the change in this PR) is that you are specifying 'haraka-plugin-outbound-rate-limit' in your `config/plugins` file. Which is redundant and kinda ugly, because then plugin.name is `haraka-plugin-outbound-rate-limit` instead of `outbound-rate-limit`.

I prefer this solution, of specifying only the plugin name, and having the `haraka-plugin-` prefix automatically prepended in the file search. That way `plugin.name` is set properly by default. If we remove the 3rd entry (which seems kinda not all that helpful), it would also enforce the `haraka-plugin` namespace prefix for plugins.

Thoughts?

baudehlo:
Fine by me. I am not using it. I wrote it as a proof of concept. And I don't know anyone using it anyway.

msimerson:
> And I don't know anyone using it anyway.

AFAIK, the watch plugin is the only such animal. So I'll package it up and separate and then path #3 in that list can go away.

PR #1651, opened 2016-10-10
More tls opts

typingArtist:
Added quite some important TLS security options.

- select location of key and cert
- enable cipher order based on server list
- add options for dhparam and EC curve selection
- allow selection of the OpenSSL "method"

[x] docs updated
[x] no tests needed updating


Dexus:
LGTM

Thank you for your PR.

Von meinem iPhone gesendet

msimerson:
thanks @typingArtist !

PR #1656, opened 2016-10-10
Update ISSUE_TEMPLATE.md

Dexus:
Added OS version to Issues

PR #1660, opened 2016-10-11
system info table format for issue reporting

msimerson:
See also, comments in #1656 

- add a shell script to collect system info for report issues
- format the data in a markdown table

### example

Haraka | Haraka.js — Version: 2.8.9
 --- | :--- 
Node | v6.7.0
| OS | FreeBSD haraka 10.3-RELEASE-p7 FreeBSD 10.3-RELEASE-p7 #0: Thu Aug 11 18:38:15 UTC 2016     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -2,7 +2,7 @@
 
 Please report your OS, Node version, and Haraka version by running this shell script on your Haraka server and replacing this section with the output.
 
-echo "Haraka | $(haraka -v)"; echo " --- | :--- "; echo "Node | $(node -v)"; echo "| OS | $(uname -a)"
+echo "Haraka | $(haraka -v)"; echo " --- | :--- "; echo "Node | $(node -v)"; echo "OS | $(uname -a)"; echo "openssl | $(openssl version)"
 
 ### Expected behavior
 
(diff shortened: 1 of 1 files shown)


Dexus:
Maybe it would be good also have the openssl version in it. (Because i see more and more problems with (for nodejs) outdated openssl versions.

msimerson:
> Maybe it would be good also have the openssl version in it.

Hrmm, a good idea. I worry it might be misleading.

`echo "Haraka | $(haraka -v)"; echo " --- | :--- "; echo "Node | $(node -v)"; echo "OS | $(uname -a)"; echo "openssl | $(openssl version)"`

Haraka | Haraka.js — Version: 2.8.9
 --- | :--- 
Node | v6.7.0
OS | FreeBSD haraka 10.3-RELEASE-p7 FreeBSD 10.3-RELEASE-p7 #0: Thu Aug 11 18:38:15 UTC 2016     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64

In this case, `openssl` is a "built-in" in /usr/bin and comes with the OS. However, most ports and packages don't use the built-in, they compile against the (usually  newer) port/package version installed in /usr/local/bin.

````
# /usr/local/bin/openssl version
OpenSSL 1.0.2j  26 Sep 2016
````

PR #1670, opened 2016-10-16
refactor plugins/tls a little

msimerson:
- move the tls_opts loading to its own function (it's grown quite a lot)
    - with test coverage
- add load_err(), to reduce redundancy
- give the hook functions better names:
    - tls_capabilities -> advertise_starttls
    - tls_unrecognized_command -> upgrade_connection
- within upgrade_connection, consolidate the two timeout functions into nextOnce()
- refactor the loginfo ternary into emit_upgrade_msg() with test coverage
- add additional tls options to tls.ini (as comments)

Checklist:
- [x] tests updated


msimerson:
> The better abstraction in plugin/tls.js looks really nice

Thanks. I've deployed this into production to watch it for a couple days and make sure I didn't mess anything up too badly. It would't hurt if you did so too. :-)

Also, thanks for the ssltest.sh script reference, that's a nice handy way to give those TLS underpinnings a workout.

Issue #1099, opened 2015-07-24
Add support for tls.TLSSocket


smfreegard:
Currently we use tls_socket.js which is a bit of a hack to allow the upgrade of a regular socket to a TLS encrypted socket when STARTTLS is called.

In Node 0.12 or greater and in io.js you can call `new tls.TLSSocket(socket, options)` where `socket` is an instance of `net.Socket` to upgrade the socket to a TLSSocket.  See https://iojs.org/api/tls.html#tls_class_tls_tlssocket

msimerson:
fixed by #1672

PR #1672, opened 2016-10-18
haraka/Haraka#1099 use TLSSocket instead of createSecurePair

typingArtist:
While this change looks like cosmetics, it is needed for advanced TLSSocket operations like OCSP Stapling. This also fixes long-lasting #1099.

Checklist:
- [x] no need for updating the docs
- [x] no need for updating tests

### DIFF
--- a/tls_socket.js
+++ b/tls_socket.js
@@ -137,15 +137,8 @@ pluggableStream.prototype.setTimeout = function (timeout) {
     return this.targetsocket.setTimeout(timeout);
 };
 
-function pipe(pair, socket) {
-    pair.encrypted.pipe(socket);
-    socket.pipe(pair.encrypted);
-
-    pair.fd = socket.fd;
-    var cleartext = pair.cleartext;
+function pipe(cleartext, socket) {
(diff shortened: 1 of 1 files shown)


msimerson:
While I'm really happy to see that crusty old tls_socket get some love, it might be better to just jettison that sucker. See #1099  (we can do that now b/c we recently dropped support for node 0.10).

typingArtist:
My aim was just to make it work with TLSSocket for allowing me to add OCSP Stapling functionality (see #1673).

If you don’t mind, I leave the refactoring to you.

smfreegard:
Why did you only change the server end to use TLSSocket and ignore the client?   Surely supporting OSCP stapling in both is desirable?

typingArtist:
Definitely it is. I just started with the one end I expected to be more work.

smfreegard:
Ok - no great.  Nice work on this.  I've wanted to support SMTP STS or DANE in Haraka for a while now, but it's reliant of Node getting support for DNSSEC and TLSA record types etc.

PR #1673, opened 2016-10-18
Ocsp stapling

typingArtist:
Adds support for OCSP Stapling in the Server. First shot: mandatory. Patch builds on #1672.

Checklist:
- [x] docs updated
- [ ] no clue for creating tests for that scenario
- [x] make it configurable (on/off)
- [x] use one OCSP cache for all running child processes
- [x] make the ocsp library an optional dependency


msimerson:
> no clue for creating tests for that scenario

Suggestion for informal test data to include:

```sh
testssl.sh -t smtp mail.my-host.com:25 > tls.before
testssl.sh -t smtp mail.my-host.com:25 > tls.after
diff before after
```

I presume you're running the tests anyway, so showing the difference it makes in the PR itself is a great "this is what this does" test.

typingArtist:
Well, OCSP Stapling requires a separate web server to provide certificate validity information which can be sent to the client as part of the ServerHello. That doesn’t make the task of testing any easier.

Let’s see how I can turn these https://github.com/indutny/ocsp/tree/master/test tests into something useful in Haraka.

smfreegard:
Note that the caching you're doing will only be local to each Haraka child process, so running with nodes=cpus will mean each child handling port 25 will have its own cache.

typingArtist:
Matt, are you fine with making the ocsp module a mandatory dependency in `package.json`? Otherwise I’d have to turn ocsp into a separate plugin, I guess.

typingArtist:
Steve, thanks for the catch. I will look into that as well.

smfreegard:
>  Matt, are you fine with making the ocsp module a mandatory dependency in package.json? Otherwise I’d have to turn ocsp into a separate plugin, I guess.

Personally - I think this should be an optional dependency given that OSCP stapling will add some overhead.   You wouldn't need to make it a separate plugin; you could just do something like this:

`````
// See if the ocsp library is available
var oscp;
var ocsp_available = false;
try {
    oscp = require('oscp');
    oscp_available = true;
}
catch (e) {
    logger.lognotice('OSCP library not available, OSCP stapling functions disabled');
}
`````

And then skip the OSCP code if `oscp_available` isn't true, or something like that.

@msimerson @baudehlo   Thoughts on this?  IMO; TLS should always be in core, not via an external plug-in, it's become too important not to have baked in support for.   OSCP stapling is a nice addition, but I feel it should be optional and it should probably work like we have done elsewhere e.g. disabled if the optional library isn't install and disabled if library is installed and the user has expressly disabled it in the config (and it should probably default off anyway).

typingArtist:
Thanks for the sample code, will add that in this way and make ocsp an optional dependency.

smfreegard:
Thanks - quick question about this (I've not played with it at all), but what happens if OSCP fails for some reason (e.g. the https call fails or the certificate is actually revoked)??

One issue we've had with Haraka's TLS support is that once you start the TLS negotiation, it's impossible currently to go back to a plain-text socket and continue without TLS.  Basically if the negotiation fails at the moment, you'll eventually hit the timeout and the connection will be dropped.

baudehlo:
+1 on making tls core. It's effectively core anyway since flags go on the connection object not in notes.

msimerson:
+1 for including this and the dependency.

If it causes issues for folks, we can always switch to optional in the future.

msimerson:
> what happens if OSCP fails for some reason

https://testssl.sh

Seriously, try it. :-)

typingArtist:
Well, it’s missing an option for enabling/disabling OCSP Stapling, so I have to touch the code anyway.

typingArtist:
@smfreegard , can you check whether the global cache introduced in d447e55 does the trick? Anyway, even if it still doesn’t do shared caching, it’s no big deal. The OCSP response is typically valid for a week, so even if each of the child processes has to fetch a new copy of the response, this will be a really rare action with less than little overhead.

typingArtist:
Guys, I’m a real node.js beginner. I have no clue how to test the server and the client functionality properly regarding TLS. I’d expect we need a test that actually creates a server on a port, creates an OCSP responder http server, and then performs real connections between the client and the server(s). How can we do that?

smfreegard:
> @smfreegard , can you check whether the global cache introduced in d447e55 does the trick? Anyway, even if it still doesn’t do shared caching, it’s no big deal. The OCSP response is typically valid for a week, so even if each of the child processes has to fetch a new copy of the response, this will be a really rare action with less than little overhead.

No - what you've done will still be local to the child process.  You'd need to put the cache into Redis (simplest, would also survive a reboot) or use cluster IPC to keep the cache in the master and have the children request it from the master (much harder, slower, still won't survive a reboot).

> Guys, I’m a real node.js beginner. I have no clue how to test the server and the client functionality properly regarding TLS. I’d expect we need a test that actually creates a server on a port, creates an OCSP responder http server, and then performs real connections between the client and the server(s). How can we do that?

@msimerson is the test guru around here ;-)

msimerson:
@typingArtist , due to merging other branches, this branch needs to be updated

typingArtist:
Let’s close this PR as a new branch is required anyway. Let’s track the feature in issue #1723.

PR #1674, opened 2016-10-18
fixed TLS client after server TLSSocket rework

typingArtist:
The TLS client connect code broke miserably through the introduction of TLSSocket into the server code. We still need the old `pipe()` code before the client gets updated for TLSSocket, too.

Checklist:
- [x] no docs update required
- [ ] We should have some STARTTLS tests, actually.


typingArtist:
Any volunteers for tests? I actually don’t know how to test the client with a live server. Or merge without TLS tests.

PR #1680, opened 2016-10-20
correct smtp_forward cfg for multiple rcpts

### LINKED ISSUE #1679: smtp_forward incorrect behaviour with multiple rcpts
### system info

Haraka 2.8.9.

### Expected behavior

When relaying an incoming message for two users on the same domain smtp_forward should send it as describe in smtp_forward.ini.

### Observed behavior

When receiving the message smtp_forward.get_config transform the first domain into the forwarded domain (in my case mail.company.com instead of company.com) and when matching with the second user it returns false (smtp_forward.js:48).

This leads to using the default target server (in my case- there's none).

### Fix

changing the line (smtp_forward.js:48) to be:

if (connection.transaction.rcpt_to[i].host !== dom) {

instead of

if (connection.transaction.rcpt_to[i].host !== dst_host) {

Works well for me

msimerson:
Fixes #1679 

- correctly compare destination hosts for multiple recipients
- add test coverage
- add config file wildcard boolean support
    - so arbitrary ini sections names like `example.com` can have boolean properties like `enable_tls`
- update tests to use haraka-test-fixtures.transaction()

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/configfile.js
+++ b/configfile.js
@@ -457,16 +457,20 @@ cfreader.load_ini_config = function(name, options) {
             if (match) {
                 is_array_match = regex.is_array.exec(match[1]);
                 if (is_array_match){
-                    setter = function(key, value){
+                    setter = function (key, value) {
                         key = key.replace('[]', '');
                         if (! current_sect[key]) current_sect[key] = [];
                         current_sect[key].push(value) };
                 }
                 else {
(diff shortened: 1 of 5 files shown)


godsflaw:
+1 on 🍵

PR #1682, opened 2016-10-20
use tls.connect for client

### LINKED ISSUE #1678: use modern TLS API for the client
Now that the server utilizes the modern `tls.TLSSocket()` API instead of the deprecated `tls.createSecurePair()`, the client code should upgrade to the modern API as well.

typingArtist:
Fixes #1678

[ ] add tests for config

### DIFF
--- a/outbound.js
+++ b/outbound.js
@@ -1700,10 +1700,8 @@ HMailItem.prototype.try_deliver_host_on_socket = function (mx, host, port, socke
                         process_ehlo_data();
                         break;
                     case 'starttls':
-                        var tkey = config.get('tls_key.pem', 'binary');
-                        var tcert = config.get('tls_cert.pem', 'binary');
-                        var tls_options = (tkey && tcert) ? { key: tkey, cert: tcert } : {};
-                        var config_options = ['ciphers','requestCert','rejectUnauthorized'];
+                        var tls_options = { servername: mx.exchange };
+                        var config_options = ['key','cert','ciphers','requestCert','rejectUnauthorized'];
 
                         for (var i = 0; i < config_options.length; i++) {
                             var opt = config_options[i];
@@ -1718,6 +1716,16 @@ HMailItem.prototype.try_deliver_host_on_socket = function (mx, host, port, socke
                                 tls_options[opt] = tls_config.outbound[opt];
                             }
                         }
+
+			if (Array.isArray(tls_options.key)) {
+			    tls_options.key = tls_options.key[0];
+			}
+			tls_options.key = config.get(tls_options.key, 'binary');
+
+			if (Array.isArray(tls_options.cert)) {
+			    tls_options.cert = tls_options.cert[0];
+			}
+			tls_options.cert = config.get(tls_options.cert, 'binary');
 
                         smtp_properties = {};
                         socket.upgrade(tls_options, function (authorized, verifyError, cert, cipher) {
--- a/outbound.js
+++ b/outbound.js
@@ -1717,15 +1717,15 @@ HMailItem.prototype.try_deliver_host_on_socket = function (mx, host, port, socke
                             }
                         }
 
-			if (Array.isArray(tls_options.key)) {
-			    tls_options.key = tls_options.key[0];
-			}
-			tls_options.key = config.get(tls_options.key, 'binary');
+                        if (Array.isArray(tls_options.key)) {
+                            tls_options.key = tls_options.key[0];
+                        }
+                        tls_options.key = config.get(tls_options.key, 'binary');
 
-			if (Array.isArray(tls_options.cert)) {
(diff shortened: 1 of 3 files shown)


msimerson:
Tested? (manually?)

Ready to merge?

typingArtist:
Tested but had trouble with `rejectUnauthorized`. It seems as if this feature now correctly barks on a bad certificate what it never did with the old code. However, the new code raises an exception. Quite strange behavior in the old code, but similarly bad in the new one if the server stops through this.

I tried this by addressing an email (of course using swaks :smile:) through Haraka to an alternative name of my current (non-Haraka) MX host. The cert is given for ##bbb##, the email is addressed to user@##aaaa##.  Haraka happily receives the mail, and it starts outbound. On upgrading to TLS, the connection hangs up, as expected, and the reason is correct, but an exception is thrown, not catched, and Haraka dies.

```
[PROTOCOL] [844170D2-B862-44C3-AA21-322E0EC7DCA1.1.1] [outbound] S: 250-ENHANCEDSTATUSCODES\r\n
[PROTOCOL] [844170D2-B862-44C3-AA21-322E0EC7DCA1.1.1] [outbound] S: 250 HELP\r\n
[PROTOCOL] [844170D2-B862-44C3-AA21-322E0EC7DCA1.1.1] [outbound] C: STARTTLS
[PROTOCOL] [844170D2-B862-44C3-AA21-322E0EC7DCA1.1.1] [outbound] S: 220 2.0.0 try it\r\n
[DEBUG] [-] [core] client TLS upgrade in progress, awaiting secured.
[PROTOCOL] [844170D2-B862-44C3-AA21-322E0EC7DCA1.1.1] [outbound] C: EHLO ######
[ERROR] [844170D2-B862-44C3-AA21-322E0EC7DCA1.1.1] [outbound] Ongoing connection failed to 192.168.11.2:25 : Error: This socket is closed
[DEBUG] [-] [core] [outbound] release_client: 192.168.##.##:25 to undefined
[DEBUG] [-] [core] [outbound] destroying pool entry for 192.168.##.##:25
[DEBUG] [-] [core] Temp fail for: Tried all MXs
[DEBUG] [844170D2-B862-44C3-AA21-322E0EC7DCA1.1.1] [outbound] running deferred hooks
[INFO] [844170D2-B862-44C3-AA21-322E0EC7DCA1.1.1] [outbound] Temp failing 1477030756478_0_12878_892.##### for 64 seconds: Tried all MXs
[CRIT] [-] [core] Error: Hostname/IP doesn't match certificate's altnames: "Host: ##aaa##. is not in the cert's altnames: DNS:##bbb##"
[CRIT] [-] [core]     at Object.checkServerIdentity (tls.js:199:17)
[CRIT] [-] [core]     at TLSSocket.<anonymous> (_tls_wrap.js:1066:29)
[CRIT] [-] [core]     at emitNone (events.js:91:20)
[CRIT] [-] [core]     at TLSSocket.emit (events.js:185:7)
[CRIT] [-] [core]     at TLSSocket._finishInit (_tls_wrap.js:584:8)
[CRIT] [-] [core]     at TLSWrap.ssl.onhandshakedone (_tls_wrap.js:416:38)
```

So in short: Not ready for merge, I need support. How can I avoid this exception to happen and continue with correct NDN processing?

typingArtist:
@msimerson, can you check 8b0f410, is this the right solution?

msimerson:
> is this the right solution

I haven't had a chance to look into it fully **but** that **is** a common issue, and adding an error handler **like** that one is usually The Right Thing[TM].

typingArtist:
Thanks. Do you already have an idea how to test Haraka client/server interaction (properly, gaining coverage at the same time)?

msimerson:
> Do you already have an idea how to test Haraka client/server interaction

Two ways:

1. refactor a bit. Move of bunch of code you have into a big if conditional into a function. When your if hits, call the function instead of running a large block of inline code. That function can now be exported, and you can pass in mock arguments to it and test the code to see that it does what you want.

2. a nested series of callbacks. Basically, mock a config and start up a server. When the server is up and running, it runs a callback which initiates a client. When the client succeeds or fails, it calls it's callback and the tests evaluate the results.

There's a tiny bit of coverage in tests/server.js.

msimerson:
Hi @typingArtist , [here's an example](https://github.com/haraka/Haraka/pull/1690/files#diff-0610ac4c7da6258a441a4f36a6358653L1703) of what I mean by moving code into functions, for testability. Exporting that function enabled me to [add these tests](https://github.com/haraka/Haraka/pull/1690/files#diff-6de221e127902dd5fb87f81d9c626b2cR45) to validate the code is doing what we need.

typingArtist:
Thanks, @msimerson, I’ll try that soon. 👍

PR #1692, opened 2016-10-26
haraka/Haraka#1690 fix outbound to avoid recursive reading key/cert after refactoring

typingArtist:
Fixes haraka/Haraka#1690

I’m getting `TypeError: Path must be a string.` as reported by others. The attached PR fixes the bug introduced by refactoring in #1690, with that change we don’t need to read the key inside the STARTTLS routine because it was already loaded.


msimerson:
I swear I removed that. I must have buggered it that later when rebasing. Thanks for fixing it.

PR #1697, opened 2016-11-08
Fix bad formatting in rcpt_to.routes plugin doc

darkpixel:
Massively small commit, but the bad formatting on the page has been annoying me.  ;)


msimerson:
Thanks!

PR #1698, opened 2016-11-10
move log.syslog to haraka-plugin-syslog

msimerson:
- replaces plugins/log.syslog.js with haraka-plugin-syslog
- updates `haraka -h` to show docs for npm packaged plugins

Checklist:
- [x] docs updated
- [x] tests updated


smfreegard:
I'm -1 on this.   IMO - syslog logging should be in core.

msimerson:
Syslog is still in "core" because it's still included as a dependency. This is primarily a packaging difference, and the most significant things this affects is whether syslog unit tests are run for every Haraka PR - vs - only when changes are made to the syslog plugin.

If we can get the few things that have compile dependencies (syslog, ldap, ??) packaged as NPM modules, then we don't need to install all the build tools for every Haraka CI (travis) build. That could reduce our build times quite a bit.

smfreegard:
> Syslog is still in "core" because it's still included as a dependency. This is primarily a packaging difference, and the most significant things this affects is whether syslog unit tests are run for every Haraka PR - vs - only when changes are made to the syslog plugin.

From my point of view - that isn't desirable.  I'd consider syslog core functionality and therefore I want tests to run on each pull.   For example - I could edit the logging hook in plugins.js and completely break all the logging plugins and I wouldn't know until I tested an install.   Surely that's a bad thing...

msimerson:
> I could edit the logging hook in plugins.js and completely break all the logging plugins and I wouldn't know until I tested an install. Surely that's a bad thing...

This plugin is a lousy example for that argument because none of the tests we have for the syslog plugin would reveal that breakage (we have only unit tests, no integration tests).

Whether this change is a bad thing depends on if you're the author of a logging plugin that's in core versus the author of a plugin that's not. If we break logging plugins, then it's best for every other plugin author that we've dog fooded it on something important like syslog, so that we get it fixed sooner and cause less harm to the more obscure logging plugins that aren't in core.

msimerson:
> I could edit the logging hook in plugins.js and completely break all the logging plugins

Followup: before **and** after this change, we have no tests in place now that would catch this type of bug. The common solutions for catching errors like this are test deployments and integration tests. The latter are much preferred because they can be [eventually] automated with Travis.

smfreegard:
Ok - having just chatted to baudehlo about this on IRC as well.   I can see your point now.  +1

msimerson:
FYI, I'm waiting to merge this until after I've done a deployment test.

PR #1702, opened 2016-11-11
Rspamd plugin: restore spam report header

### LINKED ISSUE #1563: Rspamd report header doesn't work
### Haraka version

Tested git but should affect released versions as well (probably any featuring rspamd plugin).
### Expected behavior

If headers are added (because `always_add_headers` is set or rspamd recommended `add header`)- `X-Rspamd-Report:` should contain a list of matched symbols and their scores.
### Observed behavior

`X-Rspamd-Report:` is empty.
### Explanation

Code is looking for a `score` property but nothing it sees is actually an object. Possible solutions are:
- Changing format of cleaned data to have symbols grouped in a nested object (is this bad? seems bad)
- Returning unfiltered data in addition to cleaned data used for logs etc (maybe good idea?)
- Tracking which entries are actual symbols in another object (hmmm)
### Aside (EDIT)

During the short period in which this did work, it was pretty ugly- and could likely exceed length limits- it should be wrapped. Adding symbol 'options' could be nice too . :)

fatalbanana:
Fixes #1563

Maybe I should just PR this. I planned to try make this header look better but I've not gotten round to that (nor other things in mind for Haraka & rspamd plugin) yet. @msimerson doesn't like having stuff nested under `default` and well, it is strange, but the idea is to really have the raw response from rspamd which makes this necessary evil. :)

- Restore spam report header
- Round down scores for spambar
- Restore untampered rspamd JSON
- Logs should be unaffected

Checklist:
- [N/A] docs updated
- [YEBO] tests updated

### DIFF
--- a/plugins/rspamd.js
+++ b/plugins/rspamd.js
@@ -148,41 +148,43 @@ exports.hook_data_post = function (next, connection) {
             res.on('end', function () {
                 var r = plugin.parse_response(rawData, connection);
                 if (!r.data) return callNext();
-                r.data.emit = true; // spit out a log entry
+                if (!r.data.default) return callNext();
+                if (!r.log) return callNext();
+                r.log.emit = true; // spit out a log entry
 
                 if (!connection.transaction) return callNext();
-                connection.transaction.results.add(plugin, r.data);
+                connection.transaction.results.add(plugin, r.log);
                 connection.transaction.results.add(plugin, {
                     time: (Date.now() - start)/1000,
                 });
 
                 function no_reject () {
-                    if (cfg.dkim.enabled && r.controlData['dkim-signature']) {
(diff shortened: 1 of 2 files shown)


msimerson:
> the idea is to really have the raw response from rspamd

Right, so instead of tucking it under an ambigious name like _default_, instead choose something descriptive like `rawResult` or some such.

fatalbanana:
Thing it's tucked under is called `data` (vs `log` which contains cleaned data). `default` is part of the response from rspamd.

smfreegard:
It also wouldn't be called `default` if rspamd 'settings' were used e.g. per-user, per-domain etc. would it?

fatalbanana:
My understanding is that non-default metrics are never actually used in rspamd and that `setttings` module has supplanted this functionality.

PR #1710, opened 2016-11-13
dkim is break : ds.exists return undefined use existsSync

NP6:
Fixes #        dkim is break
 
	fs.exists return undefined, use existsSync
        fs.exists is deprecated since: v1.0.0
 
-            cb2(null, fs.exists(file));
+            cb2(null, fs.existsSync(file));


smfreegard:
existsSync is not the correct fix and this has already been fixed in another PR.

PR #1716, opened 2016-11-14
Fix connection.remote_host and NXDOMAIN result

smfreegard:
The legacy property connection.remote_host was broken by https://github.com/haraka/Haraka/commit/8e412b10f2a3e646165801d6e519028f7e212563 and was undefined if the DNS lookup returned an error.

Additionally, `dns.NXDOMAIN` was incorrect - as per the Node docs, this should be `dns.NOTFOUND`, so it would appear this has been broken since day one.

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -728,8 +728,12 @@ Connection.prototype.lookup_rdns_respond = function (retval, msg) {
 Connection.prototype.rdns_response = function (err, domains) {
     if (err) {
         switch (err.code) {
-            case dns.NXDOMAIN: this.remote.host = 'NXDOMAIN'; break;
-            default:           this.remote.host = 'DNSERROR'; break;
+            case dns.NOTFOUND:
+                this.set('remote', 'host', 'NXDOMAIN');
+                break;
+            default:
+                this.set('remote', 'host', 'DNSERROR');
+                break;
         }
     }
     else {
--- a/connection.js
+++ b/connection.js
@@ -728,6 +728,7 @@ Connection.prototype.lookup_rdns_respond = function (retval, msg) {
 Connection.prototype.rdns_response = function (err, domains) {
     if (err) {
         switch (err.code) {
+            case dns.NXDOMAIN:
             case dns.NOTFOUND:
                 this.set('remote', 'host', 'NXDOMAIN');
                 break;


msimerson:
Actually, `NXDOMAIN` -> `NOTFOUND` is a rather recent node.js change, so I'd instead have BOTH listed in that case statement. IE:

```js
    case dns.NXDOMAIN:
    case dns.NOTFOUND:
        your_stuff();
```

smfreegard:
Are you sure about that?  I actually looked through the source code of Node 0.10.x and couldn't find NXDOMAIN was valid.

msimerson:
> Are you sure about that?

It wouldn't have come from the node.js source, it was coming from c-ares. There's some big long threads in the node.js docs discussing this change. I give this change a +1 in that thread sometime ago, so yeah, I'm pretty sure.

msimerson:
Also, "oops." Sorry to have broken remote_host setting. Missed that one...

smfreegard:
I've double checked and NXDOMAIN has either never been valid or it was valid on Node < 0.8 as:

`````
[root@relay1 ~]# echo "var dns = require('dns'); console.log(process.version); console.log('NXDOMAIN => ' + dns.NXDOMAIN); console.log('NOTFOUND => ' + dns.NOTFOUND);" | node
v0.8.16
NXDOMAIN => undefined
NOTFOUND => ENOTFOUND
`````

> Also, "oops." Sorry to have broken remote_host setting. Missed that one...

No worries.   It had me doubting my own sanity until I realised what the issue was.

msimerson:
I just double checked too, and the [thread in question](https://github.com/nodejs/node/issues/3891) is actually replacing the node.js customized error (NOTFOUND) with the `actual` IETF specified DNS error (from c-ares). So I had the direction of the change inverted. That thread is currently stalled, but I'd put that case in there now with this PR because it's likely to match in the future.

msimerson:
Amusingly, there's [quite](https://github.com/haraka/Haraka/blob/master/plugins/connect.fcrdns.js#L171) a [few](https://github.com/haraka/Haraka/blob/master/spf.js#L175) places in our code where we've (at _least_ you and I) have been using both for quite a long time.

msimerson:
Ooooohhhhhhhhhhhh. One reason that we'd have gotten NXDOMAIN results sometimes and NOTFOUND errors at other times is because of the use of dns.lookup (which uses getaddrinfo) vs dns.anythingElse which uses c-ares.

We (haraka) have long avoided using dns.lookup, but that likely explains why we've got 22 instances of NXDOMAIN in our source code.

smfreegard:
I think in my case - I always used the rdns_respond function as a boilerplate and when I had issues I added 'ENOTFOUND' instead and bypassed dns.NOTFOUND altogether.

Anyway - I'll add `dns.NXDOMAIN` back in, to preempt the Node change (thanks for the reference).

msimerson:
thanks @smfreegard

PR #1717, opened 2016-11-15
Correct Auth-Results cleaning

KevinMartin:
`cmd_data` has yet to receive any headers...

Correct place would be once the headers (or the full data) is fully collected, hence `data_done`.

It also doesn't affect the message size, just like before.


msimerson:
thanks @KevinMartin

PR #1718, opened 2016-11-15
remove utils.uniq()

msimerson:
- remove utils.uniq()
    - it's unused
    - implementation is poor. It only works on ordered arrays

(discovered while adding test coverage)


msimerson:
this PR obsoleted by #1720

PR #1724, opened 2016-11-19
Implement OCSP Stapling (#1723)

### LINKED ISSUE #1723: Implement OCSP Stapling for the server
What has been started as PR #1673 could not be merged after some heavy refactoring of the TLS code base. Let’s start the process anew with a properly registered issue and a mergable PR.

OCSP Stapling is a feature where, upon requested by the SMTP client, the SMTP server sends an additional short-lived ‘certificate certificate’ to the client to prove that its long-lived certificate is still valid. The benefits of this are that the checking is quite easy and fast to do by the client who doesn’t require additional http lookups for the OCSP responses, and that the load for the actual OCSP server is significantly reduced as only the SMTP server will request OCSP messages occasionally. Also note that we will see certificates in the future which adhere to RFC7633, the so-called ‘OCSP Must-Staple’ certificate extension which requires a client to terminate the connection if the server doesn’t send a stapled OCSP response.

OCSP is available as a node module, for both client and server logic. This issue only affects the server logic, so Haraka can offer OCSP Stapling to SMTP clients. The client part will be implemented as a separate feature in another issue.

The necessary changes to haraka-net-utils has been implemented already, allowing the new option `enableOCSPStapling` to be recognized and transported to the tls module.

typingArtist:
Adds support for OCSP Stapling in the Server (Fixes #1723), reworked and rebased #1673.

Checklist:
- [x] docs updated
- [x] make it configurable (on/off)
- [x] make the ocsp library an optional dependency

### DIFF
--- a/config/tls.ini
+++ b/config/tls.ini
@@ -12,7 +12,7 @@ ciphers=ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-G
 ; honorCipherOrder=false
 ; rejectUnauthorized=false
 ; requestCert=true
-
+; enableOCSPStapling=false
 
 
 [redis]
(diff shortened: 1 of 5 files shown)


celesteking:
Cool! Thanks.

Issue #1696, opened 2016-11-08
crit when redis server dead/missing


msimerson:
### system info

Haraka | Haraka.js — Version: 2.8.10
 --- | :--- 
Node | v6.7.0
OS | FreeBSD haraka 10.3-RELEASE-p7 FreeBSD 10.3-RELEASE-p7 #0: Thu Aug 11 18:38:15 UTC 2016     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64
openssl | OpenSSL 1.0.1s-freebsd  1 Mar 2016

### Expected behavior

Graceful degradation. When redis connections goes away, it is detected and reconnections are attempted.

### Observed behavior

````
[CRIT] [085A7D2F-D8B6-4267-9AD9-F5539D4D3CA6] [core] Plugin karma failed: TypeError: Cannot read property 'hgetall' of undefined
    at Plugin.exports.check_asn (/usr/local/lib/node_modules/Haraka/plugins/karma.js:873:14)
    at Plugin.exports.hook_connect (/usr/local/lib/node_modules/Haraka/plugins/karma.js:434:16)
    at Object.plugins.run_next_hook (/usr/local/lib/node_modules/Haraka/plugins.js:496:28)
    at callback (/usr/local/lib/node_modules/Haraka/plugins.js:452:32)
    at /usr/local/lib/node_modules/Haraka/plugins/dnsbl.js:149:16
    at zonesDone (/usr/local/lib/node_modules/Haraka/plugins/dns_list_base.js:126:9)
    at /usr/local/lib/node_modules/Haraka/node_modules/async/dist/async.js:952:25
    at iteratorCallback (/usr/local/lib/node_modules/Haraka/node_modules/async/dist/async.js:997:17)
    at /usr/local/lib/node_modules/Haraka/node_modules/async/dist/async.js:847:20
    at /usr/local/lib/node_modules/Haraka/plugins/dns_list_base.js:122:13
````

### Steps to reproduce

* Set up Haraka with redis.
* Watch Haraka work nicely.
* Stop the redis server.
* Watch Haraka work poorly.

PR #1727, opened 2016-11-20
redis connection error handling

msimerson:
- add redis client.on('end') event handler
- user newer redis client options.retry_strategy instead of legacy timeout

Checklist:
- [x] docs updated
- [x] tests updated

(This PR is mainly to give up on Redis faster and fail fewer tests when redis is not running, like on a dev workstation)

### DIFF
--- a/plugins/redis.js
+++ b/plugins/redis.js
@@ -37,24 +37,31 @@ exports.load_redis_ini = function () {
 exports.init_redis_connection = function (next, server) {
     var plugin = this;
 
-    // this is the server-wide redis, shared by plugins that don't
-    // set a specific db ID.
-    if (server.notes.redis && server.notes.redis.ping()) {
-        plugin.loginfo('already connected');
-        return next(); // connection is good
-    }
-
     var calledNext = false;
-    function callNext () {
+    function nextOnce () {
         if (calledNext) return;
         calledNext = true;
         next();
     }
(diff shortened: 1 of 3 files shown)


plugins/redis.js
```diff
@@ -37,24 +37,31 @@ exports.load_redis_ini = function () {
...
     }
 
-    var opts = plugin.redisCfg.opts;
-    opts.host = plugin.redisCfg.server.host;
-    opts.port = plugin.redisCfg.server.port;
-    server.notes.redis = plugin.get_redis_client(opts, callNext);
+    // this is the server-wide redis, shared by plugins that don't
+    // set a specific db ID.
+    if (server.notes.redis) {
+        server.notes.redis.ping(function (err, res) {
```
msimerson: That wouldn't do the same thing. The sticky part is doing The Right Thing under several circumstances:

1. redis works
2. no redis at all (Appveyor, local devs running `./run_tests`)
3. a redis that recently stopped responding (#1696)

The failure behavior is different for each. For tests, we want the client to give up quickly and scurry along, whereas in production we want it to retry and gracefully degrade after some period.

I did add a check in one of the tests that uses redis.connected and when I get around to #1696, I'll do some empirical testing to see how far I'm willing to trust that property (vs getting a response back from the server).

PR #1738, opened 2016-12-08
Misc. performance improvements

smfreegard:
- Fixed bug in config cache that was causing non-existent files to be checked on every call.

- As we've dropped support for earlier versions of Node, replace all calls to process.nextTick() with setImmediate() and expose setImmediate() in the plugin sandbox.

- Fix toobusy plugin, it was completely broken and not working correctly.

- Set backlog to zero on all listen() calls as it was found to improve connection handling under heavy loads (>30 connections/sec).

### DIFF
--- a/configfile.js
+++ b/configfile.js
@@ -169,7 +169,7 @@ cfreader.read_config = function(name, type, cb, options) {
     // Check cache first
     if (!process.env.WITHOUT_CONFIG_CACHE) {
         var cache_key = cfreader.get_cache_key(name, options);
-        if (cfreader._config_cache[cache_key]) {
+        if (cfreader._config_cache[cache_key] !== undefined) {
             //logger.logdebug('Returning cached file: ' + name);
             var cached = cfreader._config_cache[cache_key];
             // Make sure that any .ini file booleans are applied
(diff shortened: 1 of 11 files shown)


baudehlo:
Thinking about this Listen backlog - should we make it configurable?

Nginx has this to say:

```
backlog=number
sets the backlog parameter in the listen() call that limits the maximum length for the queue of pending connections. By default, backlog is set to -1 on FreeBSD, DragonFly BSD, and Mac OS X, and to 511 on other platforms.
```

And in the performance docs under the net.core.somaxconn settings: "Note: If you set this to a value greater than 512, change the backlog parameter to the NGINX listen directive to match"

msimerson:
I'd say that the first person to demonstrate the value of having it be configurable in Haraka can provide the impetus to make it so. Until then, as far as we know 0 is better than the default, right?

smfreegard:
For SMTP - I can't really seem to find any benefit of setting it to anything other than 0.   The problem being is that if you allow the kernel to queue up 511 connections, they all come thundering in at once.

I did consider making it configurable, but I really couldn't find a good argument for doing so.

baudehlo:
Did you test values other than 0?

PR #1740, opened 2016-12-12
set tls.sessionIdContext property

### LINKED ISSUE #1739: TLS reconnections don't work with Thunderbird
### system info

Haraka | Haraka.js — Version: 2.8.11
 --- | :--- 
Node | v6.9.1
OS | FreeBSD haraka 10.3-RELEASE-p11 FreeBSD 10.3-RELEASE-p11 #0: Mon Oct 24 18:49:24 UTC 2016     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64
### Expected behavior

1. Use the Thunderbird MUA.
2. Send a message using Haraka as the MSA (port 587 + TLS + AUTH).
3. Send another and have it got out.

### Observed behavior

1. Use the Thunderbird MUA.
2. Send a message using Haraka as the MSA (port 587 + TLS + AUTH).
3. Send another and have it fail.

Log message:
````
[INFO] [core] client [50.170.76.253] connection error: Error: 34439455744:error:140D9115:SSL routines:ssl_get_prev_session:session id context uninitialized:ssl_sess.c:682:
````

msimerson:
Fixes #1739 

- create a top level scoped secureContext (maybe store it in Redis eventually)
- consolidate 2x secureContext creations into `_getSecureContext()`
- set the context's sessionIdContext (this is the FIX)
- remove options.secureProtocol default 'SSLv23_method' (that's the node.js default since iojs)
- remove `secureOptions | constants.SSL_OP_NO_SSLv2 | constants.SSL_OP_NO_SSLv3`
    - SSL v2 and v3 were removed from node.js in since io.js
- remove crypto.createCredentials (deprecated)

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/tls_socket.js
+++ b/tls_socket.js
@@ -4,13 +4,13 @@
 /*----------------------------------------------------------------------------------------------*/
 
 var tls       = require('tls');
-var constants = require('constants');
-var crypto    = require('crypto');
 var util      = require('util');
 var net       = require('net');
 var stream    = require('stream');
 var log       = require('./logger');
 var EventEmitter = require('events');
+
+var secureContext;
 
 var ocsp;
 try {
@@ -208,6 +208,28 @@ if (ocsp) {
 
 exports.ocsp = ocsp;
 
+function _getSecureContext(options) {
+    if (secureContext) return secureContext;
+
+    if (options === undefined) options = {};
+
+    if (options.requestCert === undefined) {
+        options.requestCert = true;
+    }
+    if (options.rejectUnauthorized === undefined) {
+        options.rejectUnauthorized = false;
+    }
+    if (!options.sessionIdContext) {
+       	options.sessionIdContext = 'haraka';
+    };
+    if (!options.sessionTimeout) {
+       	// options.sessionTimeout = 1;
+    };
+
+    secureContext = tls.createSecureContext(options);
+    return secureContext;
+}
+
 function createServer(cb) {
     var serv = net.createServer(function (cryptoSocket) {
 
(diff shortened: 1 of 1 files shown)


smfreegard:
Am I right in thinking that this will not fix #1739 if running with nodes=cpus because if the reconnection happens to a different child (which is likely), then the secureContext will be different?

msimerson:
> Am I right in thinking that this will not fix #1739 if running with nodes=cpus because if the reconnection happens to a different child (which is likely), then the secureContext will be different?

Close, but not quite. Ya see, the root of the problem is `session id context uninitialized`, which this PR fixes entirely merely by setting the `id` property. Seriously.

You are very right to be thinking that the secureContext would best be shared among processes, but if not, or if a client presents a ticket that's not found in the secureContext, that's no big deal as the server will check the context and if missing, create a new ticket. It works almost exactly the same way if you set the sessionTimeout to something really low like 1 second. The TLS ticket ends up being expired after every connection.

msimerson:
One of the best explanations is on the OpenSSL users list:
- [session-id-context-uninitialized](http://openssl.6102.n7.nabble.com/error-140D9115-SSL-routines-SSL-GET-PREV-SESSION-session-id-context-uninitialized-td5749.html)
- [SSL_set_session_id_context man page](https://www.openssl.org/docs/man1.0.2/ssl/SSL_set_session_id_context.html)

msimerson:
Late at night, after all them Thunderbird users are gone to sleep, I snuck in and changed to nodes=4 (having 16 children is just kinda silly) and tested again, just to make sure. I sent a half dozen messages through unhindered. Previously (w/o patch) I could send the first and subsequent attempts (until restarting Thunderbird) would fail.

PR #1745, opened 2016-12-22
Added option to bypass SpamAssassin headers' merge

mnsilva:
- possibility not to merge SpamAssassin headers' into the message

Checklist:
- [x] docs updated
- [ ] tests updated

### DIFF
--- a/plugins/spamassassin.js
+++ b/plugins/spamassassin.js
@@ -24,11 +24,10 @@ exports.load_spamassassin_ini = function () {
         max_size:     500000,
         old_headers_action: "rename",
         subject_prefix: "*** SPAM ***",
-        add_headers: true,
     };
 
     for (var key in defaults) {
-        if (plugin.cfg.main[key] !== undefined) continue;
+        if (plugin.cfg.main[key]) continue;
         plugin.cfg.main[key] = defaults[key];
     }
 
@@ -188,10 +187,9 @@ exports.do_header_updates = function (connection, spamd_response) {
     }
 
     var modern = plugin.cfg.main.modern_status_syntax;
-    var add_headers = plugin.cfg.main.add_headers;
+    if ( !plugin.cfg.main.add_headers ) return;
 
     for (var key in spamd_response.headers) {
-        if ( !add_headers ) break;
         if (!key || key === '' || key === undefined) continue;
         var val = spamd_response.headers[key];
         if (val === undefined) { val = ''; }
--- a/plugins/spamassassin.js
+++ b/plugins/spamassassin.js
@@ -11,7 +11,11 @@ exports.register = function () {
 
 exports.load_spamassassin_ini = function () {
     var plugin = this;
-    plugin.cfg = plugin.config.get('spamassassin.ini', function () {
+    plugin.cfg = plugin.config.get('spamassassin.ini', {
+        booleans: [
+            '+add_headers',
+        ],
+    }, function () {
         plugin.load_spamassassin_ini();
     });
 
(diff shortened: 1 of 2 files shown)


msimerson:
There's already `old_headers_action=keep`, which is half-way to what you want. Rather than introducing another option that conflicts with `old_headers_action`, how about instead adding an option that makes the plugin more flexible with a setting such as, `add_headers=[true|false]`?

mnsilva:
Hello!
Yes, I checked the "old_headers_action" but, as you say, it keeps half-way and it is more "resource intensive" once both methods are still called even when the expected behaviour is not to add any additional headers to the message being processed.
What you purpose is a name-changing? If it is, the default value should be "true" and configuration will allow to change it to "false", right? It seems ok to me, I can change it that way (and I like "add_headers" most!). :)
Regards,Manuel
 
 
  Em Qui, 22 Dez, 2016 às 21:16, Matt Simerson<notifications@github.com> escreveu:   
There's already old_headers_action=keep, which is half-way to what you want. Rather than introducing another option that conflicts with old_headers_action, how about instead adding an option that makes the plugin more flexible with a setting such as, add_headers=[true|false]?

msimerson:
> both methods are still called even when the expected behaviour is not to add any additional headers

Yes, this is true. However, that's what we tend to consider a "premature optimization." Decades of experience have proven time and again that it's almost always better to optimize the code for legibility. I find that putting the config evaluation nearest the code that uses it tends to lend itself to obvious names for the config options and minimal side effects.

plugins/spamassassin.js
```diff
@@ -20,11 +24,11 @@ exports.load_spamassassin_ini = function () {
         max_size:     500000,
         old_headers_action: "rename",
         subject_prefix: "*** SPAM ***",
-        bypass_extra_headers: false,
+        add_headers: true,
     };
```
msimerson: you can remove add_headers from there, as declaring it as a boolean will insure it's initialized.

plugins/spamassassin.js
```diff
@@ -20,11 +24,11 @@ exports.load_spamassassin_ini = function () {
...
 
     for (var key in defaults) {
-        if (plugin.cfg.main[key]) continue;
+        if (plugin.cfg.main[key] !== undefined) continue;
         plugin.cfg.main[key] = defaults[key];
```
msimerson: this is an error and will prevent most of the defaults from being initialized. Please revert.

plugins/spamassassin.js
```diff
@@ -186,7 +188,10 @@ exports.do_header_updates = function (connection, spamd_response) {
     }
 
     var modern = plugin.cfg.main.modern_status_syntax;
+    var add_headers = plugin.cfg.main.add_headers;
+
```
msimerson: rather than uselessly iterating through the spamd_response headers, might it be better to return here on line 191 with something like: `if (!plugin.cfg.main.add_headers) return;`?

plugins/spamassassin.js
mnsilva: works for me... I first used an array similar to the one which casts numbers from string to Number but then I changed and forgot to remove the default value.

plugins/spamassassin.js
mnsilva: it makes sense once there's no "add_headers" key in "defaults".

the original code fails if the key is defined under defaults and configuration explicitly sets it to false... "defaults" value will always prevail. This change was to prevent this behaviour.

plugins/spamassassin.js
mnsilva: sounds good... I think I switched from premature optimization to no optimization. :D

msimerson:
Looks good, thanks!

PR #1763, opened 2017-01-12
Skip SPF check for outbound mail when context is not 'myself'

### LINKED ISSUE #1762: Skip SPF check for outbound mail when context is not 'myself'
The SPF plugin attaches SPF headers for outbound mail when context='sender'. For this, it uses `connection.remote.ip`. This is bound to fail when the remote IP is internal. In general, I am not aware why it is useful to attach SPF headers for outbound mail when context is not 'myself'.

gramakri:
Fixes #1762

### DIFF
--- a/plugins/spf.js
+++ b/plugins/spf.js
@@ -114,9 +114,10 @@ exports.hook_helo = exports.hook_ehlo = function (next, connection, helo) {
 exports.hook_mail = function (next, connection, params) {
     var plugin = this;
 
-    // For inbound message from a private IP, skip MAIL FROM check
-    if (!connection.relaying && connection.remote.is_private) {
-        return next();
+    // For messages from private IP space...
+    if (connection.remote.is_private) {
+        if (!connection.relaying) return next();
+        if (connection.relaying && plugin.cfg.relay.context !== 'myself') return next();
     }
 
     var txn = connection.transaction;


msimerson:
The edge case I see with this change affects orgs that have to send all their emails via a 3rd party (typically their ISP) mail server. Think of a small local business with an internal mail server who has outbound port 25 blocked by their ISP (this is not unusual). In that case, they set up their MTA to relay through their ISP. Their SPF records cover their and their ISPs mail server and everyone is happy. I can't see any reason not to evaluate SPF in that case. 

If you were to amend this PR such that it disabled the same check but only for private IP space (see attached diff), then I'd +1 that.

```diff
diff --git a/plugins/spf.js b/plugins/spf.js
index 0e880e7..0e55915 100644
--- a/plugins/spf.js
+++ b/plugins/spf.js
@@ -114,9 +114,10 @@ exports.hook_helo = exports.hook_ehlo = function (next, connection, helo) {
 exports.hook_mail = function (next, connection, params) {
     var plugin = this;
 
-    // For inbound message from a private IP, skip MAIL FROM check
-    if (!connection.relaying && connection.remote.is_private) {
-        return next();
+    // For messages from private IP space...
+    if (connection.remote.is_private) {
+        if (!connection.relaying) return next();
+        if (connection.relaying && plugin.cfg.relay.context !== 'myself') return next();
     }
 
     var txn = connection.transaction;
```

gramakri:
@msimerson Thanks for the review. I have modified the patch accordingly.

msimerson:
+1

PR #1766, opened 2017-01-16
redis: add redis_subscribe_pattern()

msimerson:
redis: add redis_subscribe_pattern() to enable watch to subscribe once for every connection vs subscribing anew for every connection


msimerson:
side topic: prhero is a bot that's putting a ❤️ on every PR everywhere on GitHub. I just blacklisted it.

PR #1773, opened 2017-01-19
Queue filename refactor

EyePulp:
- Standardize on the underscore for delimiting all queue filename bits
- Return a simple object for the parts
- Add the original arrival timestamp to the filename
- Calculate message "age" value when getting file parts
- Allow for [re]names with defaults & overrides
- Convert older filename formats in the queue during an update

### DIFF
--- a/outbound.js
+++ b/outbound.js
@@ -273,9 +273,10 @@ exports.load_queue_files = function (pid, cb_name, files, callback) {
                 var next_process = parts.next_attempt;
                 // maintain some original details for the rename
                 var new_filename = _qfile.name({
-                    arrival:parts.arrival,
-                    next_attempt:parts.next_attempt,
-                    attempts:parts.attempts
+                    arrival      : parts.arrival,
+                    uid          : parts.uid,
+                    next_attempt : parts.next_attempt,
+                    attempts     : parts.attempts,
                 });
                 // self.loginfo("new_filename: ", new_filename);
                 fs.rename(path.join(queue_dir, file), path.join(queue_dir, new_filename), function (err) {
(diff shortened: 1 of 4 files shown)


msimerson:
LGTM

EyePulp:
Sorry - found a file rename occurring that wasn't covered in the original submission.  Updated and corrected

outbound.js
```diff
@@ -421,18 +423,78 @@ exports.stats = function () {
...
 };
 
-function _next_uniq () {
-    var result = uniq++;
-    if (uniq >= MAX_UNIQ) {
-        uniq = 1;
+
+var _qfile = exports.qfile = {
+    // File Name Format: $arrival_$nextattempt_$attempts_$pid_$uniquenum_$host
+    name : function(overrides){
```
baudehlo: Spacing: `function (params...) {`

Same for the other functions.

outbound.js
```diff
@@ -421,18 +423,78 @@ exports.stats = function () {
     return results;
 };
 
-function _next_uniq () {
```
baudehlo: This function ensures that we don't overwrite the same file if the same thing happens on the same PID within one millisecond. You now have nothing to prevent that.

baudehlo:
Looking at it a bit now. Couple of issues so far.

outbound.js
EyePulp: I see what you're saying.  I figured the likelihood of getting two identical occurrences of 10 character alphanumeric random strings with an identical arrival moment and an identical next_attempt value would be vanishingly small (and it provides a visual cue for continuity if the same file sticks around for a while).

outbound.js
EyePulp: I'll get these corrected

plugins/dnswl.js
```diff
@@ -3,8 +3,8 @@
 
 exports.register = function() {
     var plugin = this;
-    plugin.load_dnswl_ini();
     plugin.inherits('dns_list_base');
+    plugin.load_dnswl_ini();
```
msimerson: Looks like a separate issue from the queue filename. If this were in another PR and a little bit of explanation, it would likely get merged straight away.

EyePulp: It should be separate - I'm just dumb when it comes to git/github.

I meant to make it separate, but failed, and now I'm not sure how to fix it. =\

msimerson:
````
git checkout -b another-branch
git remote update haraka (make sure your local copy of haraka/master is up-to-date)
git rebase -i haraka/master    (delete all but the last commit)
git push
````

Then you'll have a new branch in your repo that you can create a PR from.

PR #1776, opened 2017-01-24
typo: logerr -> logerror

gramakri:
Should fix the stack trace below:

[CRIT] [3893B21C-D882-4263-BA38-A3C9378F9A92] [core] Plugin tls failed: TypeError: log.logerr is not a function
    at pluggableStream.socket.upgrade (/app/haraka/tls_socket.js:260:29)


msimerson:
thanks @gramakri

PR #1777, opened 2017-01-25
Redis consistency

msimerson:
# Background

The redis plugin has three personalities:
1. add `redis` to config/plugins and at startup, the redis plugin connects to the redis server and leaves a shared db handle at `server.notes.redis`. This shared redis handle is used by the results store, the tls plugin, and any plugins that `inherit('redis')` without specifying a DB ID in their config.
2. As a base class for other plugins to inherit. This is how it has been used by rcpt_to.routes and the karma plugin. The redis plugin exports `merge_redis_ini()` and `init_redis_plugin()` which are used solely by plugins to populate their redis connection settings and establish the redis connection when Haraka starts up.
3. As a pub/sub agent helper. The result_store publishes all results as plugins emit them. Other plugins can subscribe to those redis channels. Both `karma` and `watch` plugins use redis subscriptions.

# Changes in this pull request:

- redis: 
    - only establish the shared server.notes.redis when plugin.name=redis
        - avoids extra establishment attempts when other plugins do: plugin.inherits("redis");
    - add merge_redis_ini() for plugins to call. Provides defaults for host, port, and db, when they're not specifically defined in the inheriting plugins [redis] config.
    - add test coverage for merge_redis_ini
    - add test coverage for redis_init_plugin with a DB ID specified
    - add inheritance usage example to docs
    - renamed `init_redis_connection()` -> `init_redis_shared()`  (better contrast with `init_redis_plugin`)
- rcpt_to.routes: use merge_redis_ini()
- tls: make redis comment more specific
- greylist: inherit(redis), prune redundant redis setup code
- rate_limit: inherit(redis), prune redundant redis setup code
    - rename self -> plugin (consistency)
    - use plugin.cfg w/cb for rate_limit.ini  (vs lots of var config = plugin.config.get(...))
    - update redis config format so a DB ID can be specified
        - with a backwards compatible config shim

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/config/rate_limit.ini
+++ b/config/rate_limit.ini
@@ -1,7 +1,11 @@
 ; Example configuration file for the rate_limit plugin
 
-; redis_server = 1.2.3.4
 ; tarpit_delay = 30
+
+[redis]
+; host = 1.2.3.4
+; port = 6379
+; db = 0
 
 [concurrency]
 ; NOTE: this limit is per server child and does not use Redis
--- a/config/tls.ini
+++ b/config/tls.ini
@@ -16,7 +16,7 @@ ciphers=ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-G
 
 
 [redis]
-; options in this block require redis to be available.
+; options in this block require redis to be enabled in config/plugins.
 
 ; remember when a remote fails STARTTLS. The next time they connect,
 ;     don't offer STARTTLS option (so message gets delivered).
--- a/docs/plugins/rate_limit.md
+++ b/docs/plugins/rate_limit.md
@@ -3,62 +3,55 @@ rate\_limit
 
 Enforce limits on connection concurrency, connection rate, and recipient rate.
 
-By default DENYSOFT will be returned when the limits are exceeded, but for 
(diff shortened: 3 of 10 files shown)


bmatson:
Matt,

While you’re in there messing with the rate_limit plugin you should probably address the race condition I identified in bug 1757.  Your new version still has the same problem, in the specific case where Redis decides your record has timed out between the time you initially read it and get a non-zero value and the time you subsequently call incr (which silently re-creates the record for you, but without a TTL value).  When this happens the record doesn’t time out and you end up rate limiting when you shouldn’t because there is no TTL to ever expire.

The fix I’m proposing checks the return value from the incr, and in the event it is 1 (eg. Redis just created the record, we explicitly set the TTL following the incr).  I’ve been running a local copy of the plugin and haven’t seen the problem recur since implementing this fix…..

Bill

msimerson:
Hey @bmatson , I'll have a look. Any chance you can put your code suggestion in PR? 

Also, we like to keep PRs on the smaller side, so it's better to have two PRs (this one, which is is consistency and DRY (don't repeat yourself) focused, and if I've done my job well, has no side effects, and then another PR that has the intended effect of fixing the rate_limit issue.

PR #1785, opened 2017-01-28
the one limit plugin to rule them all

msimerson:
- merges these 3 plugins into [haraka-plugin-limit](https://github.com/haraka/haraka-plugin-limit):
    - plugins/limit
    - plugins/rate_limit
    - [haraka-plugin-outbound-rate-limit](https://github.com/haraka/haraka-plugin-outbound-rate-limit)
- limit: redis only now
    - drops haraka-nosql dependency (RAM or SSC storage)
    - consequence: a local redis needs to be running to pass **all** the tests
- limit: concurrency counters set a 2 minute lifetime (nosql module did that in a brutish way, redis key expiration is much cleaner)
- inherits [haraka-plugin-redis](https://github.com/haraka/haraka-plugin-redis)  (can be fully tested outside of Harka)
- abstracts the tarpit delays into a function, shared by limit & rate limit functions
    - no longer depends on tarpit plugin
- rate_limit:
    - drops the connection concurrency limiting:
        - limit plugin already had it
        - it was RAM only, with a "TODO in Redis" comment
    - refactored into smaller simpler functions
    - added test coverage for them functions (0% -> ~80% test coverage)
    - instead of chained caller, each type of limit has registered hooks
- outbound-rate-limit
    - merged in the increment and decrement functions
    - didn't need the rest, as limit already inherits from the redis plugin
    - instead of doing a hincrby & then a get (race prone, and slower), just add the callback to `hincrby` like so:
        `plugin.db.hincrby(getOutKey(hmail), 'TOTAL', 1, function (err, count) {...}`
    - sets a 5 minute expiration on entries: we have plenty of experience with rate limits, enough to know that after some time the counters get wonky (crashes, unclean shutdowns, etc.). So expire them after some reasonable amount of time and they're self healing.

Checklist:
- [x] docs updated
- [x] tests updated


msimerson:
Closed #754 

It doesn't specifically implement the bounce and auth limits, but with the history extensions added to the `rate_limit` features of this new plugin, senders that fail auth get terrible karma and thus severe connection restrictions, accomplishing most of what I wanted.

msimerson:
## more changes

I changed the algorithm for rate limiting. The previous implementation was naive, using an  incrementing counter and a key expiration. Key expiration was inconsistent because it was based on when the first connection arrived (and expiration was set) and was thus disconnected from when in that window the connections arrived. If most arrived in the latter portion of the window, connection limits would regularly be surpassed by up to 2x.

Instead, maintain a hash with keys indexed by timestamp (ms) and a counter as the value. Even if two connections arrive at the same instant, they'll get counted correctly. When enforcing, retrieve all the keys, add up the counters where the timestamps are within the TTL window, and thus exact limits are always enforced. The record expiration is 2*TTL. 

A side benefit is that if the config is changed, The Right Limits will automatically happen on all new connections. Immediately.

## other changes

* refactored get_history out of get_limits, so it can be reused in rate limits
* added history limit support to rate_conn (so bad connections get severe limits and good ones automatically get lax limits)
* renamed history sections
* refactored mostly duplicated get_recipient_limit & get_connection_limits into get_limits
* split connection rate limiting into two functions, one that increments (on connect_init) and one that enforces (on connect).

PR #1786, opened 2017-01-28
Redis npm

msimerson:
- replace plugins/redis with [haraka-plugin-redis](https://github.com/haraka/haraka-plugin-redis)
    - haraka-plugin-redis must exist for npm packaged plugins to test outside of Haraka (karma, known-senders, limit, watch)
    - additional testing was added in the new version
- while we should be able to drop the redis dependency (since haraka-plugin-redis has it as a dep), we can't do that yet because of the `vm` tomfoolery.

# asides

- this removes the redis plugin tests from this repo
- if the last couple redis-consuming plugins get repackaged to npm (along with *their* tests), then the haraka test suite can once again pass without a redis server available.

Checklist:
- [x] docs updated
- [x] tests updated


baudehlo:
while I should be able to drop the
So much suspense and mystery in this PR.

msimerson:
> So much suspense and mystery in this PR.

Did the message you received get truncated?  Maybe visit the PR for the rest of the description?

baudehlo:
Yeah weirdly this is what Apple mail on iOS shows:

bmatson:
Looks like you missed the required changes in the rate_limit.js plugin...the version currently in master looks like it's broke based on this change since it's still using the redis plugin.

msimerson:
Aye, 'tis true. Since you're running out of master, might you care to try the [new limit plugin](https://github.com/haraka/haraka-plugin-limit) (see also #1785) with rate limit support?

bmatson:
Unfortunately I'm not actually running out of master, just the last release with a slightly hacked up local copy of rate_limit.js to get around the race condition I wrote up.  I did give the PR limit plugin a code review and I see where it does not seem susceptible to the same type of race conditions with redis.....  I'm keeping track of the changes just to make sure the "next" release has functional rate limiting, whether from the "old" rate_limit.js or your new limit plugin doesn't make much difference to me.

msimerson:
In case you don't already know, the right way to use your local hacked rate_limit is by copying it to /path/to/haraka/config/plugins, and Haraka will prefer that to the installed version in the Haraka install dir. 

Same goes for npm packaged plugins. If you:

```sh
cd /path/to/haraka
npm install haraka-plugin-limit
```

and then add `limit` to your config/plugins file, the 'local' one in your config directory takes precedence.

Issue #1768, opened 2017-01-18
SSL Connection error after PR 1740


gramakri:
I am running latest haraka (commit  f6013b9) with node 4.2.2 (I realize haraka requires 4.6.2 atleast but I am logging this issue just in case).

After https://github.com/haraka/Haraka/pull/1740/commits/e407e2f8d08505bc743716d9d13129f0bef78ad8, the server sometimes does not accept TLS connections. I have verified that reverting that change makes things consistently work.

TLS upgrade fails with the error:
```
139672910050968:error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure:s23_clnt.c:769:
```

The curious part here is that _restarting_ haraka makes the TLS connection go through fine mostly. So, maybe it is some race during setup?

gramakri:
Client debug output:
```
ubuntu@ip-10-0-0-170:~$ openssl s_client -starttls smtp -connect <haraka>:25
CONNECTED(00000003)
140015963801240:error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure:s23_clnt.c:769:
---
no peer certificate available
---
No client certificate CA names sent
---
SSL handshake has read 238 bytes and written 340 bytes
---
New, (NONE), Cipher is (NONE)
Secure Renegotiation IS NOT supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : 0000
    Session-ID: 
    Session-ID-ctx: 
    Master-Key: 
    Key-Arg   : None
    PSK identity: None
    PSK identity hint: None
    SRP username: None
    Start Time: 1484696526
    Timeout   : 300 (sec)
    Verify return code: 0 (ok)
---
```

On the haraka side, I see:
```
[core] client [<ip>] dropped connection
```

msimerson:
`sslv3 alert handshake failure`

That's expected. That PR dropped some of our "hacks to disable SSLv3" code because node.js has already done that for us in supported versions. Until Node.js gets out of the habit of shipping a bundled OpenSSL, you really do need to be running a recent node. Really really.

gramakri:
@msimerson I understand that the code to disable SSLv3 was dropped. I don't understand why haraka starts accepting connections after a restart (with node 4.2.2). Do you have any theories?

I am testing with 6.9.4 right now.

gramakri:
[...]

PR #1787, opened 2017-01-28
Do not create a top level secureContext

gramakri:
This causes problems because the secureContext ends up being used by
Haraka in server and client mode. Currently, the server mode loads
TLS keys as part of the context in (plugins/tls_socket). The client
mode does not load keys (in smtp_client.js). This discrepancy causes
tls connections to fail when there is an incoming mail immediately
followed by an outgoing mail.

### DIFF
--- a/tls_socket.js
+++ b/tls_socket.js
@@ -10,8 +10,6 @@ var stream    = require('stream');
 var log       = require('./logger');
 var EventEmitter = require('events');
 
-var secureContext;
-
 var ocsp;
 try {
     ocsp      = require('ocsp');
@@ -209,8 +207,6 @@ if (ocsp) {
 exports.ocsp = ocsp;
 
 function _getSecureContext (options) {
-    if (secureContext) return secureContext;
-
     if (options === undefined) options = {};
 
     if (options.requestCert === undefined) {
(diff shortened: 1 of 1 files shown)


gramakri:
This was the root cause for https://github.com/haraka/Haraka/issues/1768. To reproduce this manually:
1. send mail from haraka to say server S1. the SMTPClient creates a TLS context without "key".
2. send mail to haraka from server S1. the tls connection will fail. the server re-uses the above context (it requires "key").

msimerson:
I think this is fine, but I'd like to test this before merging. I'll try to get that done tomorrow.

tls_socket.js
```diff
@@ -221,12 +217,12 @@ function _getSecureContext (options) {
...
-    };
+    }
     if (!options.sessionTimeout) {
        	// options.sessionTimeout = 1;
-    };
+    }
 
-    secureContext = tls.createSecureContext(options);
+    var secureContext = tls.createSecureContext(options);
     return secureContext;
```
msimerson: Might as well remove the variable declaration entirely and just `return tls.createSecureContext(options);`

msimerson:
I've made the change in the PR. Do you agree @gramakri?

gramakri:
@msimerson yes, OK with me, thanks!

PR #1792, opened 2017-01-31
Create an outbound queue filename handler

EyePulp:
- Standardize on the underscore for delimiting all queue filename bits
- Return a simple object for the parts
- Add the original arrival timestamp to the filename
- Calculate message "age" value when getting file parts
- Allow for [re]names with defaults & overrides
- Convert older filename formats in the queue during an update


msimerson:
Ahhh, that was me using GitHub's fancy "fix merge conflicts tool," but apparently I didn't do it very well.

PR #1793, opened 2017-01-31
swap lines to fix missing inherited methods

EyePulp:
Fix a transposed pair of lines in `plugins/dnswl.js`

```javascript
    plugin.load_dnswl_ini();
    plugin.inherits('dns_list_base');
```

`plugin.load_dnswl_ini()` attempts to use inherited methods, which aren't yet present.  A line swap fixes it.

### DIFF
--- a/plugins/dnswl.js
+++ b/plugins/dnswl.js
@@ -1,10 +1,10 @@
 'use strict';
 // dnswl plugin
 
-exports.register = function() {
+exports.register = function () {
     var plugin = this;
-    plugin.load_dnswl_ini();
     plugin.inherits('dns_list_base');
+    plugin.load_dnswl_ini();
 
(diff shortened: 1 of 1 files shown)


msimerson:
Mmmm, I see the problem. load_dnswl_ini() isn't an inherited function, it's actually in dnswl.js, but I'll merge this (b/c I think inheritance should come first) and fix the problem you're seeing in another PR.

EyePulp:
I could have phrased it better - there's a conditional call inside `load_dnswl_ini()` that does call on the inherited `check_zones()` function here:
https://github.com/haraka/Haraka/blob/master/plugins/dnswl.js#L39-L41

msimerson:
Ahhhh, right you are. Good catch.

PR #1796, opened 2017-02-01
comment out working test sample auth user

msimerson:
# Fixes

With the new config loading feature, whereby local config files are merged onto the included config files, having valid sample credentials in that config file has the potential to turn Haraka instances into open relays when the admin enables the `auth_flat_file` plugin.

- comment out the example entry


baudehlo:
Makes sense to me.

PR #1805, opened 2017-02-05
add an additional plugin search path

EyePulp:
This is to account for the module-based plugin locations when installing via NPM.   They live at the  same level as the `Haraka/` directory (which is to say **outside** of it), but currently there is no lookup path in `plugin_search_paths()` that hits them.  I think this issue isn't visible when installing from git, as `npm install` is run **inside** the Haraka directory to get all the necessary modules & plugins, which puts them in the `Haraka/node_modules/` folder and matches expectations for the current list of searched paths.  

Installing via NPM puts everything in this structure:
```bash
/some/path/$ npm install Haraka
# creates the following director structure

/some/path/node_modules/
/some/path/node_modules/Haraka
/some/path/node_modules/haraka-plugin-syslog
/some/path/node_modules/haraka-plugin-geoip
...etc
```

- Add an additional lookup path for the `haraka-plugin-` prefixed modules


msimerson:
Hey @EyePulp , what we already do have support for is searching for haraka-plugin-`name` in Haraka (where any of Haraka's dependency npm modules will be) as well as in the haraka config directory (/etc/haraka, etc.). So if you:

````
cd /path/to/haraka/config
cd ..
npm install haraka-plugin-`name`
````

Then your plugins will get loaded as you expect and you don't have to install with `npm install -g`.

EyePulp:
Hey @msimerson , thanks for the note and taking a look on a late Saturday night =P

I'm not sure why the `haraka-plugin-` modules are getting installed at all if they're not intended to be discovered by default.  `npm install Haraka` dumps the `haraka-plugin-` modules in a location that isn't on any of the current plugin search paths.  If a manual install of them is needed, there may be more documentation required.  For instance the `haraka-plugin-syslog` module docs just say to add syslog to my plugins file.

I'm not married to this fix, but it does feel like it eliminates some confusion in an area that's inconsistent between NPM and git checkouts.

msimerson:
Now I see the issue.

It looks like (from local testing) that NPM 4.x still installs Haraka's modules in (my case) `../Haraka/node_modules` **when** installing with the -g (global) flag. However, when installing into a local path (not global) as you've shown above, npm@4 flattens the dependencies, so they're all installed in the top level node_modules directory. This is a change in npm behavior. (npm@2 for sure and I think v3 also installed a modules deps within that modules node_modules dir.)

````
# mkdir foo
# cd foo
# npm install Haraka

`-- Haraka@2.8.13 
  +-- address-rfc2821@1.0.1 
  +-- address-rfc2822@1.0.1 
  +-- async@2.1.4 
  | `-- lodash@4.17.4 
  +-- daemon@1.1.0 
  +-- elasticsearch@12.1.3 
  | +-- chalk@1.1.3 
  | | +-- ansi-styles@2.2.1 
 <snip>
  | `-- os-tmpdir@1.0.2 
  `-- vs-stun@0.0.7 

# ls node_modules/
.bin			dashdash		inflight		osenv
Haraka			double-ended-queue	inherits		path-is-absolute
abbrev			dtrace-provider		ipaddr.js		precond
address-rfc2821		elasticsearch		js-yaml			promise
address-rfc2822		escape-string-regexp	ldap-filter		pseudomap
ansi-regex		esprima			ldapjs			punycode
ansi-styles		extsprintf		lodash			redis
argparse		forever-agent		lru-cache		redis-commands
asap			generic-pool		maxmind			redis-parser
asn1			glob			minimalistic-assert	rimraf
asn1.js			haraka-config		minimatch		safe-json-stringify
asn1.js-rfc2560		haraka-constants	minimist		semver
asn1.js-rfc5280		haraka-net-utils	mkdirp			simple-lru-cache
assert-plus		haraka-plugin-asn	modern-syslog		sprintf-js
async			haraka-plugin-geoip	moment			strip-ansi
backoff			haraka-plugin-karma	mv			supports-color
balanced-match		haraka-plugin-limit	nan			tmp
big-integer		haraka-plugin-redis	ncp			ultron
bn.js			haraka-plugin-syslog	nopt			vasync
brace-expansion		haraka-plugin-watch	npid			verror
bunyan			haraka-results		ocsp			vs-stun
chalk			haraka-tld		once			wrappy
concat-map		haraka-utils		options			ws
core-util-is		has-ansi		os-homedir		yallist
daemon			iconv			os-tmpdir
````

I don't see any problems with this patch. It's probably the right thing to do in light of how npm installs dependencies now. I'm going to give @haraka/core a chance to pipe in before merging. There might be a better way.

EyePulp:
Awesome - I totally forgot npm changed to that flatter structure, probably because it's papered over by the built-in `require()` module just handling it.  We don't have that luxury when rolling our own discovery mechanism unfortunately.  Thanks npm. =)

PR #1808, opened 2017-02-06
This may or may not fix issues with short attachments.

baudehlo:
This is basically some belt and braces around short attachments sent in one TCP packet.

It has been running on emailitin.com for a month or so, with no adverse effects, I just don't have feedback on whether it fixed the issue or not - I was unable to replicate it (it requires major full control of the sending SMTP client library at the TCP level), and the customer who had the problem left the service. But I think we should merge it anyway.

### DIFF
--- a/attachment_stream.js
+++ b/attachment_stream.js
@@ -29,13 +29,18 @@ AttachmentStream.prototype.emit_data = function (data) {
     }
 };
 
-AttachmentStream.prototype.emit_end = function () {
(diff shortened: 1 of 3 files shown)


msimerson:
Now I remember why I had disabled "Require branch to be up-to-date."  If there's several open PRs, and one goes along to merge them, as soon as you merge one, the others are all out of date. Annoying...

PR #1809, opened 2017-02-06
Cache the node_modules folder

baudehlo:
This just makes Travis faster. Newer modules update the cache so it's pretty safe.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,3 +34,8 @@ after_success:
     - ./node_modules/.bin/codecov
 
 sudo: false
+
+cache:
+    directories:
+        - node_modules
+


msimerson:
Did Travis-CI open up caching to open repos?  Last I knew:

> For all the gory details on caching, see our docs. Note that this feature is currently only available for private repositories.  -- [Travis Blog](https://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies/)

/me goes off to read up

msimerson:
> Newer modules update the cache so it's pretty safe.

How? When?

I'm thinking two things:
1. alarm bells go off every time I hear the word cache, and I remember that old axiom, *never cache anything you don't know how to invalidate*. In this case, I'm wondering how and when Travis will invalidate the cache. If an older version of a dep is installed, and a newer version is available on NPM, which version do we get? The answer seems to be [it depends](https://github.com/travis-ci/travis-ci/issues/3758), which leaves me a tad bit uneasy.
2. Caching our NPM deps could take something that's proactive and identifies build problems with new versions of NPM deps (Greenkeeper) and not just hobble it, but cause it to report new versions are fine, when in reality they aren't.
3. This is likely to shave about 30 seconds off our ~2.25m build times. That's not inconsequential, so I'm half temped to just say, "go for it."
4. NPM installs on Travis is usually really really fast, except when compilation is required. Something else that would shave that same 30 seconds off our build time and doesn't involve caching is unbundling the syslog plugin and iconv, so that we don't have to install all the APT packages required to compile them. Drop them into their own repos so they only have to build and run their tests when they change.

PR #1822, opened 2017-02-12
add logger methods on outbound

### LINKED ISSUE #1821: outbound logging crash
### system info

Haraka | Haraka.js — Version: 2.8.13
 --- | :--- 
Node | v7.3.0
OS | FreeBSD haraka 10.3-RELEASE-p7 FreeBSD 10.3-RELEASE-p7 #0: Thu Aug 11 18:38:15 UTC 2016     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64

### Expected behavior

Haraka to start normally.

### Observed behavior

````
Feb 12 06:14:15 haraka haraka[24061]: [NOTICE] [-] [core] Shutting down
Feb 12 06:14:15 haraka haraka[15450]: [NOTICE] [-] [core] worker 4 exited with error code: 1
Feb 12 06:14:15 haraka haraka[15450]: [NOTICE] [-] [core] worker 5 started pid=24063
Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core] TypeError: this.loginfo is not a function
Feb 12 06:14:16 jails kernel: Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core] TypeError: this.loginfo is not a function
Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at Object.exports.load_pid_queue (/usr/local/lib/node_modules/Haraka/outbound.js:217:10)
Feb 12 06:14:16 jails kernel: Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at Object.exports.load_pid_queue (/usr/local/lib/node_modules/Haraka/outbound.js:217:10)
Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at process.<anonymous> (/usr/local/lib/node_modules/Haraka/outbound.js:165:17)
Feb 12 06:14:16 jails kernel: Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at process.<anonymous> (/usr/local/lib/node_modules/Haraka/outbound.js:165:17)
Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at emitTwo (events.js:111:20)
Feb 12 06:14:16 jails kernel: Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at emitTwo (events.js:111:20)
Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at process.emit (events.js:191:7)
Feb 12 06:14:16 jails kernel: Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at process.emit (events.js:191:7)
                                                                                                        60669,1       39%
Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at process.emit (events.js:191:7)
Feb 12 06:14:16 jails kernel: Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at process.emit (events.js:191:7)
Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at process.nextTick (internal/child_process.js:752:12)
Feb 12 06:14:16 jails kernel: Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at process.nextTick (internal/child_process.js:752:12)
Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at _combinedTickCallback (internal/process/next_tick.js:67:7)
Feb 12 06:14:16 jails kernel: Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at _combinedTickCallback (internal/process/next_tick.js:67:7)
Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at process._tickCallback (internal/process/next_tick.js:98:9)
Feb 12 06:14:16 jails kernel: Feb 12 06:14:16 haraka haraka[24063]: [CRIT] [-] [core]     at process._tickCallback (internal/process/next_tick.js:98:9)
Feb 12 06:14:16 haraka haraka[24063]: [NOTICE] [-] [core] Shutting down
Feb 12 06:14:16 haraka haraka[15450]: [NOTICE] [-] [core] worker 5 exited with error code: 1
Feb 12 06:14:16 haraka haraka[15450]: [NOTICE] [-] [core] worker 6 started pid=24068
Feb 12 06:14:16 dovecot dovecot: auth: digest-md5(cynthia@simerson.net,72.168.145.16,<oLsk8lVITHZIqJEQ>): password mismatch
Feb 12 06:14:17 haraka haraka[24068]: [CRIT] [-] [core] TypeError: this.loginfo is not a function
Feb 12 06:14:17 jails kernel: Feb 12 06:14:17 haraka haraka[24068]: [CRIT] [-] [core] TypeError: this.loginfo is not a function
... ad infinitum ...
````

### Steps to reproduce

1. Install latest Haraka from master.
2. wait until she goes boom (took 10 hours for me)

### PR DESCRIPTION
**Pull request by msimerson**
Fixes #1821

Using `this.log*` in outbound is generally a good thing, as it automatically prefixes the log emissions with the `[outbound]` prefix. A problem came along in #1807 when 23 lines of code were astutely replaced with one:

```js
logger.add_log_methods(HMailItem.prototype, "outbound");
```

The problem was, the previous 23 lines of code added the logger functions to **both** HMailItem.prototype and exports, and the new line of code added them to only HMailItem.prototype. This I expect caused caused much crashing during local testing as `this.log*` methods blew up.

Rather than also adding the log methods to exports, this was instead done:

```diff
-    this.loginfo("Processing domain: " + todo.domain);
+    logger.loginfo("[outbound] Processing domain: " + todo.domain);
```

Was this a conscious choice, or a "I should have gone to bed an hour ago" choice? 

This PR assumes the latter, adds the logging methods back to exports, and removes the no-longer-necessary `[outbound]` prefixes from the log invocations.

### DIFF
--- a/outbound.js
+++ b/outbound.js
@@ -89,6 +89,8 @@ exports.load_config();
 exports.net_utils = net_utils;
 exports.config    = config;
 
+logger.add_log_methods(exports, "outbound");
+
 var load_queue = async.queue(function (file, cb) {
     var hmail = new HMailItem(file, path.join(queue_dir, file));
     exports._add_file(hmail);
@@ -184,33 +186,34 @@ process.on('message', function (msg) {
 
 exports.drain_pools = function () {
     if (!server.notes.pool || Object.keys(server.notes.pool).length == 0) {
-        return logger.logdebug("[outbound] Drain pools: No pools available");
+        return this.logdebug("Drain pools: No pools available");
     }
     for (var p in server.notes.pool) {
-        logger.logdebug("[outbound] Drain pools: Draining SMTP connection pool " + p);
+        this.logdebug("Drain pools: Draining SMTP connection pool " + p);
         server.notes.pool[p].drain(function () {
             if (!server.notes.pool[p]) return;
             server.notes.pool[p].destroyAllNow();
         });
     }
-    logger.logdebug("[outbound] Drain pools: Pools shut down");
+    this.logdebug("Drain pools: Pools shut down");
 }
 
 exports.flush_queue = function (domain, pid) {
-    if (domain) {
-        exports.list_queue(function (err, qlist) {
-            if (err) return logger.logerror("Failed to load queue: " + err);
-            qlist.forEach(function (todo) {
-                if (todo.domain.toLowerCase() != domain.toLowerCase()) return;
-                if (pid && todo.pid != pid) return;
-                // console.log("requeue: ", todo);
-                delivery_queue.push(new HMailItem(todo.file, todo.full_path));
-            });
-        })
-    }
-    else {
+    if (!domain) {
         temp_fail_queue.drain();
+        return;
     }
+
+    var self = this;
+    exports.list_queue(function (err, qlist) {
+        if (err) return self.logerror("Failed to load queue: " + err);
+        qlist.forEach(function (todo) {
+            if (todo.domain.toLowerCase() != domain.toLowerCase()) return;
+            if (pid && todo.pid != pid) return;
+            // console.log("requeue: ", todo);
+            delivery_queue.push(new HMailItem(todo.file, todo.full_path));
+        });
+    })
 };
 
 exports.load_pid_queue = function (pid) {
@@ -219,8 +222,7 @@ exports.load_pid_queue = function (pid) {
 };
 
 exports.ensure_queue_dir = function () {
-    // No reason not to do this stuff syncronously -
-    // this code is only run at start-up.
+    // No reason not to do this syncronously - only run at start-up.
     if (fs.existsSync(queue_dir)) return;
 
     this.logdebug("Creating queue directory " + queue_dir);
@@ -229,7 +231,7 @@ exports.ensure_queue_dir = function () {
     }
     catch (err) {
         if (err.code !== 'EEXIST') {
-            logger.logerror("Error creating queue directory: " + err);
+            this.logerror("Error creating queue directory: " + err);
             throw err;
         }
     }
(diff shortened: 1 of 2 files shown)


msimerson:
This PR also replaces a handful of `logger.log*` calls with `self.log*`, so those log messages get the automatic `[outbound]` prefix. I did that in all the functions that were exported, and thus had the logging methods available.

baudehlo:
It was actually a conscious choice. We don't add the logger methods to exports anywhere else and it seems like a bad hack.

msimerson:
> We don't add the logger methods to exports anywhere else

Except where we do. We have them on Server, connection, and plugin all come to mind. (On the latter two, they are on the prototype, but still, the convention of using `this.log*` within Server, connection, and a plugin all apply.

I don't disagree with it being a bad hack. What I do disagree with is having mixed syntax within the same file. Sometimes you have to use logger, sometimes you use this, other times you use self, etc..  That is distinctly unfriendly to humans to get right. 

I actually started out by replacing all the calls in outbound.js with `logger.*` calls. At least that way it would be consistent and not crash my mail server. But then we lose the `[outbound]` prefixes.

What I'd much rather see us adopt project-wide is to call a local logging function whose initial implementation is just a wrapper for console.log. Calls to the local logger just prefix any messages with a `this.name` property that we've defined. Once the module is loaded, an instance of haraka logger can be passed in that does all the extra stuff we like. That approach puts us onto a path where we can drop in other loggers (winston, bunyon, log4js) instead of our own ./logger.js.

baudehlo:
Well I support the notion of calling it on the HMailItem - and that the code for that should be in a separate file.

Fixing it to use logger for everything else should have been trivial. I suspect there might have been one place left to fix.

Issue #1831, opened 2017-02-17
Split up outbound.js


baudehlo:
We need to modularize outbound.js a bit more. There's several components it could be split into:

* a queue loader/processor
* a "main" library that does the main outbound bits (send_email etc)
* The outbound connection pool code
* the ToDo library
* the HMailItem library

PR #1836, opened 2017-02-22
Move outbound to its own sub-folder

baudehlo:
First stage in #1831 but doesn't resolve that issue yet


baudehlo:
Note please don't merge yet. More changes are pending.

PR #1839, opened 2017-02-25
Corrects an issue with the logger in configfile.js causing it to crash

wuvvy:
On Ubuntu 16.04 (at least, was not tested with anything else), Haraka
would error out immediately after start, complaining about how
`logger.logerror` didn't exist. As it turns out, the inclusion of
`logger = require("./logger")` at the end of the file caused the
simplistic logger to be overwritten with Haraka's custom logger, which
doesn't have any log* (logdebug, loginfo, and logerror) functions. This
makes sure that logger has been properly included - if it hasn't, it
falls back to outputting directly to console - and calls `logger.log`
with the appropriate log level (`DEBUG`, `INFO`, and `ERROR`,
respectively).

- Basic change to the way logging is required in `configfile.js`

Checklist:
- [ ] docs updated (N/A - No feature added)
- [ ] tests updated (N/A - No feature added)

### DIFF
--- a/configfile.js
+++ b/configfile.js
@@ -6,7 +6,7 @@ var path = require('path');
 var yaml = require('js-yaml');
 var logger = require('./logger')
 
-logger.logdebug = function() {
+logger.logdebug = function () {
     if (!logger.log) {
         console.log.apply(console, arguments);
     }
@@ -15,8 +15,8 @@ logger.logdebug = function() {
     }
 }
 
-logger.loginfo = function() {
-    if(!logger.log) {
+logger.loginfo = function () {
+    if (!logger.log) {
         console.log.apply(console, arguments);
     }
     else {
@@ -24,8 +24,8 @@ logger.loginfo = function() {
     }
 }
 
-logger.logerror = function() {
-    if(!logger.log) {
+logger.logerror = function () {
+    if (!logger.log) {
         console.error.apply(console, arguments);
     }
     else {
--- a/configfile.js
+++ b/configfile.js
@@ -4,8 +4,34 @@
 var fs   = require('fs');
 var path = require('path');
 var yaml = require('js-yaml');
+var logger = require('./logger')
 
-var logger = getStubLogger();
+logger.logdebug = function() {
+    if (!logger.log) {
+        console.log.apply(console, arguments);
+    }
+    else {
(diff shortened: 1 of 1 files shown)


baudehlo:
It feels like there's got to be a better way to do this. @msimerson any thoughts?

Perhaps for v3 logger should be a (singleton) class we instantiate, so we can always ensure it can work properly when we have circular require()s?

msimerson:
> It feels like there's got to be a better way to do this. @msimerson any thoughts?

I feel exactly the same way. I've hated leaving this PR hanging, but I know this isn't any better than what we've got, and at least in case, I think it's a regression. There's a better solution out there, it just hasn't bubbled up yet.

wuvvy:
I apologize if I haven't actually explained what happens in enough detail.

The issue actually occurs _after_ the logger class has been included, which in the old code, overrode the polyfill logger "class" that was there to prevent issues with the circular reference. Since the logger class doesn't have the `log*` classes defined within it, the program would just terminate when attempting to alert me that it can't watch the directory. 

My PR was basically writing a polyfill that would attempt to use the logger class first, but if that hasn't been able to be initialized, use the `console.log` polyfill until the logger class can be properly required.

msimerson:
> I apologize if I haven't actually explained what happens in enough detail.

Sorry, it's not you, or even your PR. The problem is that we've got some circular dependencies and loading logger near the top of configfile is something we've done before and it creates different issues. Hence @baudehlo 's comment regarding making our logger a singleton.

baudehlo:
Isn't a simpler fix to move that load into a setImmediate() call? (I had to do that in the outbound code too).

baudehlo:
We need to get this fixed before the release. Is this PR likely to be updated or should I do a simpler PR with my change?

msimerson:
I'm in favor of the latter.

wuvvy:
Latter is fine with me, honestly

--
Austin Bolstridge
Eagle Scout

baudehlo:
@wuvvy can you check if #1948 fixes the issue for you?

baudehlo:
Ping?

baudehlo:
I'm giving this issue 24 hours. If no further comment I'm closing the PR and merging the other one.

PR #1849, opened 2017-03-09
assure conn/tran still exists before storing results

msimerson:


plugins/rspamd.js
```diff
@@ -131,6 +131,8 @@ exports.hook_data_post = function (next, connection) {
     }
 
     timer = setTimeout(function () {
+        if (!connection) return;
```
baudehlo: I don't think there's any way that can happen (if you set it to null elsewhere it's still an object in the plugin), but I guess we can let it slide.

PR #1851, opened 2017-03-10
tls consistency cleanups

msimerson:
- doc cleanups
- plugins/tls: export net_utils (so tests can override config dir)
- remove unnecessary punctuation
- move loadPemFiles to new function (in anticipation of tls.load_tls_opts gaining more code)
- add tests/config/tls/haraka.local.pem (test TLS dir & cert)
- replace some + ‘string‘ + patterns with string interpolation
- replace some `function (a,b) {` patterns with `(a,b) => {`


msimerson:
I have reverted the SNI changes in this PR. Server side SNI will appear in a future PR.

PR #1856, opened 2017-03-14
Added missing 'default' keyword in rspamd plugin. Score wasn't added …

gtech99:
r.data which contains the rspamd answer also contains the 'default' index which was omitted by the author in a few places. Because of this, scores weren't added properly to headers.

### DIFF
--- a/plugins/log.elasticsearch.js
+++ b/plugins/log.elasticsearch.js
@@ -336,8 +336,8 @@ exports.populate_message = function (pir, connection) {
     }
 
     if (pir.rcpt_to && pir.rcpt_to.recipient) {
-	for (var key in pir.rcpt_to.recipient) {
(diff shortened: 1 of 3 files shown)


msimerson:
Did the code you're changing never work correctly? Does it only work correctly with some versions of rspamd? 

There are tests in place that suggest the code used to work as intended, and your changes break them. Either your changes aren't correct, or the tests must be updated.

gtech99:
I need to update the tests as well. The code never worked correctly.

 

George

swerter:
+1 R-Spamd score is now added correctly

gtech99:
modified ES logging of envelope/recipients as lowercase

gtech99:
Fixed rspamd tests

msimerson:
Looks good, thanks @gtech99

PR #1861, opened 2017-03-24
Binary log file

juryghidinelli:
The log file without that replacement, is interpreted as binary file log issue.

### DIFF
--- a/plugins/clamd.js
+++ b/plugins/clamd.js
@@ -232,7 +232,7 @@ exports.hook_data_post = function (next, connection) {
 
         var result = '';
         socket.on('line', function (line) {
-            connection.logprotocol(plugin, 'C:' + line);
+            connection.logprotocol(plugin, 'C:' + line.replace(/\x00/,''));
             result = line.replace(/\r?\n/, '');
         });
 
--- a/plugins/clamd.js
+++ b/plugins/clamd.js
@@ -232,7 +232,7 @@ exports.hook_data_post = function (next, connection) {
 
         var result = '';
         socket.on('line', function (line) {
-            connection.logprotocol(plugin, 'C:' + line.replace(/\x00/,''));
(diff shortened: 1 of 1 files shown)


baudehlo:
Which log plugin are you using?

juryghidinelli:
The plugin that i use is https://github.com/haraka/haraka-plugin-log-reader

plugins/clamd.js
```diff
@@ -232,7 +232,7 @@ exports.hook_data_post = function (next, connection) {
 
         var result = '';
         socket.on('line', function (line) {
-            connection.logprotocol(plugin, 'C:' + line);
+            connection.logprotocol(plugin, 'C:' + line.split('').filter(function(x) { var n = x.charCodeAt(0); return 31 < n && 127 > n  }).join('') );
```
msimerson: We can't merge b/c lint tests are failing. We require a space after a function declaration. IE, this change:
```diff
-connection.logprotocol(plugin, 'C:' + line.split('').filter(function(x) { var n = x.charCodeAt(0); return 31 < n && 127 > n  }).join('') );
+connection.logprotocol(plugin, 'C:' + line.split('').filter(function (x) { var n = x.charCodeAt(0); return 31 < n && 127 > n  }).join('') );
```

baudehlo:
And I'd rather not merge if it turns out the fix should be in logger.*
functions instead. Any plugin could log a \x00, not just this one.

msimerson:
> And I'd rather not merge if it turns out the fix should be in logger.* functions instead. Any plugin could log a \x00, not just this one.

I'm not so sure that's the best idea. Checking every log entry (and Haraka emits TONS of them) for the presence of binary characters sounds a LOT more expensive than applying the filters in the few locations where binary may be output.

baudehlo:
True. But if it turns out to be in plugin-log-reader that might not be a
bad place to fix it.

PR #1866, opened 2017-03-27
Added RabbitMQ vhost support

alexmtv:
Added RabbitMQ vhost support. Original plugin doesn't support it. 

Reference
https://github.com/squaremo/amqp.node/blob/master/lib/connect.js
http://stackoverflow.com/questions/24945112/can-i-specify-rabbitmq-credentials-in-node-js

- new variable to support vhost (with same name)
- 

Checklist:
- [x] docs updated
- [ ] tests updated


plugins/queue/rabbitmq_amqplib.js
```diff
@@ -39,7 +40,7 @@ exports.init_amqp_connection = function () {
     var autoDelete = cfg.autoDelete === "true" || false;
     deliveryMode = cfg.deliveryMode || 2;
 
-    amqp.connect("amqp://"+user+":"+password+"@"+host+":"+port, function (err, conn){
+    amqp.connect("amqp://"+user+":"+password+"@"+host+":"+port+vhost, function (err, conn){
```
msimerson: In looking at the cited URL, it appears the vhost name is appended after the port with a `/` delimiter. I don't see that getting added anywhere.

alexmtv: Yes, my sugestion is to include it in config file as part of the variable. Like "vhost=/haraka"

msimerson: I must have missed that suggestion in your documentation updates. Hint hint.

alexmtv: Done!

PR #1871, opened 2017-03-28
tls: remove interim variables

msimerson:
- server/tls upgrade: remove interim TLS vars
- only emit OCSP not loaded error at startup (vs every connection)
- remove unused _controlReleased


msimerson:
The only change in here that I'm not entirely certain of is `if (cleartext.getCipher) {`. Near as I can tell, the only reason to have ever included that is because node < 0.7 didn't have `getCipher()`. I've been running this in production for a few days to make sure and have seen no consequences.

PR #1902, opened 2017-05-04
1861 binary log

msimerson:
don't emit binary characters into the logs.

- I intend to self merge and close this issue, unless someone has a good reason why binary data in the log files is actually useful.
- I'm also going to address the original issue at the other end, and tell grep to always assume the log file is ascii (`grep --text`). 
    - done. [haraka/haraka-plugin-log-reader#11]

closes #1861


msimerson:
thanks @smfreegard

PR #1908, opened 2017-05-08
F/outbound client certificates

### LINKED ISSUE #1905: Client Certificate on outbound TLS connections
This is a follow up to the issue I published yesterday.

I am trying to use the queue/smtp_forward plugin to relay messages to Office 365. The authorisation on SMTP connections to Office 365 is done via a Client Certificate.

Currently, SMTPClient does not allow a client certificate to be set. As a proof of concept, I prepended "key" and "cert" properties to the tls_options object before the _getSecureContext is called. This works as expected.

I propose a solution where given the configuration:
###### smtp_forward.ini
```
enable_client_cert=true
```

###### tls.ini
```
[queue/smtp_forward]
key=key.key
cert=cert.pem
```

Client Certificates would be used in the outbound connection.

I have read reports about some email servers breaking when client certificates are used, so the configuration is intended to be verbose.

chasent:
Fixes #1905 

Checklist:
- [ x] docs updated
- [ x] tests updated

### DIFF
--- a/plugins/queue/smtp_forward.js
+++ b/plugins/queue/smtp_forward.js
@@ -6,10 +6,19 @@
 
 var smtp_client_mod = require('./smtp_client');
 
+// exported so tests can override config dir
+exports.get_net_utils = function() {
+    return require('haraka-net-utils');
+}
+
 exports.register = function () {
     var plugin = this;
+    plugin.load_errs = [];
 
     plugin.load_smtp_forward_ini();
+    plugin.make_tls_opts();
+
+    if (plugin.load_errs.length > 0) return;
 
     if (plugin.cfg.main.check_sender) {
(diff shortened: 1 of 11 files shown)


plugins/queue/smtp_forward.js
```diff
@@ -6,10 +6,19 @@
 
 var smtp_client_mod = require('./smtp_client');
 
+// exported so tests can override config dir
+exports.get_net_utils = function () {
+    return require('haraka-net-utils');
+}
```
msimerson: What we often do in a case like this is instead:

```js
exports.net_utils = require('haraka-net-utils'); 
```

That lets you override net_utils config dir from tests and without having to run the node module loader every time that code is run in production (many many many millions of times).

PR #1922, opened 2017-05-16
Set the date of edits

acharkizakaria:
Fixes #

- 
- 
- 

Checklist:
- [ ] docs updated
- [ ] tests updated


msimerson:
That date is wrong. It will get updates when 2.8.14 is released.

PR #1927, opened 2017-05-20
Set graceful shutdown off by default

### LINKED ISSUE #1926: cluster shutdown/restart issues
### Background

We recently upgraded from haraka 2.7.something to 2.8.13.  We use clustering with 20 workers per instance.  After the upgrade, I noticed that haraka would now take ~20 minutes to shut down after it received a SIGTERM, when it used to happen instantaneously.  This was after spending some time fixing our plugins to properly shutdown when requested, which I initially assumed was causing the issue.

### Problem

I traced the problem back to open connections from clients sending us mail (or not, but leaving the connection open anyway).  Haraka doesn't appear to track client connections anywhere explicitly after they're created, but of course open connections prevent the node main loop from exiting, even after everything else in each worker has been properly closed *and* worker.disconnect() has been called.  This is exacerbated by the way workers are shut down [in serial](https://github.com/haraka/Haraka/blob/f42e75849bcb99b8dce9893baa17541e9db07bc7/server.js#L135) and waited on ([up to 1 minute](https://github.com/haraka/Haraka/blob/f42e75849bcb99b8dce9893baa17541e9db07bc7/server.js#L129)) to fully exit.  For deployments of sufficient volume (where you're almost guaranteed at least a few client connections will hold the line without doing anything) using clustering, this means that shutting down haraka now takes as many minutes as there are workers.

### Steps to Reproduce

* Set your smtp.ini with `nodes=2` or more.
* Run haraka, e.g. from master: `node haraka.js`.
* Open a few connections to haraka, e.g. `telnet localhost 25`.
* Leave those connections open and untouched, and send `SIGTERM` to the master process, e.g. `kill ...`.
* Observe that haraka starts to shut down immediately, but stays up, and the connections stay active.
* After one minute, whatever connections that hit the first haraka worker will get closed and haraka will move on to the second worker, etc.
* The whole process only goes down after as many minutes as you had workers with open connections.

I'd expect it to go down in a fixed, even if somewhat long, amount of time.  It's not a huge issue, but I'd expect anyone running the latest version at scale to see the same thing, and it makes scaling slightly more painful.

### Potential Fixes

I'm guessing it's a feature that connections are allowed time to wrap up and close on their own.  I'd love to see the connections being tracked (bonus points for actually keeping tabs on what part of the conversation they're in and waiting until the transaction isn't in a state where dropping the connection might lead to data loss, e.g. after `data_post` but before it's been confirmed from the `smtp_forward` host) and closed after some shorter timeout than the workers themselves get, especially if inactive.

I think there's a quick solution here that gets us part of the way there without changing too much code:  how would people feel about changing the worker shutdown logic to use `async.eachLimit()` instead of [`async.eachSeries()`](https://github.com/haraka/Haraka/blob/f42e75849bcb99b8dce9893baa17541e9db07bc7/server.js#L135), with a limit of `nodes / 2`.  That way there are still always nodes up listening for connections, as is needed if you're doing a graceful restart, but the total time is always capped at a fixed value, instead of growing with the number of nodes.

If that seems like a good idea, I'd be happy to whip up a patch (I'd also take a stab at getting those worker timeouts to be configurable).

All of this was tested on master, node v4.2.6, linux 4.4.0-72-generic x86_64 (ubuntu 16.04).

### Bonus Problem

It doesn't appear that the graceful restart option actually works if you have cluster enabled.  If you use the bin to do `--graceful`, it properly connects to a node and goes through the `INTERNALCMD` rigamarole, even [sending the master process the right message](https://github.com/haraka/Haraka/blob/f42e75849bcb99b8dce9893baa17541e9db07bc7/server.js#L208).  The master process doesn't seem to be listening for that message anywhere though, so it never does the restart.

I'd be happy to include this in my patch, providing someone can give me security guidance on it (i.e. the `INTERNALCMD` mechanism is required to come from localhost *and* have the secret key, and I'm guessing node implements this IPC mechanism as a pipe, but if it's a named pipe or socket of some kind it should get the same treatment I assume).

Also happy to break this out into a separate issue either way.

baudehlo:
Fixes #1926

- Still allow haraka -c <path> --graceful
- Also sets graceful shutdown to shutdown in max 2xChildren amount of time it sets to shutdown.
- Shutdown gracefully now requires smtp.ini to contain `graceful_shutdown=true`.

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/configfile.js
+++ b/configfile.js
@@ -266,6 +266,7 @@ cfreader.ensure_enoent_timer = function () {
             })(files[i]); // END BLOCK SCOPE
         }
     }, 60 * 1000);
+    cfreader._enoent_timer.unref();
 };
 
 process.on('message', function (msg) {
--- a/server.js
+++ b/server.js
@@ -1,6 +1,7 @@
 'use strict';
 // smtp network server
 
+// var log = require('why-is-node-running');
 var net         = require('./tls_socket');
 var conn        = require('./connection');
 var outbound    = require('./outbound');
@@ -27,6 +28,7 @@ Server.load_smtp_ini = function () {
     Server.cfg = Server.config.get('smtp.ini', {
         booleans: [
             '-main.daemonize',
+            '-main.graceful_shutdown',
         ],
     }, function () {
         Server.load_smtp_ini();
@@ -98,25 +100,28 @@ Server.gracefulRestart = function () {
 }
 
 Server.gracefulShutdown = function () {
+    if (!Server.cfg.main.graceful_shutdown) {
+        logger.loginfo("Shutting down.");
+        process.exit(0);
+    }
     logger.loginfo('Shutting down listeners');
     Server.listeners.forEach(function (server) {
         server.close();
     });
     Server._graceful(function () {
+        // log();
         logger.loginfo("Failed to shutdown naturally. Exiting.");
         process.exit(0);
     });
 }
 
 Server._graceful = function (shutdown) {
-    if (!Server.cluster) {
-        if (shutdown) {
-            ['outbound', 'cfreader', 'plugins'].forEach(function (module) {
-                process.emit('message', {event: module + '.shutdown'});
-            });
-            var t = setTimeout(shutdown, Server.cfg.main.force_shutdown_timeout * 1000);
-            return t.unref();
-        }
+    if (!Server.cluster && shutdown) {
+        ['outbound', 'cfreader', 'plugins'].forEach(function (module) {
+            process.emit('message', {event: module + '.shutdown'});
+        });
+        var t = setTimeout(shutdown, Server.cfg.main.force_shutdown_timeout * 1000);
+        return t.unref();
     }
 
     if (gracefull_in_progress) {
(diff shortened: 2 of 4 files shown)


msimerson:
Hrmm, the nodemailer integration test is failing with this change. Not sure why.

baudehlo:
Fixed that issue and made it a bit more sane.

configfile.js
```diff
@@ -266,6 +266,7 @@ cfreader.ensure_enoent_timer = function () {
             })(files[i]); // END BLOCK SCOPE
         }
     }, 60 * 1000);
+    cfreader._enoent_timer.unref();
```
baudehlo: BTW how come we now have a haraka-config package and this file is still in Haraka?

server.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 // smtp network server
 
+// var log = require('why-is-node-running');
```
baudehlo: BTW I left this in but commented because it makes it easy to debug any kind of "won't quit" issues in the future.

tls_socket.js
```diff
@@ -137,6 +137,10 @@ pluggableStream.prototype.setKeepAlive = function (bool) {
 pluggableStream.prototype.setNoDelay = function (/* true||false */) {
 };
 
+pluggableStream.prototype.unref = function () {
+    return this.targetsocket.unref();
+};
```
baudehlo: I added this in case we need to unref a socket in the future.

configfile.js
msimerson: Since moving config into a module (so npm plugins could depend on it) I've been moving more and more stuff to depend on haraka-config, with the eventual plan to deprecate the build-in conifg.js and configfile.js. We're a ways down that road and I don't recall right now what's left to do.

server.js
msimerson: Says the guy that use to be adamantly opposed to such a thing. ;-)  Welcome to the dark side. ;-)

tls_socket.js
msimerson: Could this ever get called while targetsocket is undefined?

baudehlo:
No we never undef the targetsocket. Everything else would break too if we did.

PR #1929, opened 2017-05-21
Run dumped logs through plugins not console

### LINKED ISSUE #1914: Log Plugin Issue
I'm writing my own log plugin and I have it working to my satisfaction except that it never captures initial startup logs from "Starting up Haraka" to just after all the plugins are loaded.  The first entry I ever receive is "Listening on ...".  It seems that at the end of `plugins.load_plugins` a call to `logger.dump_logs` is made.  This call dumps the buffered startup log output directly to the console ignoring any log hooks.  If I comment out the call to `logger.dump_logs`, then all log output is properly captured by my plugin.  Is this the intended behavior?  I would prefer to have all log info go through my hooks.

baudehlo:
Fixes #1914 

- In dump_logs() run that dumping through plugins

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -56,13 +56,7 @@ logger.colorize = function (color, str) {
 logger.dump_logs = function (cb) {
     while (logger.deferred_logs.length > 0) {
         var log_item = logger.deferred_logs.shift();
-        var color = logger.colors[log_item.level];
-        if (color && stdout_is_tty) {
-            console.log(logger.colorize(color, log_item.data));
-        }
-        else {
-            console.log(log_item.data);
-        }
+        plugins.run_hooks('log', logger, log_item);
     }
     // Run callback after flush
     if (cb) process.stdout.write('', cb);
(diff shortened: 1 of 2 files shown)


logger.js
```diff
@@ -56,13 +56,7 @@ logger.colorize = function (color, str) {
...
     while (logger.deferred_logs.length > 0) {
         var log_item = logger.deferred_logs.shift();
-        var color = logger.colors[log_item.level];
-        if (color && stdout_is_tty) {
-            console.log(logger.colorize(color, log_item.data));
-        }
-        else {
-            console.log(log_item.data);
-        }
+        plugins.run_hooks('log', logger, log_item);
```
msimerson: Will this still dump logs to console if there's no logging plugin enabled? If so, +1 from me.

baudehlo:
Yup. Tested that. Ran it without any plugins and then installed syslog going to mail. (Hence my other change to the syslog plugin). I think we just never re visited this system after making logging pluggable. I went back in the history of the file and it looks like we just never revisited this.

PR #1931, opened 2017-05-22
Allow "Unknown Result" and Socket Error to Try Next Host

### LINKED ISSUE #1924: Issues with clamd
In the clamd plugin, I find myself getting at least 1 socket read ECONNRESET error and at least 15 unknown result errors (with the unknown result being entirely blank) per day.  I have addressed this by forcing a call to try_next_host() in these 2 conditions.  Currently, I'm only running one local instance of ClamAV and so I list localhost twice in the list of hosts.  It always works for me correctly the second time and thus pevents all 450 responses due to a failed virus scan.  However, despite my odd situation, it doesn't seem unreasonable that in error cases and unknown result cases that additional hosts should be tried if there are multiple hosts listed.  I think I would want this even if my installation wasn't so prone to errors (which I assume isn't the norm for most people).  So, for example in the unknown result case, the modified code would look like below (with the top 'if' being new) at the bottom of try_next_host:
```
if (hosts.length) {
    connection.logwarn(plugin, 'unknown result: "' + result + '" from host ' + host);
    socket.destroy();
    return try_next_host();
}       
txn.results.add(plugin, { err: 'unknown result: "' + result + '" from host ' + host });
if (!plugin.cfg.reject.error) return next();
return next(DENYSOFT, 'Error running virus scanner');
```

If you feel like this makes sense, I can submit a pull request.  Otherwise, I just leave this as a personal change to address my situation.

superman20:
Fixes #1924

- If multiple hosts are listed for clamd and an "unknown result" is returned or a socket error occurs after connection, then attempt to try remaining hosts before returning DENYSOFT.

### DIFF
--- a/plugins/clamd.js
+++ b/plugins/clamd.js
@@ -213,7 +213,7 @@ exports.hook_data_post = function (next, connection) {
                        'Connection to ' + host + ' failed: ' + err.message);
                 return try_next_host();
             }
-            
+
             // If an error occurred after connection and there are other hosts left to try,
             // then try those before returning DENYSOFT.
             if (hosts.length) {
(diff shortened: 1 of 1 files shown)


superman20:
I have tested this with corrupted archives, corrupted executables, misnamed archives (i.e, renaming a file that isn't a zip file to have a zip extension), misnamed executables, and password protected archives.  None of these situations activates the new code in this PR.  ClamAV seems to handle these situations properly and returns a defined response that is not interpreted as an error condition by the clamd plugin.  ClamAV even has options that allow you to control whether you want to allow "broken" executables and password protected content or block them.  

I agree with your concern completely.  If you can think of any other scenarios, I'd be happy to test those as well.

msimerson:
Thanks for considering and testing. And thanks for this contribution!

baudehlo:
Yeah thanks - in recognition of this and other code I've added you to the
Collaborators list for Haraka.

superman20:
Thanks!

PR #1932, opened 2017-05-22
prep release 2.8.14

msimerson:
- [ ] Start a N-week quiet period
- [x] Update [haraka-tld](https://github.com/haraka/haraka-tld)
- [x] Draft a release on the [releases page](https://github.com/haraka/Haraka/releases) (Save draft)
- Create a PR with:
    - [x] Updated [Changes](https://github.com/haraka/Haraka/blob/master/Changes)
    - [x] Bumped version in [package.json](https://github.com/haraka/Haraka/blob/master/package.json)
- For major version releases:
    * [x] check for waiting PRs for that milestone version.
- [x] Update release date in [Changes](https://github.com/haraka/Haraka/blob/master/Changes)
- [ ] `npm publish`
- [ ] Post `Changes` file on the draft release and publish the release 
- [ ] update [haraka's wikipedia page](https://en.wikipedia.org/wiki/Haraka_(software))
- [ ] Email the mailing list
- [ ] Publish a blog post
- [ ] Edit the /topic on #haraka IRC channel

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,7 +1,17 @@
 
-2.8.14 - May 27, 2017
+2.8.14 - Jun 26, 2017
 
 * Changes
+    * update async to 2.5.0 #1982
+    * update iconv to 2.3.0 #1981
+    * require node.js v6+ #1958
+    * update ipaddr.js to 1.4.0 #1972
+    * support newer address-rfc2822 #1970
+    * update node-address-rfc2821 version to 1.1.1 #1968
+    * outbound: be consistent with todo.domain #1960
+    * bump haraka-results required version #1949
+    * logger: load in a setImmediate call #1948
+    * logger: strip intermediate \n chars #1947
     * tls consistency cleanups #1851
     * Get pool config handling simplifcation #1868
         * add integration test: send message w/smtp_client
@@ -26,12 +36,22 @@
     * Add .editorconfig #1884
     * tls: remove interim variables #1871
 * New Features
+    * Use punycode domain (support SMTPUTF8) #1944
     * Added RabbitMQ vhost support #1866
     * clamav: allow "Unknown Result" and Socket Error to try next host #1931
     * outbound client certificates #1908
     * Implement the missing upgrade method on SMTPClient #1901
     * Remove typo from relay.md #1886
 * Fixes
+    * outbound: fix queue not loaded for single process #1941
+    * outbound: Fix undefined variable platformDOT in hmail.js #1943
+    * outbound: fix undefined FsyncWriteStream var #1953
+    * Fix cluster messaging for node v6+ #1938
+    * outbound: fix loading under cluster. #1934
+    * Check pool exists before delete #1937
+    * be more strict in attachment filename matching #1957
+    * doc typo fix #1963
+    * RabbitMQ: fix encoding of user and password string #1964
     * spf: improve modifier regexp #1859
     * rabbitmq doc typo in config file name #1865
     * URL to manual was 404, point to Plugins.md #1844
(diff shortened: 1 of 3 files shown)


baudehlo:
I think there's just 1 PR waiting now. That logging/configfile one.

msimerson:
Due to changes in haraka-results, these must also be done:

- [ ] publish new version of haraka-plugin-watch
- [ ] publish new version of haraka-plugin-karma

baudehlo:
Only issue I'd like to see fixed for the release now is #1954 - that's waiting on seeing if we can catch some logging from it.

superman20:
I actually found 1 more issue today, but I've not yet had time to get more debug info.  I can easily reproduce so I hope it will be simple to track down.  I'll make another post when I can get some more debug info, but the gist of the problem is that if you send an email to a non-existent domain, the bounce that Haraka generates loses the domain somewhere when bouncing back to the originating mail server.  So, if user@domain.com sends an email to bad@nonexistentdomain.com, then Haraka tries to deliver the bounce to just "user" and not "user@domain.com".  Obviously, the originating server rejects the bounce with an invalid recipient.

baudehlo:
@superman20 is there any chance #1960 fixed that? Most of the changes were in the bounce area.

superman20:
I just updated to current Master (plus my debug changes for #1954) and I still see the same problem.  I'm going to try and track down more details.  I currently use [this](https://gist.github.com/smfreegard/ff79d02aeb94b9065359) as my outbound queue.  Do you think it has any relation to this problem?

baudehlo:
No that shouldn't matter. I mean even if it does it's still a bad bug.

superman20:
First off, let me know if you'd prefer I start a separate thread for this.  I'm starting to find the problem, but I wouldn't mind some advice to help narrow down to the real bug more quickly.  So, I'm currently in `rcpt_to.routes.js` at line 113:
`address = hmail.todo.rcpt_to[0].address().toLowerCase();`
In my case, the bounce e-mail's rcpt_to[0] object looks like this:
```
Address {
  original: 'email@domain.com',
  user: 'email',
  host: 'domain.com' }
```
The relevant address function line in `haraka/node-address-rfc2821` looks like:
```return (this.user || '') + (this.original_host ? ('@' + (use_punycode ? this.host : this.original_host)) : '');```
With the Address object I have, the logic only returns "email" instead of "email@domain.com".  So the error could be here (which seems doubtful) or in the construction of the Address object for the bounce, or some other place I'm not thinking of?

superman20:
Maybe something going on with [this commit](https://github.com/haraka/node-address-rfc2821/commit/74d906308122cb90fb5457bda97adb05af209e04) which seems to produce the change from host to original_host?

baudehlo:
Yeah that module got changed recently. I'll look into it.

superman20:
@baudehlo Did you get a chance to look into this?  I know you're wanting to release soon.  Let me know if you'd like me to do something.

baudehlo:
Yes I fixed that today.

superman20:
Ah yes...I see it now.  My apologies.  That notification got buried in the inbox.

msimerson:
The issue causing test failures in this PR is eslint 4 which is being addressed in #1979

msimerson:
Changes.md is now up to date again. 2.8.14 is overdue for getting pushed out the door.

baudehlo:
Yup. I'm clearly not going to solve the outbound weird under load bugs any time soon. So let's get it out.

msimerson:
````
$ npm publish
+ Haraka@2.8.14
````

PR #1941, opened 2017-05-24
Fix queue not loaded for single process

### LINKED ISSUE #1940: Outbound queue is never loaded for single process
It appears that for a single process, the Outbound queue is never loaded on startup.  I believe the logic that controls this is in server.js, line 480-490:
```
    var c = Server.cfg.main;
    Server.ready = 1;

    // Load the queue if we're just one process
    if (!(cluster && c.nodes)) {
        if (c.outbound) {
            outbound.load_queue();
        }
        Server.setup_http_listeners();
        return;
    }
```
For my installation, `c.outbound` is never defined.  There is also no documented "outbound" option in smtp.ini and I can't find any code that would create it.

Perhaps the test is supposed to be `if (outbound) {` ?

superman20:
Fixes #1940

### DIFF
--- a/server.js
+++ b/server.js
@@ -482,7 +482,7 @@ Server.init_master_respond = function (retval, msg) {
 
     // Load the queue if we're just one process
     if (!(cluster && c.nodes)) {
-        if (c.outbound) {
+        if (outbound) {
             outbound.load_queue();
         }
         Server.setup_http_listeners();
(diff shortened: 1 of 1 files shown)


baudehlo:
Weird - this has been broken since: aade1949ffc11f3e204e40e2a1326d501327bdac

Can you just take out the condition? It has no place here.

PR #1944, opened 2017-05-24
Use punycode domain

### LINKED ISSUE #1864: Feature request: SMTPUTF8 support (RFC6531)
I didn't find any mentions on supporting SMTPUTF8 extension. Haraka already supports 8BITMIME so SMTPUTF8 would be a great addition to this.

It is mostly needed to support Email Address Internationalization (EAI) email addresses where both user and domain part can contain non-latin characters. Without SMTPUTF8 it is only possible to use the A-label formatted IDN domains (xn--...) but no way to use non-latin characters in user part.

**What needs to be changed to support SMTPUTF8:**

1. The server should add SMTPUTF8 as one of the supported extensions

```
S: 250-Macintosh.lan Hello Unknown [::1], Haraka is at your service.
S: 250-PIPELINING
S: 250-8BITMIME
S: 250 SMTPUTF8 <-- new keyword
```

2. The server must switch to unicode mode if the client uses SMTPUTF8 keyword with MAIL FROM

Without SMTPUTF8 keyword:

```
C: MAIL FROM:<andris.reinman@gmail.com>
S: 250 sender <andris.reinman@gmail.com> OK
C: RCPT TO:<андрис@уайлддак.орг>
S: 552 Syntax error (8-bit characters not allowed)
```

With SMTPUTF8 keyword:

```
C: MAIL FROM:<andris.reinman@gmail.com> SMTPUTF8
S: 250 Accepted
C: RCPT TO:<андрис@уайлддак.орг>
S: 250 recipient <андрис@уайлддак.орг> OK
```

If SMTPUTF8 keyword is used then UTF8 encoded addresses should be allowed to use with both MAIL FROM and RCPT TO commands.

I don't know much about Haraka internals, so I don't know if this would be something simple or difficult to implement. If it would be too hard to implement in the short term, then maybe you'd take it into the long term plans? I'm pretty sure that EAI addresses are the future.

----

The email address андрис@уайлддак.орг is a working address. I host it in my IMAP server [Wild Duck](https://github.com/wildduck-email/wildduck). Currently I have a my own SMTP server instance running as MX but I would rather use Haraka for this.

baudehlo:
Fixes #1864 

- Uses a flag in the new (unreleased as of today) address-rfc2821 which allows you to ask for the punycode domain if it's there. 
- I should probably also just enable SMTPUTF8 capabilities with this PR. Thoughts?

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -196,6 +196,7 @@ function Connection (client, server) {
     this.current_data = null;
     this.current_line = null;
     this.state = states.STATE_PAUSE;
+    this.encoding = 'utf8';
     this.prev_state = null;
     this.loop_code = null;
     this.loop_msg = null;
@@ -285,7 +286,7 @@ Connection.prototype.process_line = function (line) {
         return;
     }
 
-    this.current_line = line.toString('binary').replace(/\r?\n/, '');
+    this.current_line = line.toString(this.encoding).replace(/\r?\n/, '');
     if (logger.would_log(logger.LOGPROTOCOL)) {
         this.logprotocol("C: " + this.current_line + ' state=' + this.state);
     }
@@ -304,7 +305,7 @@ Connection.prototype.process_line = function (line) {
             this.disconnect();
             return;
         }
-        else {
+        else if (this.hello.verb == 'HELO') {
             return this.respond(501, 'Syntax error (8-bit characters not allowed)');
         }
     }
@@ -1368,6 +1369,7 @@ Connection.prototype.cmd_mail = function (line) {
         self.transaction.mail_from = from;
         if (self.hello.verb == 'HELO') {
             self.transaction.encoding = 'binary';
+            self.encoding = 'binary';
         }
         plugins.run_hooks('mail', self, [from, params]);
     });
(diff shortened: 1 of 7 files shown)


baudehlo:
I don't think disabled by default is a good idea. It's disabled if you use HELO of course.

Correct - it means that the library returns the pretty version by default. This mimicks how the rest of the library works - format() should basically give you what you put in. I did think about whether that was the right thing to do or not, but when you're using .format() you generally want it to look right.

An alternative would be to change .address() to use the punycode domain by default, and change outbound to use that function instead and add it's own `<>`s. I think that's reasonable in terms of the naming of the API.

Regarding the last comment, actually a lot more needs changed too I think, so don't merge yet.

baudehlo:
Fixes to connection.js done. Let me know what you think of my other comment.

baudehlo:
Then can anyone give me help with the changes required for smtp_client.js? (or _forward and _proxy if that's where the changes are needed?). I think it those modules use the magic .toString() (which defaults to the utf8 domain) so it's a bit confusing.

baudehlo:
OK any objections to merging this?

smfreegard:
No - LGTM.

PR #1947, opened 2017-05-25
Remove intermediate \n chars from protocol level logs

### LINKED ISSUE #1946: Protocol logging question
Starting on line 80 of logger.js we see:
```
logger.log = function (level, data) {
    if (level === 'PROTOCOL') {
        data = data.replace(/\n/g, '\\n\n');
    }
```
Is there a reason why the actual line-feed character is kept in the output?  This makes some of the protocol log entries wrap to the next line.  It seems like this would be generally unfavorable...but I assume there is some other reason I haven't thought of to keep it?

For an example, SpamAssassin output looks like this since the \n is retained:
```
[00:04:38] [PROTOCOL] [8211E3C4-D775-429E-ABFC-2861568571C8.1] [spamassassin] Spamd C: SPAMD/1.1 0 EX_OK\r\n
 state=line0
[00:04:38] [PROTOCOL] [8211E3C4-D775-429E-ABFC-2861568571C8.1] [spamassassin] Spamd C: Content-length: 2595\r\n
 state=response
[00:04:38] [PROTOCOL] [8211E3C4-D775-429E-ABFC-2861568571C8.1] [spamassassin] Spamd C: Spam: True ; 40.2 / 5.0\r\n
 state=response
[00:04:38] [PROTOCOL] [8211E3C4-D775-429E-ABFC-2861568571C8.1] [spamassassin] Spamd C: \r\n
 state=response
```
In my opinion, having the complete log entry on 1 line makes the logs easier to read.

superman20:
Fixes #1946

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -79,7 +79,7 @@ logger.dump_and_exit = function (code) {
 
 logger.log = function (level, data) {
     if (level === 'PROTOCOL') {
-        data = data.replace(/\n/g, '\\n\n');
+        data = data.replace(/\n/g, '\\n');
     }
     data = data.replace(/\r/g, '\\r')
                .replace(/\n$/, '');


logger.js
```diff
@@ -79,7 +79,7 @@ logger.dump_and_exit = function (code) {
 
 logger.log = function (level, data) {
     if (level === 'PROTOCOL') {
-        data = data.replace(/\n/g, '\\n\n');
+        data = data.replace(/\n/g, '\\n');
```
msimerson: when you consider `.replace(/\n$/, '');` from 2 lines below, is there any point at all in doing this replace?

superman20: That replace only grabs the last one at the end of the line.

msimerson: Exactly. When else would there ever be a `\n\n` in the line? (possible answer: wrapped headers, but then, don't you want to strip them anyway and have you're really long lines that your editor handles?)

msimerson: (btw, these are questions of curiosity. You've given more thought to this than anyone else in a long while so I'm inclined to +1 this PR on that basis alone.)

superman20: No problems.  I don't mind a good discussion to make sure the best solution ends up in Haraka.  Feel free to disagree with me any time.

I did consider combing the logic, but decided against it because it also changes other output (like DEBUG) and I was attempting to be minimally invasive.  If other "wrapped" output (especially headers) showed up very frequently, I would definitely want to get rid of the wrapped lines.  With my configuration (with DEBUG logging on), the only output I see that has wrapped content is a couple of TLS DEBUG lines at startup and the spamassassin protocol.  I can live with the wrapped DEBUG output at startup because it doesn't happen often and honestly that wrapping is helpful.  The spamassassin output is nasty because it replays the e-mail's headers which can have unwieldy output (especially from Microsoft's crap).  Having the frequent stuff stay on one line makes it a ton easier to read the logs and diagnose problems.

Perhaps another reason not to universally combine everything on 1 line is for debugging purposes.  I have found it helpful to dump objects to the log while debugging stuff.  I would not want these object's output combined on 1 line.  However, I also wouldn't want the object's output dumped regularly in a "production" server.

To make the logs useful an application should tend to dump manageable and useful single line content and Haraka really does a good job at that already (except for the protocol piece I have brought up).

So, I'm happy with the PR I submitted.  We'll only alter the PROTOCOL to keep it on one line.  The `.replace(/\n$/, '');` line is still needed to get rid of any rogue \n chars at the end of lines because manner logger systems will add one for you and we don't need 2.  Obviously for PROTOCOL lines that replace will do nothing.

Sorry for the long speech....just spelling out my justification.

msimerson:
Thanks for the PR and the thoughtful consideration.

PR #1951, opened 2017-06-05
Add node v8

baudehlo:
This just adds v8 to travis.

Currently failing due to: https://github.com/nodejs/node/issues/13325

Edit: Newer bug report: https://github.com/nodejs/node/issues/13548

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,19 @@
 language: node_js
 node_js:
-#   - "0.6"     # no longer supported by async
-#   - "0.8"     # no longer supported by iconv
-#   - "0.10"    # no longer maintained by node.js (2016-10-31)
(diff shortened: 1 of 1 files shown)


msimerson:
I rebased this against master and now the v7 test failure is resolved.

msimerson:
I have added node v8 to `allow_failures` so that the test suite is run but failures on v8 only don't cause the test suite to fail. That permits this PR (and subsequent) to be merged, will run the v8 tests against our code, and when the upstream issue gets fixed, we can remove v8 from the allowed_failures list.

superman20:
What impact does this upstream bug have on Haraka?  Is it a serious issue such that you wouldn't run v8 until resolved or is it a fringe issue only encountered when using certain features?

baudehlo:
It means Haraka won't start.

PR #1953, opened 2017-06-06
Fix undefined FsyncWriteStream var

superman20:
This is the obvious fix to using an undefined var FsyncWriteStream in `split_to_new_recipients`.  If you'd prefer some other refactoring fix, then let me know.


baudehlo:
Can you move it down after the other `./` requires please. Other than that it's good. I don't understand why eslint didn't catch this.

superman20:
No problem.  I only stuck it there to mimic the organization of index.js.

PR #1957, opened 2017-06-09
Be more strict in attachment filename matching

### LINKED ISSUE #1956: Incorrectly processing attachments with apostrophe in filenames
Haraka will truncate a filename when extracting the attachment filename.

e.g. filename: Demande d'ouverture compte.pdf

The problem is located here: https://github.com/haraka/Haraka/blob/master/mailbody.js#L118

We need to parse that out more intelligently.

baudehlo:
This copes with filename="d'euvre"

Fixes #1956 

- Drop support for filename='foo' which isn't RFC compliant anyway

Checklist:
- [ ] docs updated
- [x] tests updated

### DIFF
--- a/mailbody.js
+++ b/mailbody.js
@@ -115,14 +115,14 @@ Body.prototype.parse_start = function (line) {
         this.state = 'body';
     }
     else if (/^multipart\//i.test(ct)) {
-        match = ct.match(/boundary\s*=\s*["']?([^"';]+)["']?/i);
+        match = ct.match(/boundary\s*=\s*"?([^";]+)"?/i);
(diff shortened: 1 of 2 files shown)


baudehlo:
Working on a test. Hold off for the weekend maybe :)

baudehlo:
OK, test added - it's a pretty simple test but does the trick.

msimerson:
Perfect, looks great.

baudehlo:
BTW I just double checked the RFCs to make sure I'm not screwing up here. The value is `token | quoted-string`, and `quoted-string` is always quoted with double quotes.

PR #1958, opened 2017-06-10
require node LTS version (6+)

msimerson:
- drop node.js v4 support
- add node.js v6 testing
- add node.js v7 testing


msimerson:
node v4 is in maintenance mode, which means they'll only be fixing critical bugs going forward. But the bigger reason to drop it is that node v4 is bundled with npm v2, which has the goofy nested install structure that causes issues with plugins that want to inherit other plugins. Lets be done with v4.

PR #2016, opened 2017-07-25
build_todo() is part of the outbound/index.js api

nebulade:
Fixes a crash when handling an email for a non-existing user. This came up after getting over the first crash related to this case: https://github.com/haraka/Haraka/pull/2011

Not entirely sure if this is the correct fix, as it was not clear to me what the "todo" item would be.


baudehlo:
Nice catch. Merged.

Issue #1954, opened 2017-06-07
Error: Callback was already called


superman20:
I'm getting this crash about once or twice a week on average.  I can't quite put together what the problem may be.  During the times of failure, it appears the server is processing several mails simultaneously, so that makes it more difficult to pin down (or is a clue).  However, in all of the cases just before the crash, there does appear to be a socket or delivery failure of various causes (time out, hang up, grey-listing delay, etc), so perhaps that is a clue.  It is always in Outbound delivery.  I'll show a less busy log snippet as an example:
```
[11:10:43] [ERROR] [-] [outbound] Ongoing connection failed to 25.10.148.12:25 : Error: socket hang up
[11:10:43] [DEBUG] [-] [core] [outbound] release_client: 25.10.148.12:25 to undefined
[11:10:43] [DEBUG] [-] [core] Temp fail for: Tried all MXs
[11:10:43] [DEBUG] [-] [outbound] running deferred hooks
[11:10:43] [INFO] [-] [outbound] Temp failing 1496848206803_1496848302397_1_7660_YYOOgO_411_omega12 for 128 seconds: Tried all MXs
[11:10:43] [CRIT] [-] [core] Error: Callback was already called.
[11:10:43] [CRIT] [-] [core]     at C:\Haraka\npm\node_modules\Haraka\node_modules\async\dist\async.js:903:32
[11:10:43] [CRIT] [-] [core]     at HMailItem.hmail.next_cb (C:\Haraka\npm\node_modules\Haraka\outbound\queue.js:41:9)
[11:10:43] [CRIT] [-] [core]     at C:\Haraka\npm\node_modules\Haraka\outbound\hmail.js:1251:15
[11:10:43] [CRIT] [-] [core]     at FSReqWrap.oncomplete (fs.js:114:15)
[11:10:43] [NOTICE] [-] [core] Shutting down
```
In every case, the server immediately restarts and then successfully delivers whatever mail it crashed on.  I've tried to find a technique to reproduce this error but have been unsuccessful.  I'd appreciate any advice on how to track this problem down and fix it.

baudehlo:
What Haraka version?

superman20:
Sorry.  It is Master on June 1 (with the PR's I've submitted since then).

baudehlo:
So the problem is that for some reason next_cb() is being called twice.

The way to track that down would be to look at all the places next_cb() is
called (all in outbound/hmail.js) and try and figure out a code path that
would result in a double-call.

[...]

PR #2018, opened 2017-07-28
Improve logging UUID tracking in Outbound

superman20:
Create logging methods in Outbound similar to Connection so that UUID is reported in log and it is easier to isolate and track outbound e-mails during high concurrency.  

Helps with debugging #1954

### DIFF
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -71,7 +71,20 @@ class HMailItem extends events.EventEmitter {
 
 module.exports = HMailItem;
 
-logger.add_log_methods(HMailItem.prototype, "outbound");
+// copy logger methods into HMailItem:
+for (var key in logger) {
+    if (!/^log\w/.test(key)) continue;
+    HMailItem.prototype[key] = (function (level) {
+        return function () {
+            // pass the HMailItem instance to logger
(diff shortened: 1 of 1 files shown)


baudehlo:
I feel like this PR could do so much more. There are tons of cases in the outbound code where we don't include the object where it should be available. This is an opportunity to fix that.

superman20:
I agree. I'll finally be back home next week where I could (in theory) have more time to complete the concept. This fix hits 80+% of the missing UUID's for me. I would also like to focus my available time on fixing #1954 but it is being elusive. I'm a C++ programmer by profession and JavaScript is more of a tinker language for me. I'm having to spend a lot of time trying to figure out Node's asynchronous behavior/design and the overall structure of Haraka's outbound delivery.  I'm getting the problem every other day now and sometimes multiple times per day

I'll work on completing the logging more as time permits because that is also important to me. I like having an easy/complete trail for each connection and transaction.

baudehlo:
That's fair. And 2.8.14 is so close that we should just merge this.

msimerson:
> we should just merge this.

Done.

and thanks @superman20 !

PR #2020, opened 2017-07-28
remove stun dep to haraka-net-utils

msimerson:
- remove optional dep vs-stun
- require version of haraka-net-utils that includes vs-stun as optional dep
- use windows compatible syntax for npm scripts

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,7 +18,7 @@ before_test:
   - npm --version
 
 test_script:
-  - node run_tests
+  - npm run test
 
 after_test:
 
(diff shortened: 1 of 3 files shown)


smfreegard:
+1 - but this also has the commit for the shutdown too.  That needs to be in a separate pull as it needs some explanation as to why it is needed etc.  I'd also be in favour of removing the graceful shutdown completely or at least re-working it (e.g. close the listeners, remove the cluster handlers so the children can run to completion and exit the master process immediately so that the service manager can start a new Haraka master process up).

baudehlo:
I'll open a separate ticket to discuss graceful shutdown.

PR #2022, opened 2017-07-28
URL fixes and additions to Plugin registry

msimerson:


smfreegard:
Looks fine - but you've got a commit in there that shouldn't be (the process.exit change)

msimerson:
> but you've got a commit in there that shouldn't be (the process.exit change)

Yeah I know, that's just there so the test suite will pass.

PR #2023, opened 2017-07-28
move plugins from deps to optionalDeps

msimerson:
plugins really are optional, and optional deps get installed by default. The only difference is that optional deps won't stop the build process. Since most plugins aren't "required," declare them as optional.

Thoughts?


smfreegard:
LGTM +1

PR #2032, opened 2017-08-05
TLS cert dir

### LINKED ISSUE #1666: honorCipherOrder is false by default
As discussed in #1662 the new `honorCipherOrder` option of the tls plugin defaults to `false`. In fact, without configuring this parameter to `true`, server setups are considered insecure today. So the question is whether Haraka’s default setup should set `honorCipherOrder` to `true`, making it the proper secure default, or `false` as it was in the past and as it is with mostly any other TLS server (apache, nginx, etc.).

Note that in general it wouldn’t be a problem for well-configured clients to dictate the cipher preference. However, most clients and client libraries cannot be configured for a manual cipher order, and quite some of the older clients have some weird ordering. For example, [Android 4.4.2](https://www.ssllabs.com/ssltest/viewClient.html?name=Android&version=4.4.2&key=62) orders weak non-PFS ciphers like `TLS_RSA_WITH_3DES_EDE_CBC_SHA` ahead of PFS-ensured state-of-the-art ciphers like `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256`. For making use of the better ciphers even on these systems, admins _do_ have to enforce the order and let the server pick the best supported client’s cipher on its own.

Note, however, that modern browsers and clients do have significantly improved on their cipher list and order. The list was drastically shortened, and the order has been changed from speed/power optimized to strict security quality. Such hardened clients don’t actually benefit from `honorCipherOrder=true` because the server cannot actually present any better order, just worse.

So the question is: Should we make `honorCipherOrder` default `true` for a more secure SMTP world? Then perhaps we should also default the `ciphers` string?

Note that a secure connection will be established independent from the actual setting of `honorCipherOrder`. However, if the server admin has put a strict order on the `ciphers` setting based on security, the best cipher can be picked. If we make this setting default to `true` and let admins configure arbitrary ordered `ciphers`, an up-to-date client could be downgraded to a worse encryption than the best supported from both client and server. Take, for example, the web page of ryanair.com which, [according to SSL Lab’s analysis](https://www.ssllabs.com/ssltest/analyze.html?d=www.ryanair.com&latest) from today, was configured to enforce the cipher order with a weak 3DES cipher ahead of better AES256 and AES128 ciphers.

### LINKED ISSUE #1686: Server support for SNI
Subject says it all. Quite some clients do already support SNI these days but only a very small number of SMTP servers support it. E.g. neither sendmail nor postfix support SNI on the server side. Let’s Haraka support it.

### LINKED ISSUE #1828: TLS certificate directory
Extracting from #1800 

### TLS cert management

I want a TLS setting like haproxy: `listen 587 ssl cert /etc/certs`, where /etc/certs is a directory filled with .pem files like
- mail.example.com.pem
- mail.example2.com.pem
- ...

Each file is self-contained with the key, cert, and intermediates all concatenated. SNI is going to be a nightmare without this.

This makes SNI & Let's Encrypt (or any other automated provisioning) integration really easy as automated processes can just install the newly provisioned certs into the Haraka config directory.

msimerson:
- Fixes #1828 TLS Certificate Directory
- Fixes #1686 Server support for SNI
- Fixes #1666 honorCipherOrder is false by default
- [x] docs updated.

### before

* start up servers
* load plugins
* parse tls.ini
    * replace tls.cfg.[key|cert|dhparam] file paths with file contents
    * apply [main] settings
* see if TLS plugin loaded & had good config
* apply STARTTLS upgrade but don't announce STARTTLS unless good config
* other [potential] TLS consumers (smtp_client, smtp_forward, outbound, http) duplicated TLS config loading and TLS server setup for themselves.

### after

* server.js
* loads tls.ini && config/tls via `tls_socket.getSocketOpts('*')`
    * saves 'default' config to `tls_socket.certsByHost['*']`
    * all other certs saved at `tls_socket.certsByHost[servername]`
        * [main] TLS settings applied to all certs
    * generates secure contexts at `tls_socket.ctxByHost[servername]`
* start up port 465 server (if configured) & tls config is good
* don't announce STARTTLS unless a working config

A prerequisite for SNI was moving to a store of pre-parsed TLS configurations. Previously the TLS config logic was sprawled across server.js, tls_socket.js, and plugins/tls.js. Each caller had to parse options and pass along what it needed. I've consolidated all the config loading and parsing into tls_socket. The TLS server callers no longer need to pass config options at run time. Clients do, only to provide client-side SNI (servername) to the remote server.

### Significant Changes

1. SNI support. Haraka serves more than one TLS cert
2. To the extent possible, make it "just work."
    * automatically generate dhparams, if missing
    * generate TLS contexts for each TLS cert immediately after config loaded
        * this dramatically pares down the TLS config parsing sprawl
        * setting up TLS servers just got much easier
3. if http.ini specifies port 443, enable TLS with default (`*`) cert & SNI
4. honorCipherOrder defaults to true now. I've been running with this setting for quite some time and have seen no issues.
5. tls.ini ciphers has default list removed. At times specifying the ciphers was a security improvement. Today, commenting them out is a security improvement as it leaves only the secure ciphers enabled by node.js. Secure by default is best default, and the list is present in the file for broader compatibility.

### Aspirations / TODO / Future Plans

- TLS is internet plumbing. Only the plumbers need to know the details. It should be simple for everyone else.
- If a cert and key aren't found, use the hostname from `config/me` and:
    - generate a self-signed cert
    - try requesting a cert from Let's Encrypt.
        - requires port 80 to be listening and a /.well-known/acme-challenge route created
        - an optionalDependency on one of the acme node.js modules
        - test that after a cert in config/tls has been updated, it's served by the SMTP & HTTP servers
- [x] verify that OCSP still works


msimerson:
note to self: to test ocsp, add `-tlsextdebug -status` to `openssl s_client`

msimerson:
This PR is finally ready. I had a "more than once" OCSP response callback issue that surfaced only during STARTTLS and only when SNI was present. The solution was to assure the OCSPRequest listener wasn't added more than once.

To test this out, deploy Haraka from this branch with `npm install -g msimerson/Haraka#tls-cert-dir`, install a couple TLS certificates in `/path/to/config/tls`, and then point openssl at the install like so:

````
openssl s_client -connect mail.theartfarm.com:587 -starttls smtp -tlsextdebug -status -servername mail.theartfarm.com 
openssl s_client -connect mail.theartfarm.com:587 -starttls smtp -tlsextdebug -status -servername mail.simerson.net 
openssl s_client -connect mail.theartfarm.com:465 -tlsextdebug -status
````

You should see an OCSP stapled response with a SSL certificate that matches the hostname you sent. Here's an example:

````
$ openssl s_client -connect mail.theartfarm.com:587 -starttls smtp -tlsextdebug -status -servername mail.simerson.net
CONNECTED(00000003)
TLS server extension "renegotiation info" (id=65281), len=1
0001 - <SPACES/NULS>
TLS server extension "EC point formats" (id=11), len=4
0000 - 03 00 01 02                                       ....
TLS server extension "session ticket" (id=35), len=0
TLS server extension "status request" (id=5), len=0
depth=2 O = Digital Signature Trust Co., CN = DST Root CA X3
verify return:1
depth=1 C = US, O = Let's Encrypt, CN = Let's Encrypt Authority X3
verify return:1
depth=0 CN = mail.simerson.net
verify return:1
OCSP response: 
======================================
OCSP Response Data:
    OCSP Response Status: successful (0x0)
    Response Type: Basic OCSP Response
    Version: 1 (0x0)
    Responder Id: C = US, O = Let's Encrypt, CN = Let's Encrypt Authority X3
    Produced At: Aug 20 22:54:00 2017 GMT
    Responses:
    Certificate ID:
      Hash Algorithm: sha1
      Issuer Name Hash: 7EE66AE7729AB3FCF8A220646C16A12D6071085D
      Issuer Key Hash: A84A6A63047DDDBAE6D139B7A64565EFF3A8ECA1
      Serial Number: 04F212B04D552401298EC1D946E8241CDD51
    Cert Status: good
    This Update: Aug 20 22:00:00 2017 GMT
    Next Update: Aug 27 22:00:00 2017 GMT

    Signature Algorithm: sha256WithRSAEncryption
         2c:92:7d:55:24:03:b6:40:a2:5f:58:96:f4:1c:01:1a:6e:e4:
         <snip>
         c6:3a:80:b2:02:e4:8a:b4:db:24:56:e7:12:c4:af:45:1d:b9:
         90:5b:78:56
======================================
---
Certificate chain
 0 s:/CN=mail.simerson.net
   i:/C=US/O=Let's Encrypt/CN=Let's Encrypt Authority X3
 1 s:/C=US/O=Let's Encrypt/CN=Let's Encrypt Authority X3
   i:/O=Digital Signature Trust Co./CN=DST Root CA X3
---
Server certificate
-----BEGIN CERTIFICATE-----
MIIFBTCCA+2gAwIBAgISBPISsE1VJAEpjsHZRugkHN1RMA0GCSqGSIb3DQEBCwUA
<snip>
6msTz4D4+mIr7443KwXLay2eDEFAW1rW+rLBQCF3yx9PwoHvAuuET+w=
-----END CERTIFICATE-----
subject=/CN=mail.simerson.net
issuer=/C=US/O=Let's Encrypt/CN=Let's Encrypt Authority X3
---
No client certificate CA names sent
Client Certificate Types: RSA sign, DSA sign, ECDSA sign
Requested Signature Algorithms: RSA+SHA512:DSA+SHA512:ECDSA+SHA512:RSA+SHA384:DSA+SHA384:ECDSA+SHA384:RSA+SHA256:DSA+SHA256:ECDSA+SHA256:RSA+SHA224:DSA+SHA224:ECDSA+SHA224:RSA+SHA1:DSA+SHA1:ECDSA+SHA1
Shared Requested Signature Algorithms: RSA+SHA512:DSA+SHA512:ECDSA+SHA512:RSA+SHA384:DSA+SHA384:ECDSA+SHA384:RSA+SHA256:DSA+SHA256:ECDSA+SHA256:RSA+SHA224:DSA+SHA224:ECDSA+SHA224:RSA+SHA1:DSA+SHA1:ECDSA+SHA1
Peer signing digest: SHA512
Server Temp Key: ECDH, P-256, 256 bits
---
SSL handshake has read 3969 bytes and written 516 bytes
---
New, TLSv1/SSLv3, Cipher is ECDHE-RSA-AES256-GCM-SHA384
Server public key is 2048 bit
Secure Renegotiation IS supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : ECDHE-RSA-AES256-GCM-SHA384
    Session-ID: A4A287CD1CB86E160FCEFAEAFFE97BFCF558E66CFF87754EB5B82FE85E3DBA34
    Session-ID-ctx: 
    Master-Key: F318CCF2E0618680A3F2FBA9D76B1C29E44A5E4EBEDF5CF20DE703A93C2EF3B298AC1E056AA0783E11AE18EC30D2EB71
    Key-Arg   : None
    PSK identity: None
    PSK identity hint: None
    SRP username: None
    TLS session ticket lifetime hint: 300 (seconds)
    TLS session ticket:
    0000 - 3c 4e 9e f1 22 7a 4b e6-3c 5c 12 f2 1b 0b 78 a3   <N.."zK.<\....x.
    <snip>
    00a0 - de f8 d2 e5 90 f5 5b 89-ba 0c 73 6a 01 1b 4f 25   ......[...sj..O%

    Start Time: 1503269810
    Timeout   : 300 (sec)
    Verify return code: 0 (ok)
---
250 STARTTLS
EHLO home.simerson.net
quit
250-mail.theartfarm.com Hello 71-212-95-233.tukw.qwest.net [71.212.95.233], Haraka is at your service.
250-PIPELINING
250-8BITMIME
250-SMTPUTF8
250-SIZE 45000000
250 AUTH PLAIN LOGIN CRAM-MD5
221 mail.theartfarm.com closing connection. Have a jolly good day.
closed
````

smfreegard:
As you committed this without waiting for a review - I've just had the first bug report on IRC that having config/tls_key.pem and config/tls_cert.pem do not work any more.  I've told them to raise an issue.

msimerson:
If that's the case, it's a bug. They do still work and the report I saw on IRC was for 2.8.14, not HEAD. Loading `tls_*.pem` loads the first and default TLS certificate, and the TLS certs in config/tls/* are then uses for SNI matches.

smfreegard:
No - I confirmed that it was definitely broken in HEAD and installing 2.8.14 made it work straightaway.

msimerson:
I'll test now with only config/tls*.pem files (I've been testing with both). The new TLS code does more now, like validating the certs at startup time (via openssl calls) and generating the secure contexts in advance.

msimerson:
I have just tested by deploying Haraka from HEAD (npm install haraka/Haraka) with only a single self-signed TLS certificate configured in the traditional way. There is no `config/tls` directory present (the default).

````
# swaks -to postmaster@vmware.imac27.simerson.net -server 172.16.15.9 -protocol esmtps
=== Trying 172.16.15.9:25...
=== Connected to 172.16.15.9.
<-  220 haraka ESMTP Haraka 2.8.14 ready
 -> EHLO monitor.simerson.net
<-  250-haraka Hello stage [172.16.15.254], Haraka is at your service.
<-  250-PIPELINING
<-  250-8BITMIME
<-  250-SMTPUTF8
<-  250-SIZE 0
<-  250 STARTTLS
 -> STARTTLS
<-  220 Go ahead.
=== TLS started with cipher TLSv1.2:ECDHE-RSA-AES128-GCM-SHA256:128
=== TLS no local certificate set
=== TLS peer DN="/C=US/ST=Washington/L=Seattle/O=Mail Toaster/CN=vmware.imac27.simerson.net"
 ~> EHLO monitor.simerson.net
<~  250-haraka Hello stage [172.16.15.254], Haraka is at your service.
<~  250-PIPELINING
<~  250-8BITMIME
<~  250-SMTPUTF8
<~  250-SIZE 0
<~  250 AUTH PLAIN LOGIN
 ~> MAIL FROM:<root@monitor.simerson.net>
<~  250 sender <root@monitor.simerson.net> OK
 ~> RCPT TO:<postmaster@vmware.imac27.simerson.net>
<~  250 recipient <postmaster@vmware.imac27.simerson.net> OK
 ~> DATA
<~  354 go ahead, make my day
 ~> Date: Tue, 29 Aug 2017 21:35:38 -0700
 ~> To: postmaster@vmware.imac27.simerson.net
 ~> From: root@monitor.simerson.net
 ~> Subject: test Tue, 29 Aug 2017 21:35:38 -0700
 ~> Message-Id: <20170829213538.005141@monitor.simerson.net>
 ~> X-Mailer: swaks v20170101.0 jetmore.org/john/code/swaks/
 ~> 
 ~> This is a test mailing
 ~> 
 ~> .
<~  250 ok 1504067738 qp 5145 (CDB425B0-6E71-47F4-B773-46979CB6A309.1)
 ~> QUIT
<~  221 haraka closing connection. Have a jolly good day.
````

I looked back to see details and found:

> brucegg: guys i need to ask question - i got [TLS]no valid config when starting haraka - i copied config with keys/cert in right order
> brucegg: once again i wanna to ask about tls config - i got error (No Valid TLS config) - i see that lines added in tls.js 6 days ago. I create bundle with key/cert/cat in configt/tls/domain_name.pem. Still got 'no valid tls config'. Where now should be places all tls confs?

Without anything else to go on, it appears brucegg was attempting to configure *only* config/tls/domain_name.pem. Having not populated the config/tls_*.pem files is a likely way to get the `No Valid TLS config` error. That error is being returned by nodes tls module, likely because we attempted to generate a secure context without a cert and key.

It appears the bug here is in the documentation. The docs say, "If no match is found, the default certificate (see Certificate Files above) is presented."  I don't explicitly say, "even when using the cert dir, you must still configure key and cert to point to files containing just the key and cert." 

An alternative solution is to not require key/cert when config/tls is populated, but then we're stuck choosing which certificate is the default when no TLS servername is presented. Some servers just sort the cert filenames alphabetically and make the first one the default. Since we already had the cert/key file mechanism, that seems like a reasonable way to explicitly specify the default.

baudehlo:
+1 to just picking the first from a list but can you also fs.symlink it (this is cross platform in Node you just have to give full paths for it to work in windows). That way next run it won't have to search but more importantly it's a mechanism to change the default, for free.

smfreegard:
Ok - I've just tested it myself and it doesn't work for me.  

See https://gist.github.com/smfreegard/2fc9e38655a65b783c03e34d1c7a56ec for the full session logs of how I installed Haraka, generated some self-signed certificates and then tested it with openssl s_client and it reported to me exactly the same problem experienced by brucegg on IRC.

smfreegard:
I've tried debugging this myself - the problem appears that no configuration is loaded as I never see `loading tls.ini` in the log which should be emitted by `load_tls_ini` in `tls_socket.js`.

Confusingly there is a load_tls_ini in both tls_socket and in the haraka-net-utils package.

In server.js there is Server.load_default_tls_config, but it's only ever called by the tests.  It would appear that `tls_socket.getSocketOpts` calls `tls_socket.load_tls_ini`.

smfreegard:
Ok - I got it working with this:

`````
index b7cebdb..d5a0eff 100644
--- a/server.js
+++ b/server.js
@@ -361,9 +361,11 @@ Server.get_smtp_server = function (host, port, inactivity_timeout, done) {
     }
     else {
-        Server.listeners.push(server);
-        done(server);
-    }
+        Server.load_default_tls_config(function (cb) {
+            Server.listeners.push(server);
+            done(server);
+        });
+    } 
 };
`````

However Haraka starts up and starts answering connections and TLS works (verified by s_client), but I noticed my CPU kicked into overdrive and I found `openssl` hammering my CPU and upon investigation I found that Haraka was generating a dhparams file which had taken a few minutes, when it completed - it crashed Haraka with:

`````
[INFO] [-] [core] Saved to /home/smf/test/Haraka/config/dhparams.pem
[CRIT] [-] [core] TypeError: Path must be a string. Received undefined
[CRIT] [-] [core]     at assertPath (path.js:7:11)
[CRIT] [-] [core]     at Object.resolve (path.js:1146:7)
[CRIT] [-] [core]     at Config.get (/home/smf/test/Haraka/node_modules/haraka-config/config.js:28:26)
[CRIT] [-] [core]     at ChildProcess.o.on.code (/home/smf/test/Haraka/tls_socket.js:530:35)
[CRIT] [-] [core]     at emitTwo (events.js:106:13)
[CRIT] [-] [core]     at ChildProcess.emit (events.js:194:7)
[CRIT] [-] [core]     at maybeClose (internal/child_process.js:899:16)
[CRIT] [-] [core]     at Socket.<anonymous> (internal/child_process.js:342:11)
[CRIT] [-] [core]     at emitOne (events.js:96:13)
[CRIT] [-] [core]     at Socket.emit (events.js:191:7)
`````

msimerson:
- net_utils.load_tls_ini has to remain until after the next Haraka release. I've created haraka/haraka-net-utils#23 to track that. 

- The server.js changes are not necessary. Instead, update plugins/tls to load tls.ini via the tls_socket function. PR is landing momentarily.

PR #2034, opened 2017-08-14
Fixed swaks test command

dtruel:
The swaks test command in the tutorial for outbound email has the to and from addresses reversed.  I'm creating this pull request so nobody else wonders why their server is timing out trying to reach example.com's ip

Checklist:
- [x] docs updated

### DIFF
--- a/docs/tutorials/SettingUpOutbound.md
+++ b/docs/tutorials/SettingUpOutbound.md
@@ -60,11 +60,11 @@ Now you can start Haraka. That's all the configuration you need.
     haraka -c .
 
 Now in another window you can run swaks to test this - be sure to substitute
-an email address you can monitor in place of test@example.com, and the
-username and password you added for the -au and -ap params:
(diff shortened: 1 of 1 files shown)


msimerson:
Actually, both the -to and -from should be changed to real values. Maybe we should make that more obvious with something like this?

    swaks -to me@example.com -from me@example.com -s localhost \

dtruel:
I agree, I think your solution is definitely more readable which is exactly what you want in a tutorial.  Aren't you supposed to do two dashed (`--to`) instead of one? I know it works with one but the Swaks manpages always use two dashes instead of one.

msimerson:
Sounds great. Please update this PR accordingly and I'll merge it.

msimerson:
thanks @dtruel

dtruel:
Yup. Thanks for all your hard work maintaining this amazing repo.

PR #2036, opened 2017-08-19
Fix for node8, use forked daemon package

Infern1:
Fixes #

- Use forked daemon package, see here https://github.com/silverwind/droppy/issues/324

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "address-rfc2821"       : "^1.1.1",
     "address-rfc2822"       : "^2.0.0",
     "async"                 : "~2.5.0",
-    "daemon"                : "https://github.com/zipang/daemon.node/tarball/48d0977c26fb3a6a44ae99aae3471b9d5a761085",
+    "daemon"                : "~1.1.0",
     "generic-pool"          : "~2.5.0",
     "iconv"                 : "~2.3.0",
     "ipaddr.js"             : "~1.4.0",
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "address-rfc2821"       : "^1.1.1",
     "address-rfc2822"       : "^2.0.0",
     "async"                 : "~2.5.0",
-    "daemon"                : "~1.1.0",
+    "daemon"                : "https://github.com/zipang/daemon.node/tarball/48d0977c26fb3a6a44ae99aae3471b9d5a761085",
     "generic-pool"          : "~2.5.0",
     "iconv"                 : "~2.3.0",
     "ipaddr.js"             : "~1.4.0",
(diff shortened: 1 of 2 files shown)


baudehlo:
I'd rather we fix this by specifying a cwd instead.

Infern1:
> I'd rather we fix this by specifying a cwd instead.
As you wish, same method as here: https://github.com/NodeBB/NodeBB/commit/c3a34bf81a8c2c4519612c826c0ce25a0ac86fd3

baudehlo:
Looks good, but has a lint error (trailing comma).

gramakri:
This patch seems to break the tests. Initially, I thought that was just node 6, but it doesn't work for me on on node 8 either.

```
~/tmp/Haraka$ node --version
v8.4.0
~/tmp/Haraka$ npm test

> Haraka@2.8.14 test /home/girish/tmp/Haraka
> node run_tests

loaded TLD files: 1=1534 2=6132 3=2170
loaded 8229 Public Suffixes
~/tmp/Haraka$ git reset --hard HEAD^
HEAD is now at 35929c2 chore(package): update redis to version 2.8.0 (#2033)
~/tmp/Haraka$ npm test

> Haraka@2.8.14 test /home/girish/tmp/Haraka
> node run_tests

loaded TLD files: 1=1534 2=6132 3=2170
loaded 8229 Public Suffixes

config
✔ arrange_args - name
✔ arrange_args - name, type
✔ arrange_args - name, callback
✔ arrange_args - name, callback, options
...
```

msimerson:
Yeah, I added a [fix](https://github.com/haraka/Haraka/pull/2032/commits/de14201fe90c59fd5a28f69bc7d6e3a61ac256c7#diff-78c12f5adc1848d13b1c6f07055d996eR11) for that in my big TLS pr.

msimerson:
> it doesn't work for me on on node 8 either.

Now that's curious. I saw the failure on my desktop with node 6 and assumed it was a 6 -vs- 8 thing. I think I'll revert this PR until we can figure out why that change breaks the test suite.

msimerson:
Unfortunately, even after it's been reverted, there's no way to re-open this PR. We need to address the daemon issue and the method in this PR breaks the test suite (for unknown reasons).

msimerson:
Another possible solution: the daemon module is just 2 functions. We could import them with a fix for cwd. The `daemon` module is unmaintained for years so there's little chance of getting an upstream fix.

baudehlo:
Hey is there any chance that Node has since included this behaviour? If not, then let's either find a maintained solution or import the code and fix it.

msimerson:
If Node had included it, it's not obvious to me which module they stuck it into. Honestly, it seems like something that should __not__ be in node.js. If anything, discarding that functionality entirely and using OS supplied utilities like `daemon` (included on *BSD, Mac OS X, and most Linux distros) or `systemd` or `launchd` seems like a really good default. Beyond the OS utilities for daemons, npm also offers pm2, forever, daemonize, and more.

PR #2047, opened 2017-08-25
Add logfmt support

KingNoosh:
-  Add support for the logfmt format (key=val)

Checklist:
- [x] docs updated
- [x] tests updated


KingNoosh:
The logfmt format is also supported and can be enabled by changing
`LOGFMTDEFAULT` to `LOGFMTLOGFMT` in the `config/logformat` file which will
start logging in the following format below.

    level=PROTOCOL connection_uuid=9FF7F70E-5D57-435A-AAD9-EA069B6159D9.1 source=core message="S: 354 go ahead, make my day"

Any objects you pass through will also be appeneded to the log line as
key=value and will look like this:

    level=NOTICE connection_uuid=9FF7F70E-5D57-435A-AAD9-EA069B6159D9.1 source=core message=disconnect ip=127.0.0.1 rdns=Unknown helo=3h2dnz8a0if relay=N early=N esmtp=N tls=N pipe=N errors=0 txns=1 rcpts=1/0/0 msgs=1/0/0 bytes=222 lr="" time=0.052

You can find out more about logfmt here: [https://brandur.org/logfmt](https://brandur.org/logfmt)

connection.js
```diff
@@ -23,6 +22,7 @@ var trans       = require('./transaction');
 var plugins     = require('./plugins');
 var rfc1869     = require('./rfc1869');
 var outbound    = require('./outbound');
+var ResultStore = require('haraka-results');
```
msimerson: Just curious, why move this down? The general format you'll find in most files is a section of node.js builtin modules, then a section of NPM modules, and finally a section of local (./) modules.

KingNoosh: It was a mistake, I based this off the older one and reverted a lot of stuff, this should've been one of them, fixing now

docs/CoreConfig.md
```diff
@@ -3,36 +3,33 @@ Core Configuration Files
...
+* log.ini
 
-  Can contain either a number or a string. See the top of logger.js for the
-different levels available.
+Contains settings for log level, timestamps, and format. See the example log.ini file for examples.
+
+* logformat
+
+  Can contain either be LOGFMTDEFAULT for the default log format or
+LOGFMTLOGFMT for the logfmt format.
```
smfreegard: I thought this was changes to LOGFMT and DEFAULT?

logger.js
```diff
@@ -1,14 +1,16 @@
 'use strict';
 // Log class
 
+var util      = require('util');
+var tty       = require('tty');
+
 var constants = require('haraka-constants');
+var logfmt    = require('logfmt');
```
smfreegard: Couldn't we lazy load this?  Seems silly to require it for everyone whether they're using LOGFMT mode or not.

package.json
```diff
@@ -23,6 +23,7 @@
     "iconv"                 : "~2.3.0",
     "ipaddr.js"             : "~1.5.0",
     "js-yaml"               : "~3.9.0",
+    "logfmt"                : "^1.2.0",
```
smfreegard: Shouldn't this be an optional dependency?

docs/CoreConfig.md
```diff
@@ -3,36 +3,33 @@ Core Configuration Files
 
 The Haraka core reads some configuration files to determine a few actions:
 
-* loglevel
```
smfreegard: Why are we removing logevel from the docs - it's still valid.  It's also not being deprecated.

msimerson: We are deprecating it, in favor of `[main]level=` in log.ini.

baudehlo:
It shouldn't be part of this PR though. Deprecating it requires a separate
discussion elsewhere.

msimerson:
I reopened #2048 as #2054, which was just the switch to .ini in a separate PR. I PR'ed that against @KingNoosh branch to avoid merge conflicts. The old log* files still work but they are deprecated in favor of log.ini.

docs/CoreConfig.md
smfreegard: I don't believe the consensus on the IRC channel was to deprecate this.

It's *far* more useful to be able to do:   `````echo LOGDEBUG > /etc/haraka/config/loglevel```` than to have to edit an .ini file.

msimerson: You can still do that. The changes in the PR permit **both** methods of configuration. If the loglevel file exists, it gets priority. During runtime, the last file to get edited takes precedence.

msimerson:
I have removed the log.ini portions from this PR, having moved them over to #2054 where they have continued.

msimerson:
@KingNoosh , are you familiar with the git witchcraft known as rebasing?  If so, please do the needful. If not, catch me on IRC, or just use `git merge`.

logger.js
KingNoosh: We use this for the old key value comments too so all the old key=value stuff. The implementation is on like 275.

docs/CoreConfig.md
KingNoosh: It was, I need to remove this since log.ini was introduced

package.json
KingNoosh: Nope, this is being used for default too, line 275 of logger.js

smfreegard:
Ok - you don't appear to have benchmarked this like I suggested on IRC.  I also didn't appreciate that you had made logfmt integral to logger.

Adding new dependencies into core shouldn't be done without very good reason.  Here is why:

I've just done some really basic benchmarks which do a console.log of a single log line 1 million times.  I picked the disconnect log line as it's the worst case as it puts everything inside an Array and then joins the array by ' ', so this is the worst example.  It's also quite a long log line which was the other reason I picked it.

This is the original disconnect log line with the original code:

`````
smf@i7desktop:~$ time node haraka_logger1.js 2>&1 > /dev/null

real	0m4.343s
user	0m4.148s
sys	0m0.244s
`````

Here is the same but using logfmt.stringify:

`````
smf@i7desktop:~$ time node haraka_logger2.js 2>&1 > /dev/null

real	0m11.633s
user	0m11.448s
sys	0m0.252s
`````

That's **167.86% slower**....

So, there's no way I'm going to approve this pull request with logfmt being used by default in `logger.js`.   It should be lazy loaded if it's enabled and you should probably document the performance hit over standard logging as this seems pretty considerable to me.

For completeness I also changed the code to use es6 literals:

`````
smf@i7desktop:~$ time node haraka_logger3.js 2>&1 > /dev/null

real	0m3.589s
user	0m3.472s
sys	0m0.168s
`````

So that would be **17.36% faster**

And I did the same by changing to the code to + ' ' + concatenate everything without using an Array:

`````
smf@i7desktop:~$ time node haraka_logger4.js 2>&1 > /dev/null

real	0m5.060s
user	0m4.864s
sys	0m0.228s
`````

Whist surprised me a bit as it's **16.51% slower** than doing `array.concat(' ' )`.

So es6 literals are the fastest way to handle this.  However I suggest you simply revert the changes to all of the log lines as changing log lines to es6 literals should be done in a separate pull request.

**Comment by KingNoosh** [comment]
With the changes suggested from IRC, this should be ready to review again.

File: docs/Logging.md [hunk 6]
```diff
@@ -49,9 +49,24 @@ and similarly for the log methods on the plugin object:
...
 failing to provide a connection and/or plugin object will leave
-the default values in the log (currently "core" and
-"no\_connection").
+the default values in the log (currently "core").
 
-this is implemented by testing for argument type in
+This is implemented by testing for argument type in
 the logger.js log\* method. objects-as-arguments are then sniffed
 to try to determine if they're a connection or plugin instance.
+<<<<<<< HEAD:docs/Logging_API.md
```
smfreegard: You've got a merge conflict marker here that you've forgotten to remove

docs/Logging.md
```diff
@@ -49,9 +49,24 @@ and similarly for the log methods on the plugin object:
...
+    level=PROTOCOL connection_uuid=9FF7F70E-5D57-435A-AAD9-EA069B6159D9.1 source=core message="S: 354 go ahead, make my day"
+
+Any objects you pass through will also be appeneded to the log line as
+key=value and will look like this:
+
+    level=NOTICE connection_uuid=9FF7F70E-5D57-435A-AAD9-EA069B6159D9.1 source=core message=disconnect ip=127.0.0.1 rdns=Unknown helo=3h2dnz8a0if relay=N early=N esmtp=N tls=N pipe=N errors=0 txns=1 rcpts=1/0/0 msgs=1/0/0 bytes=222 lr="" time=0.052
+
+You can find out more about logfmt here: [https://brandur.org/logfmt](https://brandur.org/logfmt)
+=======
+>>>>>>> upstream/master:docs/Logging.md
```
smfreegard: And another two here.

docs/Logging.md
```diff
@@ -49,9 +49,24 @@ and similarly for the log methods on the plugin object:
...
+This is implemented by testing for argument type in
 the logger.js log\* method. objects-as-arguments are then sniffed
 to try to determine if they're a connection or plugin instance.
+<<<<<<< HEAD:docs/Logging_API.md
+
+The logfmt format is also supported and can be enabled by changing the format
+from `default` to `logfmt` in the `config/log.ini` file which will
+start logging in the following format below.
+
+    level=PROTOCOL connection_uuid=9FF7F70E-5D57-435A-AAD9-EA069B6159D9.1 source=core message="S: 354 go ahead, make my day"
```
smfreegard: I thought `connection_uuid` was changed to `uuid` ?

logger.js
```diff
@@ -208,44 +252,75 @@ logger._init();
 logger.log_if_level = function (level, key, plugin) {
     return function () {
         if (logger.loglevel < logger[key]) { return; }
-        var levelstr = '[' + level + ']';
-        var str = '';
-        var uuidstr = '[-]';
-        var pluginstr = '[' + (plugin || 'core') + ']';
+        var logobj = {
+            level,
+            connection_uuid: '-',
```
smfreegard: Ok - it appears not.  This isn't going to be a connection_uuid, it could be a transaction ID or an outbound recipient ID.  So `uuid` is far more appropriate (and terse).

plugins.js
```diff
@@ -532,15 +532,15 @@ function log_run_item (item, hook, retval, object, params, msg) {
...
-        object[log]([
-            'hook='     + hook,
-            'plugin='   + item[0].name,
-            'function=' + item[1],
-            'params="'  + ((params) ? ((typeof params === 'string') ?
-                params : params[0]) : '') + '"',
-            'retval='   + constants.translate(retval),
-            'msg="'     + ((msg) ? msg : '') + '"',
-        ].join(' '));
+        object[log]({
```
smfreegard: Can you align this block like it was before.

docs/Logging.md
KingNoosh: It's still `connection_uuid`, do you want me to rename to `uuid`?

KingNoosh: nvm

docs/Logging.md
```diff
@@ -21,25 +21,25 @@ See also
 ------------------
 [https://github.com/haraka/Haraka/pull/119](https://github.com/haraka/Haraka/pull/119)
 
-logline will always always be in the form:
+The logline by default will be in the form of:
 
     [level] [connection uuid] [origin] message
```
smfreegard: Ooops - missed one.

PR #2048, opened 2017-08-25
move log settings into log.ini

msimerson:
- move loglevel and logtimestamp settings into log.ini
- with backwards compat for loglevel and timestamps files.
- removed obsolete log.syslog.ini file

Checklist:
- [x] docs updated
- [x] tests updated


msimerson:
subsumed into #2047

bin/haraka
```diff
@@ -223,6 +223,7 @@ function setupHostname (confPath) {
 function setupBaseConfig (confPath) {
     copyFile(path.join(base, 'config', 'smtp.ini'), path.join(confPath, 'smtp.ini'));
     copyFile(path.join(base, 'config', 'loglevel'), path.join(confPath, 'loglevel'));
+    copyFile(path.join(base, 'config', 'logformat'), path.join(confPath, 'logformat'));
```
msimerson: this added line can be removed and the line above it (loglevel) should be replaced with log.ini.

logger.js
```diff
@@ -1,14 +1,16 @@
 'use strict';
 // Log class
 
+var util      = require('util');
+var tty       = require('tty');
+
 var constants = require('haraka-constants');
+var logfmt = require('logfmt');
```
msimerson: In the interests of keeping things lean, can we move this require into the `_init_format` function, and only load it when that log format has been specified?

KingNoosh: We use this in `default` to keep support for the older key=val stuff from back when they were in an array or concatenated.

msimerson: Ah, okay.

PR #2054, opened 2017-08-28
Log ini

msimerson:
This is the same PR as #2048 (closed).

- move loglevel and logtimestamp settings into log.ini
- with backwards compat for loglevel and timestamps files.
- removed obsolete log.syslog.ini file

Checklist:
- [x] docs updated
- [x] tests updated


msimerson:
While the config file `loglevel` still works, I removed it from the docs because it's much simpler to document a the INI file. Because log.ini will exist by default, and because it has the possible values enumerated in the file, it's much easier to discover and twiddle. The people that are accustomed to `echo $LEVEL > loglevel' don't need the docs.

smfreegard:
I know it still works - I reviewed the code.  I just don't see the point of removing if from the docs so new users can easily find the shortcut way of doing it.   

Deprecated means it will be removed in future and in this case it won't be.

msimerson:
If we call it "undocumented" instead of "deprecated", does that satisfy you? 

I almost never change the value of loglevel so this doesn't mean much to me. The only horse I have in this race is making it easier for users of Haraka.

baudehlo:
I'd say if you're going to remove it in that one location of the docs it
probably requires changed in a few Tutorial type places too. But I'm also
against removing it from the docs, but instead saying "This is a shorthand
version of changing log.ini, and can be useful on the command line via
`echo LOGPROTOCOL > config/loglevel` to instantly change the current log
levels".

msimerson:
I already updated all instances of loglevel in this repo (with a sed command), but I'll restore docs for the `loglevel` file.

config/log.ini
```diff
@@ -0,0 +1,11 @@
+[main]
+
+; level=DATA, PROTOCOL, DEBUG, INFO, NOTICE, WARN, ERROR, CRIT, ALERT, EMERG
```
baudehlo: if the level is in lower case, the examples should be too.

msimerson: Agreed.

msimerson: done.

config/log.syslog.ini
```diff
@@ -1,9 +0,0 @@
-[general]
```
baudehlo: Removing this file seems incorrect.

msimerson: It's an anacronism, the new syslog plugin uses syslog.ini.

baudehlo: Gotcha

PR #2066, opened 2017-08-31
This just cleans up the code, probably no fix

baudehlo:
WIP

Fixes #

Changes proposed in this pull request:
- 
- 

Checklist:
- [ ] docs updated
- [ ] tests updated
- [ ] [Changes](https://github.com/haraka/Haraka/blob/master/Changes.md) updated

PR #2073, opened 2017-09-02
es6: var -> const or let

### LINKED ISSUE #2065: ES6: replace most uses of var with let or const.
Rather than polluting hundreds of PRs over many months to get this done, this should be done in its own PR. Soon.

* [x] add to eslintrc.json
* [ ] add to [haraka-eslint](https://github.com/haraka/haraka-eslint) so it trickles down
    * [ ] in warning mode only. It can't get switched to error until updates have been made everywhere

msimerson:
Fixes #2065

- replace uses of var with const or let

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/bin/haraka
+++ b/bin/haraka
@@ -4,16 +4,17 @@
 // https://github.com/visionmedia/express/blob/master/bin/express
 // https://github.com/tnantoka/LooseLeaf/blob/master/bin/looseleaf
 
-var fs   = require('fs');
-var path = require('path');
-var nopt = require('nopt');
-var os   = require('os');
+const fs   = require('fs');
+const net  = require('net');
+const nopt = require('nopt');
+const path = require('path');
+const os   = require('os');
 
-var utils   = require('haraka-utils');
-var sprintf = require('sprintf-js').sprintf;
-var base = path.join(__dirname, '..');
-var ver  = JSON.parse(fs.readFileSync(base+'/package.json', 'utf8')).version;
-var knownOpts = {
+const utils   = require('haraka-utils');
+const sprintf = require('sprintf-js').sprintf;
+const base = path.join(__dirname, '..');
+const ver  = JSON.parse(fs.readFileSync(base+'/package.json', 'utf8')).version;
+const knownOpts = {
     "version":     Boolean,
     "verbose":     Boolean,
     "help":        [String, null],
@@ -40,7 +41,7 @@ var knownOpts = {
     "skip-deny":   Boolean,
     "set-relay":   Boolean,
 };
-var shortHands = {
+const shortHands = {
     "v": ["--version"],
     "h": ["--help"],
     "c": ["--configs"],
@@ -51,9 +52,9 @@ var shortHands = {
     "o": ["--order"],
     "t": ["--test"],
 };
-var parsed = nopt(knownOpts, shortHands, process.argv, 2);
+const parsed = nopt(knownOpts, shortHands, process.argv, 2);
 
-var usage = [
+const usage = [
     "\033[32;40mHaraka.js\033[0m — A Node.js Email Server project",
     "Usage: haraka [options] [path]",
     "Options:",
@@ -84,17 +85,17 @@ var usage = [
 ].join('\n');
 
 
-var listPlugins = function (b, dir) {
+function listPlugins (b, dir) {
 
     if (!dir) { dir = "plugins/"; }
 
-    var plist = dir + "\n";
-    var subdirs = [];
-    var gl = path.join((b ? b : base), dir);
-    var pd = fs.readdirSync(gl);
+    let plist = dir + "\n";
+    const subdirs = [];
+    const gl = path.join((b ? b : base), dir);
+    const pd = fs.readdirSync(gl);
 
     pd.forEach(function (p) {
-        var stat = fs.statSync(gl + '/' + p);
+        const stat = fs.statSync(gl + '/' + p);
         if (stat.isFile() && ~p.search('.js')) {
             plist += "\t" + p.replace('.js', '') + "\n";
         }
@@ -153,19 +154,19 @@ function mkDir (dstPath) {
 function copyDir (srcPath, dstPath) {
 
     mkDir(dstPath);
-    var files = fs.readdirSync(srcPath);
+    const files = fs.readdirSync(srcPath);
 
-    for (var i = 0; i < files.length; i++) {
+    for (let i = 0; i < files.length; i++) {
 
         // Ignore ".*"
         if (/^\./.test(files[i])) {
             continue;
         }
 
-        var srcFile = path.join(srcPath, files[i]);
-        var dstFile = path.join(dstPath, files[i]);
+        const srcFile = path.join(srcPath, files[i]);
+        const dstFile = path.join(dstPath, files[i]);
 
-        var srcStat = fs.statSync(srcFile);
+        const srcStat = fs.statSync(srcFile);
 
         // Recursive call If direcotory
         if (srcStat.isDirectory()) {
@@ -204,15 +205,15 @@ function copyFile (srcFile, dstFile) {
 }
 
 function setupHostname (confPath) {
-    var hlPath = path.join(confPath, 'host_list');
+    const hlPath = path.join(confPath, 'host_list');
     try {
         if (fs.statSync(hlPath).isFile()) return;
     }
     catch (ignore) {}
 
-    var hostname = os.hostname() + "\n";
+    const hostname = os.hostname() + "\n";
     try {
-        var fd = fs.openSync(hlPath, 'w');
+        const fd = fs.openSync(hlPath, 'w');
         fs.writeSync(fd, hostname, null);
     }
     catch (e) {
@@ -227,7 +228,7 @@ function setupBaseConfig (confPath) {
     copyDir(path.join(base, 'config', 'dkim'), path.join(confPath, 'dkim'));
 }
 
-var readme = [
+const readme = [
     "Haraka",
     "",
     "Congratulations on creating a new installation of Haraka.",
@@ -253,7 +254,7 @@ var readme = [
     "",
 ].join("\n");
 
-var packageJson = JSON.stringify({
+const packageJson = JSON.stringify({
     "name": "haraka_local",
     "description": "An SMTP Server",
     "version": "0.0.1",
@@ -262,7 +263,7 @@ var packageJson = JSON.stringify({
     "license": "MIT"
 });
 
-var plugin_src = [
+const plugin_src = [
     "// %plugin%",
     "",
     "// documentation via: haraka -c %config% -h plugins/%plugin%",
@@ -272,7 +273,7 @@ var plugin_src = [
     "",
 ].join("\n");
 
-var plugin_doc = [
+const plugin_doc = [
     "%plugin%",
     "========",
     "",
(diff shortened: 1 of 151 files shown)


endreszabo:
wow!

msimerson:
> wow!

😁 

It wasn't as bad as I thought. The first couple commits were `eslint --fix`, which caught the majority of cases. I didn't see a single error made by `--fix`. What it did do is leave behind quite a few manual changes, but they were straight forward: either hoist the const/let or add block scope to a switch/case with `{}`.

baudehlo:
So what I would like to do, if you agree, is push a quick point release for the outbound fixes before merging this. Thoughts?

msimerson:
> push a quick point release for the outbound fixes

Not a big fan, for a couple reasons: quick releases tend to be followed up by quick releases and because a lot of TLS code recently landed. I'd rather get the outstanding PRs (including this) merged and then start a quiet period. We can point folks with outbound issues to test with `npm install -g haraka/Haraka#05ab2d5` to exclude this (and subsequent commits). Meanwhile, the brave and intrepid HEAD runners will get time to deploy and shake out any bugs in HEAD during the quiet period.

msimerson:
and #2066 would likely get concluded during the quiet period...

msimerson:
>  if you agree, is push a quick point release for the outbound fixes

Also, to be clear, I'm not opposed. If someone were pushing the release, I would not object.

host_pool.js
```diff
@@ -56,26 +56,26 @@ function HostPool (hostports_str, retry_secs) {
...
-    var self = this;
-    var key = host + ':' + port;
-    var retry_msecs = self.retry_secs * 1000;
+    const self = this;
+    const key = host + ':' + port;
+    const retry_msecs = self.retry_secs * 1000;
     self.dead_hosts[key] = true;
 
-    var cb_if_still_dead = function (){
+    function cb_if_still_dead () {
```
KingNoosh: Keep declarations consistent

host_pool.js
```diff
@@ -56,26 +56,26 @@ function HostPool (hostports_str, retry_secs) {
...
         self.dead_hosts[key] = true;
         // console.log(1);
         setTimeout(function () {
             self.probe_dead_host(host, port, cb_if_still_dead, cb_if_alive);
         }, retry_msecs);
-    };
+    }
 
-    var cb_if_alive = function (){
+    function cb_if_alive () {
```
KingNoosh: Keep declarations consistent

logger.js
```diff
@@ -1,22 +1,23 @@
...
 
-var constants = require('haraka-constants');
+const constants = require('haraka-constants');
 
-var config    = require('./config');
-var plugins;
-var connection;
-var outbound;
+const config    = require('./config');
+/* eslint-disable prefer-const */
```
KingNoosh: Why the disable?

logger.js
```diff
@@ -1,22 +1,23 @@
...
+let plugins;
+let connection;
+let outbound;
 
 const regex = /(^$|[ ="\\])/;
 const escape_replace_regex = /["\\]/g;
 
 function stringify (obj) {
```
KingNoosh: More my fault but would it be worth updating this to `const stringify`?

plugins.js
```diff
@@ -427,13 +427,15 @@ plugins.run_next_hook = function (hook, object, params) {
...
-    var item;
-    var callback = function (retval, msg) {
+    let called_once = false;
+    let timeout_id;
+    let timed_out = false;
+    let cancelled = false;
+    /* eslint-disable prefer-const */
+    let item;
+
+    function cancel () { cancelled = true; }
```
KingNoosh: Keep declarations consistent

plugins.js
```diff
@@ -427,13 +427,15 @@ plugins.run_next_hook = function (hook, object, params) {
...
+    function callback (retval, msg) {
```
KingNoosh: Keep declarations consistent

plugins.js
```diff
@@ -427,13 +427,15 @@ plugins.run_next_hook = function (hook, object, params) {
...
-    var timed_out = false;
-    var cancelled = false;
-    var cancel = function () { cancelled = true; };
-    var item;
-    var callback = function (retval, msg) {
+    let called_once = false;
+    let timeout_id;
+    let timed_out = false;
+    let cancelled = false;
+    /* eslint-disable prefer-const */
```
KingNoosh: If we're going to disable eslint stuff, wouldn't it be better if we at least put this at the top of the file?

plugins/attachment.js
```diff
@@ -101,13 +101,14 @@ exports.unarchive_recursive = function (connection, f, archive_file_name, cb) {
...
-    var count = 0;
-    var done_cb = false;
-    var timer;
+    const plugin = this;
+    const files = [];
+    const tmpfiles = [];
+    const depth_exceeded = false;
+    let count = 0;
+    let done_cb = false;
+    /* eslint-disable prefer-const */
```
KingNoosh: Put at top of file?

plugins/avg.js
```diff
@@ -119,9 +119,8 @@ exports.hook_data_post = function (next, connection) {
                         socket.send_command('SCAN', tmpfile);
                     }
                     break;
-                case 'scan':
-                    var end_time = Date.now();
-                    var elapsed = end_time - start_time;
+                case 'scan': {
```
KingNoosh: Weird braces

plugins/bounce.js
```diff
@@ -337,17 +337,12 @@ exports.bounce_spf = function (next, connection) {
...
+    let timer;
 
-    var timer = setTimeout(function () {
-        connection.logerror(plugin, 'Timed out');
-        txn.results.add(plugin, { skip: 'bounce_spf(timeout)' });
-        return run_cb(true);
-    }, (plugin.timeout - 1) * 1000);
-
-    var run_cb = function (abort, retval, msg) {
+    function run_cb (abort, retval, msg) {
```
KingNoosh: Keep declaration's consistent

host_pool.js
msimerson: Not necessarily. If you don't need variable hoisting, why maintain this unusual style of declaration?

KingNoosh: The weird way that js starts setting functions is that if they're defined with `function x` means it's everywhere even if the block it's contained in doesn't run.

Whereas if you define them as a `var`, `let`, or `const`, you can have a function named `process` within two different blocks of code and they wouldn't interfere with each other.

It's just keeping it cleaner and nothing messes with each other or introduces any weird bugs because someone's defined `function x` in two different blocks of code.

http://adripofjavascript.com/blog/drips/variable-and-function-hoisting

bin/haraka
```diff
@@ -84,17 +85,17 @@ var usage = [
 ].join('\n');
 
 
-var listPlugins = function (b, dir) {
+function listPlugins (b, dir) {
```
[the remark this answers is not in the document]
msimerson: That can go either way, I have no strong preference. This way required the least significant change. If we're going to go with arrow functions (a topic for another Issue and PR), that will get addressed then. I'm rather keen to keep this PR limited to deprecating var and minimizing the changes required due to const, let, and function not getting hoisted.

connection.js
```diff
@@ -906,10 +908,10 @@ Connection.prototype.ehlo_respond = function (retval, msg) {
                 self.disconnect();
             });
             break;
-        default:
+        default: {
```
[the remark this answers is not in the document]
msimerson: Yes, read the eslint docs. Remove them and run eslint to see. ;-)

connection.js
```diff
@@ -1182,39 +1184,40 @@ Connection.prototype.rcpt_respond = function (retval, msg) {
             this.last_rcpt_msg = msg;
             plugins.run_hooks('rcpt_ok', this, rcpt);
             break;
-        default:
+        default: {
```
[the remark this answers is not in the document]
msimerson: The braces establish a block that contains the scoped declarations.

logger.js
msimerson: because when I added that, I hadn't yet found this option:
````
"prefer-const": ["error", {"ignoreReadBeforeAssign": true}],
````
I'll remove it (it's not longer needed).

bin/haraka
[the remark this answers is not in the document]
KingNoosh: Could we at the very least with the functions that used to be declared with a `var` be declared with a `const`?

That seems like a lesser change than turning them into pure function declarations.

msimerson:
> The only thing I'd be adamant on is putting the eslint-disable comments at the top of the file

I strongly disagree. Putting them inside the scope they effect limits them to only the section of code that needs the exception, as well as making it way more obvious **why** they are needed. (Hey ya'll, this chunk of code right here is an exception to the rule). It's no matter though, because I've removed them all as they were only needed before I found and set prefer-const's ignoreReadBeforeAssign option.

msimerson:
> Keep declaration's consistent

I'd like to hear additional opinions. Changing `var foo = function ()` to `const foo = function ()` isn't consistent because `const` doesn't do variable hoisting like `var` does. OTOH, switching from a function expression `var foo = function ()` to a function declaration `function foo ()` **does** hoist the function declaration to the top of its scope. I'd ague that preserving *how it works* is more consistent than preserving how it *looks*. Switching to declaration syntax also makes the functions named instead of anonymous (sometimes helpful for debugging) and avoids the [temporal dead zone](https://stackoverflow.com/questions/33198849/what-is-the-temporal-dead-zone#33198850).

References:
- [var functionName = function() {} vs function functionName() {}](https://stackoverflow.com/questions/336859/var-functionname-function-vs-function-functionname)
- [Proper use of const for defining functions in JavaScript](https://stackoverflow.com/questions/33040703/proper-use-of-const-for-defining-functions-in-javascript)
- [Variables and scoping in ECMAScript 6](http://2ality.com/2015/02/es6-scoping.html)

smfreegard:
> I'd ague that preserving how it works is more consistent than preserving how it looks. 

I agree 100%.   This PR is scary enough as it is.

smfreegard:
> So what I would like to do, if you agree, is push a quick point release for the outbound fixes before merging this. Thoughts?

Personally - I think this is a prudent approach.  This pull is massive and whilst it might lint OK, there's always the potential for breakage in such a big change.

We've been waiting for the outbound changes for ages and we've finally got something that we think should work better, so I'd rather get that out so we don't have a broken 'latest' as we currently do

 It would be annoying if we pushed out the fixes for outbound only for something else to get broken in the meantime by such a big change.   It's not like this pull is going to make a fundamental difference to the performance or bring any new functionality (aside from variable checking).   It's just a 'nice to have'.

host_pool.js
msimerson: > The weird way that js starts setting functions is that if they're defined with function x means it's everywhere even if the block it's contained in doesn't run.

That's not quite how hoisting works. Function declarations aren't available **everywhere** unless you declare them in global scope. Function declarations are only hoisted to the top of their current block/scope. See the links I posted below which go into much greater details. `var` is hoisted the same way, but `let` and `const` are not. Hence, prior to this change, a function expression using var was hoisted the same as a function declaration. When you switch to const/let, you lose the hoisting entire, creating the infamous Temporal Dead Zone. Again, the links below delve into the gritty details.

msimerson:
rebased, fixing all the merge conflicts.

msimerson:
rebased again, fixing merge conflicts. Any ETA on that quick release @baudehlo, I'd like to get this landed and start a quiet period.

msimerson:
For @baudehlo's planned quick release, I've created a draft release tagged to HEAD right now. That'll make it easy to draft a release prior to this big scary PR landing. In the meantime, we really ought to get this landed.

PR #2079, opened 2017-09-04
add transaction.queue for queue plugins

msimerson:
This is a lite version of #2067 

Changes proposed:
- adds transaction.queue object (to extend queue plugins without having to `if (txn.notes.queue && txn.notes.queue.$var)` or creating a mess of `if (txn.notes.queue_$var)`.
- `queue.wants` is a plugin name.
    - when defined, plugins that *aren't* that name skip queueing
        - this adds an "off switch" to on-by-default plugins: deliver, lmtp, qmail-queue, smtp_*, rabbit_*
    - when defined, plugins that *are* that name are enabled
        - this adds an `on` switch to off-by-default plugins: discard, quarantine
- greatly simplifies coordination when 3+ queue plugins are enabled

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -4,6 +4,7 @@
     * Permit log settings to be set w/o LOG prefix #2057
 * New Features
     * TLS certificate directory (config/tls). #2032
+    * transaction.queue.wants #2079
 
 ## 2.8.14 - Jul 26, 2017
 
--- a/docs/Transaction.md
+++ b/docs/Transaction.md
@@ -166,7 +166,7 @@ Store results of processing in a structured format. See [docs/Results](http://ha
 
 An object populated with queue related data, including:
 
-    * **wants** Default: undefined.
+* transaction.queue.wants (Default: undefined)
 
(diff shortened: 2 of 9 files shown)


msimerson:
> Why do we need a 'lite' version of #2067? - I've already reviewed it and was simply awaiting an answer to my question

Because my powers of prophecy are weak and this was created 6 hours before your question appeared.

smfreegard:
> Because my powers of prophecy are weak and this was created 6 hours before your question appeared.

I commented over 3 days ago as I made the comment immediately after the comment regarding the change from `regiter()` to `init_master()` -  however Github has changed the time of it since I submitted the Review earlier.

msimerson:
Neither of those comments was visible to me, I only saw the comments regarding register/init_master in #haraka. Closing and moving this over to #2067

smfreegard:
Ok - sorry, that's my bad then.  I didn't realise that you don't see the comments until I actually submit the review.  Github doesn't make that clear, but it's my fault.

Issue #2075, opened 2017-09-03
es6: use destructuring assignment


msimerson:
replace instances of code like this:

```js
var res = /some/fancy/regex/.exec(something);
if (res[1]) .....
if (res[2]) ....
```

with:

```js
const [ , well, named, variables]  = /some/fancy/regex/.exec(something);
...
if (well || named) ...
```

romainwn:
Hi, I would liketo work on this ! starting the PR's from October 1st :)

KingNoosh:
@romainwn No problem, please could you do a PR per file you update? This is to make it easier for us to review them.

Thanks!

shabbir1993:
Hello, Can I work on this ???

romainwn:
@shabbir1993 sure !

PR #2089, opened 2017-09-08
Fix rfc2231 parsing code to cope with continuation

baudehlo:
Fixes a bug found by an emailitin.com user.

Basically this header:

```
Content-Disposition: attachment; \r
   filename*0="marketron_lbasubmission_FTA Cricket Brentwood 3006445\r
 Jackso"; filename*1="n TN_08282017.xlsx"\r
```

Would cause Haraka to fail to properly process the header.

Checklist:
- [ ] docs updated
- [x] tests updated

### DIFF
--- a/mailheader.js
+++ b/mailheader.js
@@ -168,7 +168,7 @@ function _decode_rfc2231 (params, str) {
 
 Header.prototype.decode_header = function decode_header (val) {
     // Fold continuations
-    val.replace(/\r?\n/g, '');
+    val = val.replace(/\r?\n/g, '');
 
     const rfc2231_params = {
         kv: {},
--- a/mailheader.js
+++ b/mailheader.js
@@ -99,57 +99,59 @@ function _decode_header (matched, encoding, lang, cte, data) {
     return data.toString();
 }
 
-function _decode_rfc2231 (params) {
-    return function (matched, str) {
-        var sub_matches = /^(([^=]*)\*)(\d*)=(\s*".*?[^\\]";?|\S*)\s*$/.exec(str);
-        if (!sub_matches) {
-            return " " + str;
-        }
-        var key = sub_matches[1];
-        var key_actual = sub_matches[2];
-        var key_id = sub_matches[3] || '0';
-        var value = sub_matches[4].replace(/;$/, '');
-
-        var key_extract = /^(.*?)(\*(\d+)\*)$/.exec(key);
-        if (key_extract) {
-            key_actual = key_extract[1];
-            key_id = key_extract[3];
-        }
-
-        var quote = /^\s*"(.*)"$/.exec(value);
-        if (quote) {
-            value = quote[1];
-        }
-
-        var lang_match = /^(.*?)'(.*?)'(.*)/.exec(value);
-        if (lang_match) {
-            if (key_actual == params.cur_key && lang_match[2] != params.cur_lang) {
-                return ''; // same key, different lang, throw it away
-            }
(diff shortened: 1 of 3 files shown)


msimerson:
````
mailheader
× basic - parse_basic
AssertionError: 'multipart/alternative;   boundary=Apple-Mail-F2C5DAD3-7EB3-409D-9FE0-135C9FD43B69' == 'multipart/alternative; boundary=Apple-Mail-F2C5DAD3-7EB3-409D-9FE0-135C9FD43B69'
√ basic - content type w/parens
````

File: mailheader.js [hunk 1]
```diff
@@ -99,57 +99,77 @@ function _decode_header (matched, encoding, lang, cte, data) {
...
+                  ; to use within parameter values
+    */
+    var sub_matches = /(([!#$%&'*+.0-9A-Zdiff^_`a-z{|}~-]*)\*)(\d*)=(\s*".*?[^\\]";?|\S*)/.exec(str);
+    if (!sub_matches) {
+        return;
+    }
+    var key = sub_matches[1];
+    var key_actual = sub_matches[2];
+    var key_id = sub_matches[3] || '0';
+    var value = sub_matches[4].replace(/;$/, '');
```
msimerson: Is it too soon to use es6 destructing assignment? IE:

```js
let [, key, key_actual, key_id, value] = /(([!#$%&'*+.0-9A-Zdiff^_`a-z{|}~-]*)\*)(\d*)=(\s*".*?[^\\]";?|\S*)/.exec(str);
if (!key) return;
value = value.replace(/;$/, '');
```

baudehlo: I'm not even sure this works on regexps because it's not returning an array, it's some funky object which happens to have 0,1,2,3, etc, but also has .index and .input values.

Anyway, I'm not going to change it. It works, and that trumps my time for experimenting ;)

msimerson: I haven't used it for regex yet myself but you around my curiosity. It seems to work as advertised:
```js
$ node
> let [,a,b]= /(the) (rain) (in)/.exec('the rain in spain');
undefined
> console.log(a);
the
undefined
```

We'll update it later when we get around to #2075

tests/mailheader.js
```diff
@@ -32,7 +32,7 @@ exports.basic = {
         test.equal(h.lines().length, 12);
         test.equal(
             h.get_decoded('content-type'),
-            'multipart/alternative; boundary=Apple-Mail-F2C5DAD3-7EB3-409D-9FE0-135C9FD43B69'
+            'multipart/alternative;   boundary=Apple-Mail-F2C5DAD3-7EB3-409D-9FE0-135C9FD43B69'
```
msimerson: If I'm reading the changes correctly, that additional WS was required due to an improved parser that no longer strips the folded whitespace. Correct? If so, 👍.

baudehlo:
Yeah - it should never have been trimmed as far as I can tell.
Continuations are to be just glued on the end, not trimmed then glued with
a space. I think.

msimerson:
> Yeah - it should never have been trimmed as far as I can tell. Continuations are to be just glued on the end, not trimmed then glued with a space. I think.

Yes, we shouldn't be trimming WS. IIRC, the oldest rules were folding whitespace was stripped (FWS = CRLF[WS+] and replaced with a single space, but then around either RFC 822 or 2822, it was updated to instead insert the CRLF before a WS char in the header, so that unfolding left the exact original header intact. I think.

PR #2091, opened 2017-09-10
Use the available multipart content-type to add HTML code to bounces

### LINKED ISSUE #2090: Use the available multipart content-type to add HTML code to bounces
### system info

Haraka | Haraka.js — Version: 2.8.14
 --- | :--- 
Node | v8.4.0
OS | Linux devs01.spamfree.io 4.4.0-1022-aws #31-Ubuntu SMP Tue Jun 27 11:27:55 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
openssl | OpenSSL 1.0.2g  1 Mar 2016

### Expected behavior

Noticed that although the bounces have the type: "Content-Type: multipart/report; report-type=delivery-status;", it is currently not possible to add an HTML part in the bounce like Gmail does for example. Making it more easy to view the bounce and possibly even add the mail itself to the bounced message as attachment?

### Observed behavior

Ugly bounce that misses headers which has been logged +1 year ago in issue #1688 
Also the content of the bounced mail could be added to the bounce itself?

### Steps to reproduce

Trigger a bounce.
For example by creating an email forward (using aliases) to an email address that does not exist.

Bramzor:
Fixes #2090

My first pull request so sorry if it isn't perfect.
(Needs to be tested. Was unable to test it myself in my current setup. Code itself is copied over but I was behind of the master.)
Todo: Fix Address objects so they print out an email address without the < >.

- Added HTML Bounces to Haraka 

Checklist:
- [x] docs updated
- [ ] tests updated


smfreegard:
Ok - thanks, this looks good.  I'll try and test it later, for now though I've marked it as work in progress.

msimerson:
Are we intending to change the default bounce to include the HTML and image? If not, shouldn't the outbound/bounce_message_* files live in test/config?

smfreegard:
> Are we intending to change the default bounce to include the HTML and image?

I think we should make this the default.  It should give us some free advertising and it's easier to read for most people.

What do you think?

msimerson:
I was more curious than caring. My only caution would be getting "free advertising" we don't want. "Why are you Haraka people spamming me with these messages?," type of stuff, b/c recipients don't understand the message (IME, very few people grok DSN/bounces).

smfreegard:
>  "Why are you Haraka people spamming me with these messages?

That's a good point.  I'll think about it and look at how the image is presented in the HTML.

> b/c recipients don't understand the message (IME, very few people grok DSN/bounces).

Indeed - although that is one of the reasons for including an HTML version of the bounce as we can address this shortcoming and make it far more clearer than it otherwise would.

Once I get around to testing this - I'll include a screenshot here so you can see the proposed changes and wording.

Bramzor:
We could start making it optional now as intended, so good point, we can move it to test/config in that case. But we can make it the default afterwards as it is a real improvement to users. Bounces are indeed not user friendly but with the HTML, it is at least an improvement on that.
The image that is included by default is the image from the google bounce. Just an error image. It's not the purpose to add additional images for marketing or anything like that. Just to make it clear the message is not delivered.

msimerson:
Something else useful to test is how the bounce HTML is rendered in gmail/outlook/yahoo webmail. MUA's are generally pretty good but the webmail providers sometimes cruelly butcher the appearance of otherwise well-formed HTML.

Bramzor:
Will give it a test later today. But should be ok as it is probably already tested by google :-)

Bramzor:
Screenshot example: https://ibb.co/dfFZgv
Remembers me that I had to do an additional fix to show the "to" field as the current includes the < and >.

msimerson:
Looks great!  Although I think I want to clean up the HTML and replace most of the inline styling and table with a tiny bit of CSS and a couple divs.

Bramzor:
Than you might introduce issues with webmail, that is why they don't use that. The HTML looks ugly and old, but it does work in webmail clients as it cannot be impacted by external css and you don't see the HTML code anyway unless in the source.

msimerson:
> Than you might introduce issues with webmail, that is why they don't use that.

Well that's sad. I have dim recollections of this from years ago and expected things to be a little better now.

baudehlo:
Yeah we have to use `juice` on our HTML emails at work to inline all the
styles.

https://www.npmjs.com/package/juice

KingNoosh:
What's remaining to do on this pr?

msimerson:
Same with this one, I'd like to merge this tomorrow, after cutting a release. It's got some manual testing and it can get more widely tested in HEAD.

PR #2100, opened 2017-09-11
Replace console.log with stdout

KingNoosh:
I was facing an issue where two things I was running were trying to override console.log, mocha and Haraka, this is to avoid any sort of future compatibility issues.

- Replace console.log with stdout to avoid conflict with anything else that overrides console.log

Checklist:
- [x] docs updated
- [x] tests updated


logger.js
```diff
@@ -160,14 +161,17 @@ logger.log = function (level, data) {
 logger.log_respond = function (retval, msg, data) {
     // any other return code is irrelevant
     if (retval !== constants.cont) { return false; }
-
+    let timestamp_string = '';
+    if (logger.timestamps) {
+        timestamp_string = `${new Date().toISOString()} `;
```
msimerson: Don't you have a "why are my log timestamps N days ago" problem going on here?

KingNoosh: Nope, but there's another bug actually, set_timestamps is being set twice, first by the .ini, second by the value file.

value seems to always be used no matter what the ini says.

I'm working on that now

baudehlo:
Didn't I read somewhere that console.log had different behaviour wrt buffering?

KingNoosh:
https://github.com/nodejs/node/blob/v6.x/lib/console.js#L43
It just does stdout with a new line

PR #2110, opened 2017-09-13
remove eslint rules to plugin

msimerson:
depends on haraka/haraka-eslint#13


msimerson:
CI tests pass now, after publishing newer `haraka-eslint`, manually nuking the Travis-CI caches, and then restarting the tests.

PR #2119, opened 2017-09-20
replace the rest of ./config with haraka-config

msimerson:
- [x] depends on haraka/haraka-config#27

- replace ./config with haraka-config

Checklist:
- [x] docs updated
- [x] tests updated
- [x] deployment tested


msimerson:
I moved the connection state constants into haraka-constants because it created race conditions in the test suite. Especially [right here](https://github.com/haraka/Haraka/blob/afd53239fb76e8818791e8f7cdb273ffa45b4865/plugins.js#L14).

I also discovered a little bug in plugins.js:

`if (object.state >= states.DISCONNECTING)`

That didn't work as intended because `states.DISCONNECTING` didn't exist, in connection.js it was defined as states.STATE_DISCONNECTING. It's more natural to drop the redundant STATE prefix so that's what I did in [haraka-constants](https://github.com/haraka/haraka-constants/pull/8/files), so it can be referenced either way.

smfreegard:
My only issue with this is that `haraka-config` uses the split `readers` that I objected to when you tried to pull it in to Haraka several years ago.

There's no mention in`haraka-config` of the JSON/YAML overrides that I added and are currently supported in Haraka right now (that I use extensively).

My opinion on this hasn't changed, if you're going to switch config loaders (e.g. code that works now), then absolutely everything needs to be exercised in tests so that swapping in/out loaders proves that everything works as it did before the switch.

I'll try and add some additional tests for these, but my time is really limited at present.

msimerson:
> There's no mention in haraka-config of the JSON/YAML overrides that I added

If you're referring to the "if config.json doesn't exist, use config.yaml" feature, that's [been in haraka-config](https://github.com/haraka/haraka-config/blob/master/configfile.js#L339-L353) since it's inception. Here is [another test](https://github.com/haraka/haraka-config/pull/28/files) to to along with the [existing one for a missing .json file](https://github.com/haraka/haraka-config/blob/master/test/config.js#L292-L294).

If you're referring to some other overrides behavior, please explain.

haraka-config is already **the** config loader used by all npm packaged plugins and nearly all of the test suite (via haraka-test-fixtures). 

- bin/haraka, connection.js, logger.js, mailbody.js, outbound/*.js, plugins.js, server.js, transaction.js

haraka-config is some of the best tested code we've got [Coverage Status](https://codecov.io/github/haraka/haraka-config?branch=master) but there's always room for additional coverage.

smfreegard:
I'm taking about this:  https://github.com/haraka/haraka-config/blob/master/configfile.js#L382-L407

You can use JSON or YAML files to override any other file by prefixing the outer variable name with a ! e.g.

`````
{
    "!smtpgreeting": [ 'this is line one', 'this is line two' ]
}
`````

If the config/smtpgreeting file did not exist, then this value would replace it.

NOTE: You must ensure that the data type (e.g. Object, Array or String) for the replaced value is correct. This cannot be done automatically.

**Comment by KingNoosh** [comment]
People who rely on 'Haraka/config' will have their addin broken, is it possible to keep config.js but having that export haraka-config and spit out a log message that it's going to be removed soon?

**Comment by msimerson** [comment]
> You can use JSON or YAML files to override another file by prefixing the outer variable name with a !

I see. I've [added another commit](https://github.com/haraka/haraka-config/pull/28/commits/a86774a2149357a754c0abd3f2b26d7ed5616b01) which adds another test covering this example. Here's the test output:

````
✔ jsonOverrides - no override for smtpgreeting
Overriding file smtpgreeting with config from /Users/matt/git/haraka-config/test/config/main.json
{ '!smtpgreeting': [ 'this is line one', 'this is line two' ] }
✔ jsonOverrides - with smtpgreeting override
````

**Comment by smfreegard** [comment]
Ok - thank you.

**Comment by msimerson** [comment]
> People who rely on 'Haraka/config' will have their addin broken

Nobody should be relying on Haraka/* for anything in their plugins. Somewhere on the 3.0 list of things to do is remove `require('./')`. Haraka's "external" API is the plugin interface and we provide the config loader to plugins at `this.config`.

**Comment by msimerson** [comment]
It appears we're going to see a bunch of PR activity in October. Should we cut another release very soon (we've got handful of bug fixes in already)? Then get this and the other open PRs merged before the onslaught begins?

**Comment by baudehlo** [comment]
Probably, unless we think there will be progress soon on the callback called twice bug. I’d love to see that fixed.

**Comment by msimerson** [comment]
I had a chance to perform deployment tests on this today and found no surprises. I'd like to cut a release release tomorrow and then get this merged after. @smfreegard ?

**Comment by baudehlo** [comment]
> People who rely on 'Haraka/config' will have their addin broken, is it possible to keep config.js but having that export haraka-config and spit out a log message that it's going to be removed soon?

I agree with this. If anyone has a custom plugin that does `require('./config')` it will be broken by this change, and by semver we promise not to do that.

It can be fixed by just changing config.js to:

```javascript
"use strict"

module.exports = require('haraka-config');
// use emit is the same way util.deprecate does it, so follow that style
process.emit('warning', 'Loading config via require("./config") is deprecated, please use: require("haraka-config") instead.');
```

msimerson:
I have added the shim with warning as suggested.

PR #2120, opened 2017-09-20
Travis ubuntu now has new enough gcc

msimerson:
Yay, we don't need these any more. This will speed up builds by quite a bit.


msimerson:
From the Travis run:

````
gcc version
gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4
````

PR #2128, opened 2017-09-26
Fix ability to set log level to emerg

### LINKED ISSUE #2127: New loglevel "emerg" is invalid
### system info

```
Docker Base Image: node:alpine
Haraka | Haraka.js — Version: 2.8.15
 --- | :---
Node | v8.5.0
OS | Linux haraka-alpine-1 4.4.0-96-generic #119-Ubuntu SMP Tue Sep 12 14:59:54 UTC 2017 x86_64 Linux
openssl | OpenSSL 1.0.2k  26 Jan 2017
```

### Expected behavior

If "emerg" is a loglevel it should not fail when used.

### Observed behavior

When setting loglevel to "emerg" in log.ini the following happens in Haraka:

```
loglevel: EMERG
[DEBUG] [-] [core] Returning boolean false for main.timestamps=false
invalid loglevel: emerg defaulting to LOGWARN
log format: DEFAULT
```

### Steps to reproduce

log.ini

```
[main]

; level=data, protocol, debug, info, notice, warn, error, crit, alert, emerg
level=emerg

; prepend timestamps to log entries? This setting does NOT affect logs emitted
; by logging plugins (like syslog).
timestamps=false

;  format=default, logfmt
format=default
```

KingNoosh:
Fixes #2127

- Fix ability to set log level to `emerg`

Checklist:
- [x] tests updated


msimerson:
PR is good, thanks. It would be even better if it included a test case.

KingNoosh:
Yeah, just need to get home and update changes too.

tests/logger.js
```diff
@@ -76,6 +76,13 @@ exports.set_format = {
         test.equal(this.logger.format, this.logger.formats.DEFAULT);
         test.done();
     },
+    'set format to EMERG' : function (test) {
+        test.expect(1);
+        this.logger.format = '';
+        this.logger.set_format('EMERG');
+        test.equal(this.logger.format, this.logger.formats.EMERG);
```
msimerson: I think your problem is here. You want to compare against logger.levels.EMERG.

KingNoosh: d'oh

msimerson: While you're in there, maybe also add a test for `this.logger.set_format('emerg');` as well, since there doesn't appear to be any lower case tests.

KingNoosh: you mean set_level?

msimerson: LOL, yes. That's what I meant. :)

msimerson: @KingNoosh, you going to fix this or what? My merge finger is getting itchy.

KingNoosh:
I've been far too outgoing recently, tmrw morning I'll be doing it

msimerson:
Still waiting...

tests/logger.js
```diff
@@ -114,6 +121,12 @@ exports.set_loglevel = {
         test.equal(this.logger.loglevel, this.logger.levels.INFO);
         test.done();
     },
+    'set loglevel to EMERG' : function (test) {
+        test.expect(1);
+        this.logger.set_loglevel('emerg');
+        test.equal(this.logger.loglevel, this.logger.levels.INFO);
```
msimerson: eh, INFO?

logger.js
```diff
@@ -176,19 +176,16 @@ logger.log_respond = function (retval, msg, data) {
 };
 
 logger.set_loglevel = function (level) {
-
-    if (!level) return;
```
msimerson: instead of removing this, I'd replace it with an `if (level === undefined || level === null)`

msimerson: Your current Travis test failure answers: why?

Issue #2129, opened 2017-09-28
ES6: Replace string concatenations with template literals


KingNoosh:
We currently use string concatenations all over the codebase, easily found whenever we try to log something.

It is currently known that template literals are more efficient than string concatenation.

To make the codebase just that bit neater, we'd like to receive pull requests PER file of transforming string concatenations to template literals.

If you have any questions, drop them in the comments. 🎃

kajes:
I'll start working on this.

KingNoosh:
Thanks @kajes, if you're doing this in the hopes for Hacktoberfest, remember that only pull requests from the 1st of October will be counted by GitHub.

And remember that we want ONE easily reviewable pull-request per file. :)

kajes:
Well, given the amount of concatenations on my first look, I don't expect to make a pull request before the 1st of october anyway ;). Already gone through som of the files though. I'll keep a fork with my commits until I've gone through everything at least once.

SamuelGrave:
can I help on that issue too?

KingNoosh:
@kajes make sure it's one file per PR ;)
@SamuelGrave go ahead, fork, branch, pick a file, change it, commit it, push it, create a pr we'll will review

baudehlo:
Why one file per pr?

KingNoosh:
It's more to make the PRs easier to review. That last const/let PR was
pretty painful to go through.

jattoabdul:
i would also like to join on this too. Will go through the codebase and start on OCT 1st. Hopefully it will make me happy and fulfilled as my birthday is OCT 3rd.

KingNoosh:
Go ahead, just put in a PR and we'll review it!
No need to keep asking for permission. 😃

PSSGCSim:
Calling dibs on haraka/haraka-config if possible :smile_cat:

shabbir1993:
I will work on it...

martypowell:
@KingNoosh what else if left, I can contribute.

mtrivera:
@KingNoosh I'm currently working on dkim.js

pvsousalima:
@KingNoosh  I am working on logger.js

pvsousalima:
I got plugins/clamd.js !

Hari70a:
@pvsousalima 
Can I change this file and give a PR on this

Hari70a:
@pvsousalima 
I have change this file and made a PR.Can u please review it #2234

PR #2151, opened 2017-10-01
Replaced concatenated strings with template literals in the plugins/auth/auth_ldap.js

stefanlw:
- Replaced concatenated strings with template literals in the plugins/auth_ldap.js


PSSGCSim:
https://github.com/haraka/Haraka/issues/2129#issue-261253525

As @KingNoosh said: 

> To make the codebase just that bit neater, we'd like to receive **pull requests PER file** of transforming string concatenations to template literals.

PR #2158, opened 2017-10-01
replaced concatenated strings with template literals in the plugins/auth/auth_ldap.js

stefanlw:
- replaced concatenated strings with template literals in the plugins/auth/auth_ldap.js

Checklist:


msimerson:
duplicate of #2139

PR #2159, opened 2017-10-01
replace concatenated strings with template literals in plugins/bounce

stefanlw:
…bounce.js

- replaced concatenated strings with template literals in the plugins/auth/bounce.js

Checklist:


plugins/bounce.js
```diff
@@ -116,8 +116,8 @@ exports.single_recipient = function (next, connection) {
         return next();
     }
 
-    connection.loginfo(plugin, "bounce with too many recipients to: " +
-        connection.transaction.rcpt_to.join(','));
+    connection.loginfo(plugin, `bounce with too many recipients to: 
+        ${connection.transaction.rcpt_to.join(',')}`);
```
msimerson: This change introduces an extra \n in the string.

plugins/bounce.js
```diff
@@ -370,22 +370,22 @@ exports.bounce_spf = function (next, connection) {
                     connection.logerror(plugin, err.message);
                     return run_cb();
                 }
-                connection.logdebug(plugin, 'ip=' + ip + ' ' +
-                                            'spf_result=' + spf.result(result));
+                connection.logdebug(plugin, `ip=${ip}
```
msimerson: Same here

plugins/bounce.js
```diff
@@ -370,22 +370,22 @@ exports.bounce_spf = function (next, connection) {
...
                 switch (result) {
                     case (spf.SPF_NONE):
                         // falls through, domain doesn't publish an SPF record
                     case (spf.SPF_TEMPERROR):
                     case (spf.SPF_PERMERROR):
                         // Abort as all subsequent lookups will return this
-                        connection.logdebug(plugin, 'Aborted: SPF returned ' +
-                                                    spf.result(result));
+                        connection.logdebug(plugin, `Aborted: SPF returned 
+                                                    ${spf.result(result)}`);
```
msimerson: And hee

plugins/bounce.js
stefanlw: What would be the preferred way of resolving this?
(See my new commit)

plugins/bounce.js
```diff
@@ -370,22 +370,22 @@ exports.bounce_spf = function (next, connection) {
                     connection.logerror(plugin, err.message);
                     return run_cb();
                 }
-                connection.logdebug(plugin, 'ip=' + ip + ' ' +
-                                            'spf_result=' + spf.result(result));
+                connection.logdebug(plugin, `ip=${ip} \
+                spf_result=${spf.result(result)}`);
```
msimerson: It's a judgement call, and this is okay, but I think it would be better on one line. Thoughts?

plugins/bounce.js
```diff
@@ -370,22 +370,22 @@ exports.bounce_spf = function (next, connection) {
...
                 switch (result) {
                     case (spf.SPF_NONE):
                         // falls through, domain doesn't publish an SPF record
                     case (spf.SPF_TEMPERROR):
                     case (spf.SPF_PERMERROR):
                         // Abort as all subsequent lookups will return this
-                        connection.logdebug(plugin, 'Aborted: SPF returned ' +
-                                                    spf.result(result));
+                        connection.logdebug(plugin, `Aborted: SPF returned \
+                        ${spf.result(result)}`);
```
msimerson: same here.

plugins/bounce.js
```diff
@@ -116,8 +116,8 @@ exports.single_recipient = function (next, connection) {
         return next();
     }
 
-    connection.loginfo(plugin, "bounce with too many recipients to: " +
-        connection.transaction.rcpt_to.join(','));
+    connection.loginfo(plugin, `bounce with too many recipients to: \
+    ${connection.transaction.rcpt_to.join(',')}`);
```
msimerson: You've introduced a bunch of inadvertent whitespace. See:

````
$ cat foo.js 
let a=`this is a\
   line.`;
console.log(a);
$ node foo.js 
this is a   line.
````

plugins/bounce.js
stefanlw: I agree however I was just trying to keep consistent with the original formatting.

msimerson:
closing, changes already made in #2129

PR #2161, opened 2017-10-01
2129/use string templates host pool

brunolm:
related to #2129

Checklist:


tests/host_pool.js
```diff
@@ -11,9 +11,9 @@ exports.HostPool = {
         const host = pool.get_host();
 
         test.ok( /\d\.\d\.\d\.\d/.test(host.host),
-            "'" + host.host + "' looks like a IP");
+            `' ${host.host}' looks like a IP`);
```
msimerson: there's an extra space after the first single quote.

brunolm:
Sorry, fixed.

PR #2163, opened 2017-10-01
Rewriting the code of `haraka.js` to ES6/ES2015

### LINKED ISSUE #2075: es6: use destructuring assignment

hejty:
The code of the `haraka.js` file was rewritten in `ES6`

### DIFF
--- a/haraka.js
+++ b/haraka.js
@@ -2,6 +2,7 @@
 
 'use strict';
 const path = require('path');
+const makePathJoin = () => path.join(process.env.HARAKA, 'node_modules');
 
 if (!process.env.HARAKA) {
     console.warn("WARNING: Not running installed Haraka - command line arguments ignored")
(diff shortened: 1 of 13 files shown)


msimerson:
Since this is for a single file and the issue applies to many, you should remove the `fixes ...` from the description.

haraka.js
```diff
@@ -10,67 +11,64 @@ if (!process.env.HARAKA) {
...
+process.on('uncaughtException', err => {
     if (err.stack) {
-        err.stack.split("\n").forEach(function (line) {
-            logger.logcrit(line);
-        });
-    }
-    else {
-        logger.logcrit('Caught exception: ' + JSON.stringify(err));
+        err.stack.split("\n").forEach(line => logger.logcrit(line));
+    } else {
```
msimerson: style issue: we don't cuddle elses.

haraka.js
```diff
@@ -10,67 +11,64 @@ if (!process.env.HARAKA) {
...
+        process.title = path.basename(filename, '.js');
+
+        logger.lognotice(`${sig} received`);
+        logger.dump_and_exit(() => {
             if (server.cluster && server.cluster.isMaster) {
                 server.performShutdown();
-            }
-            else if (!server.cluster) {
+            } else if (!server.cluster) {
                 server.performShutdown();
```
msimerson: same, we don't cuddle elses.

PR #2165, opened 2017-10-02
connection: set this.local properties

msimerson:
- ~~default to os.hostname() when config/me does not exist~~
    - previously, if config/me didn't exist: `220 null ESMTP Haraka 2.8.16 ready`
- replace most concat strings with template literals
- replace some anon callbacks with arrow functions

Checklist:
- [x] tests updated

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -25,7 +25,6 @@ const plugins     = require('./plugins');
 const rfc1869     = require('./rfc1869');
 const outbound    = require('./outbound');
 
-const hostname    = (os.hostname().split(/\./))[0];
 const hpj         = fs.readFileSync(path.join(__dirname, 'package.json'));
 const version     = JSON.parse(hpj).version;
 const states      = constants.connection.state;
@@ -62,7 +61,7 @@ class Connection {
         this.local = {           // legacy property locations
             ip: null,            // c.local_ip
             port: null,          // c.local_port
-            host: null,
+            host: config.get('me') || os.hostname(),
         };
         this.remote = {
             ip:   null,          // c.remote_ip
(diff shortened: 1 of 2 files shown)


msimerson:
This collides with #2166 so hold off on this. After that one lands, I'll update this.

msimerson:
I found [the commit by some @smfreegard that removed it](https://github.com/haraka/Haraka/commit/80e028821f37693cc9b374756a432b327657b8b6#diff-e22b50e740e102c8de5e858dbcd7753d) not too long after he added it. 🤷‍♂️ Sooooo, this has been broken before and after 2012. Basically forever.

I created the (now mostly unused) `connection.local.host` property a while back with this intent. Modifying config.js (again) to provide a consistent local hostname (as well as test coverage for it) for plugins that load `me` is an okay idea but that doesn't exclude setting a connection property. From an API perspective, I'd much rather see plugins inspect the connection property.

smfreegard:
> I found the commit by some @smfreegard that removed it not too long after he added it. :man_shrugging: > Sooooo, this has been broken before and after 2012. Basically forever.

Ok - I suspect I removed this when we made `haraka -i` automatically create `config/me` so it was always set.  Since we've npmized everything - that no longer holds true.

> I created the (now mostly unused) connection.local.host property a while back with this intent. Modifying config.js (again) to provide a consistent local hostname (as well as test coverage for it) for plugins that load me is an okay idea but that doesn't exclude setting a connection property. From an API perspective, I'd much rather see plugins inspect the connection property.

The problem with `connection.local.host` is that it's not going to be available outside of core, so using it over `config.get('me');` is going to be more problematic, whereas the latter is going to come from `haraka-config` and will therefore always work.

msimerson:
> The problem with connection.local.host is that it's not going to be available outside of core,

Even if incomplete, this change is an improvement because `connection` is hot code and calling `config.get` and then accessing a local object property is always going to be faster than calling `config.get('me')` multiple times.

smfreegard:
> calling g config.get and then accessing a local object property is always going to be faster

It's going to be tiny as the result for config.get is cached.

PR #2188, opened 2017-10-05
Base64 encoding

### LINKED ISSUE #2180: Base64 decoding appears to truncate data in some cases.
This issue was noted in "attachment" parsing, but applies equally to "body text" processing.

Haraka appears to parse base64 content line-by-line, and assembling the buffer of each line and then emitting the full data:

https://github.com/haraka/Haraka/blob/b310442c30eb2ebaa81c37838a72a083f7b7861a/mailbody.js#L381

When line lengths are not evenly divisible by `8/6`, this technique, when combined with node's `new Buffer(line, 'base64')` constructor can lead to data truncation.

Why `8/6`? Typically, `base64` data is broken with 76 chars of base 64 data, and a `\r\n` per line in a MIME part. This is in accordance with the recommendation in RFC 2822. Each character of a base64 encoded string carries 6 bits of data. When nodejs decodes this as a buffer, incomplete bytes are dropped.

However, RFC 2822 does not _require_ that base64 lines always be 76 chars, lines are sometimes encoded with fewer chars per line. So, for example, a message folds base64 lines every 75 chars, we'll end up with nodejs "trimming" 2 bits per line.

The way to correct for this is to capture all of the base64 lines before doing a single node buffer decode. This would also be beneficial, because if there are other filters, Haraka currently "reassembles" the base64 from the constituent line buffers.

Please let me know if you agree with this analysis, and if you'd be open to a patch to resolve it.

atheken:
Fixes #2180 

- Causes decode_base64 function to align buffers on base64 and byte boundaries.

Checklist:
- [ ] ~~docs updated~~
- [X] tests updated


mailbody.js
```diff
@@ -421,7 +422,39 @@ Body.prototype.parse_attachment = function (line) {
 Body.prototype.decode_qp = utils.decode_qp;
 
 Body.prototype.decode_base64 = function (line) {
-    return new Buffer(line, 'base64');
+    var toProcess = this.decode_accumulator + line.trim();
```
baudehlo: We prefer `to_process` style in this project.

baudehlo: Same with all the other variables btw.

tests/transaction.js
```diff
@@ -1,4 +1,7 @@
-const Transaction  = require('../transaction');
+const fs = require('fs');
+const path = require('path');
+
+const Transaction = require('../transaction');
```
baudehlo: This doesn't return a class (the class is in `exports.Transaction`), so don't capitalize.

atheken: The diff is showing this as a change, but that require statement was already there. I can update it, but it would introduce churn that wasn't required for this fix.

Let me know if you want me to introduce changes to correct lint warnings, even for parts of the files that were not related/modified by this PR.

baudehlo: Oh yeah ignore me then.

msimerson: It's just a test file. Churn implies no improvement, and I think we can all agree that non-class variables shouldn't be capitalized. I'd prefer it was corrected.

msimerson:
Are you all done @atheken ? Ready to merge?

atheken:
@msimerson Yep. Sorry for the turbulence, I was trying to rebase to clean up the commit order, but I practically never use that feature of git.

At any rate, I'm done. Thanks for the quick turnaround.

msimerson:
Thank you @atheken !

PR #2192, opened 2017-10-06
Set remote.is_private automatically when remote.ip is set

smfreegard:
Set connection.remote.is_private automatically whenever connection.remote.ip is set.

I noticed that when a host uses XCLIENT, the is_private property isn't set, so reflects the original value - rather than just settings it in xclient.js, I decided it made more sense to do it automagically so there's no way it can happen elsewhere.

Checklist:

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -238,14 +238,13 @@ class Connection {
         }
     }
     set (obj, prop, val) {
-        const self = this;
         if (!this[obj]) this.obj = {};   // initialize
 
         this[obj][prop] = val;  // normalized property location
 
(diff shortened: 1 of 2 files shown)


connection.js
```diff
@@ -239,10 +238,16 @@ class Connection {
         }
     }
     set (obj, prop, val) {
+        const self = this;
```
baudehlo: This seems like a pointless addition. Other than that this change looks fine.

smfreegard: *nod* - changed.

Issue #2133, opened 2017-09-28
ES6: Replace current classes with ES6 classes


baudehlo:
Currently all our classes in Haraka use the old style function declaration then adding `Name.prototype.method = function (...) { ... }` methods to them.

The plan would be to use ES6 `class` declarations instead.

This is a hacktoberfest ticket, so please don't submit PRs until after Oct 1st.

All PRs should be one file per PR to make it easier for us to review (it also gets you to 4 PRs faster, so double bonus!).

MuraraAllan:
i will be working on it. Starting with connection.js. After 1 OCT i will make the PR (respecting 1 file per PR).  As soon as i finish a file i ask for another.
There are bunch of functions to convert guys, don't be intimidate. I'm at IRC if you are willing to help in the conversion and need some support.

MuraraAllan:
For functions like setupClient i suggest turn into fat arrow scope too, what do you think about it?

mdew192837:
What's your IRC nick @MuraraAllan ?

MuraraAllan:
muraraallan

pvsousalima:
I can help with this!

MuraraAllan:
Working over spf.js

rw251:
ES6 classes allow use of getters and setters. If a class property appears to be a 'getter' should it be set as such? E.g. the 'length' property in outbound/timer_queue.

This does mean that other files would need updating as getting the length of the timerqueue would change from `timerqueue.length()` to `timerqueue.length`.

baudehlo:
We can do that another time.

Hari70a:
@msimerson  and @baudehlo 
Is this issue still persists.Also could u please tell me if some files is still pending.If so, Iam eager to do.Can u help me on this

andela-ieyo:
How do I get involved in this. Would love to contribute to this project.

KingNoosh:
Pick a file that needs updating, update it. If you can't find any, I'll try
to help locate one for you.

More specifically, fork this repository, clone it, branch, make changes,
make pull request, get reviewed, get merged.

Happy Hacktoberfest!
🎃🎃🎃

PR #2193, opened 2017-10-07
line_socket: Replaced previous classes with es6 classes

pvsousalima:
Changes #2133

Replaced previous classes with es6 classes

Checklist:
- [ ] docs updated
- [ ] tests updated


baudehlo:
These changes don't make sense and this file was already ES6 classes. Closing.

PR #2194, opened 2017-10-07
host_pool: Replaced previous classes with es6 classes 

pvsousalima:
Fixes #2133

- ES6: Replace current classes with ES6 classes on message stream module

Checklist:
- [ ] docs updated
- [ ] tests updated


pvsousalima:
@baudehlo changes reverted to cover only host_pool

PR #2197, opened 2017-10-09
outbound es6 + bug fix

msimerson:
- outbound/hmail: move the prototype functions into es6 class declaration (updates #2133)
- replace a few `function () {` with arrow functions `() => {`
- outbound: get tls config via tls_socket (vs net_utils)
- whenever an attachment had a multibyte character in the filename, the JSON header in the queue file was being truncated.

Checklist:
- [ ] ~~docs updated~~
- [x] tests updated


msimerson:
This outbound queue file JSON truncation issue is subtle and it was introduced in #2066. The `TLDR;` version is that the JSON.stringify length property is in characters and some characters (unicode, especially) take up more bytes than others. What is needed is the actual bytes consumed in the file, so create a Buffer from the JSON string and measure **that** length instead.

msimerson:
@baudehlo , I'd really like **you** to review this since it makes changes in `outbound/hmail.js` which you recently modified.

baudehlo:
LGTM

PR #2209, opened 2017-10-13
Add remove_msgid option to outbound

smfreegard:
Add `options.remove_msgid` to `outbound.send_email` which is useful when sending mail from a quarantine where the destination mailserver might discard messages if multiple messages with the same Message-Id header are received (Gmail/G-Suite and Exchange being two such examples).

Checklist:
- [X] docs updated


smfreegard:
Typo fixed.

smfreegard:
I've just added `remove_date` as I found I needed this as well.

outbound/index.js
```diff
@@ -158,6 +158,16 @@ exports.send_email = function () {
     }
 
     transaction.message_stream.add_line_end();
+
+    // Allow for the removal of Message-Id and/or Date headers which
+    // is useful when resending mail from a quarantine.
+    if (options && options.remove_msgid) {
```
msimerson: There's already several instances of `(options && options.something)`. I should think that at some point we'd do something proactive about it like `if (options === undefined) options = {}` or `const options = arguments[4] || {}` and thus simplify the following usages

docs/Outbound.md
```diff
@@ -348,16 +348,17 @@ should queueing to disk fail e.g:
...
-```dot_stuffed: true``` option, like so:
-    
-    outbound.send_email(from, to, contents, outnext, { dot_stuffed: true });
+    outbound.send_email(from, to, contents, outnext, options);
+
+Where `options` is a Object that may contain the following keys:
 
+| Key/Value              | Description                                                                               |
+|------------------------|--------------------------------------------------------------------------------------------
+| `dot_stuffed: true`    | Use this if you are passing your content dot-stuffed (a dot at the start of a line is doubled, like it is in SMTP conversation, see https://tools.ietf.org/html/rfc2821#section-4.5.2).|
```
msimerson: I realize you didn't add it but updating this to be a link: `[RFC 2821](https://tools.ietf.org/html/rfc2821#section-4.5.2)` would improve it.

smfreegard:
> @smfreegard , you still happy with this (after my changes)?

Of course - thanks for making them.  I hadn't had time to do them myself, but I've finally finished what I was working on now....

PR #2212, opened 2017-10-15
Read TLS key and cert files from tls.ini in auth_proxy

### LINKED ISSUE #1892: auth_proxy config
### Expected behavior
`auth_proxy` should use the `tls` plugin config for `key` and `cert` path. 

### Observed behavior
auth_proxy has hardcoded names (paths) for `key` and `cert`

**Current auth_proxy** (line 118 and 119)
```
var key = self.config.get('tls_key.pem', 'binary');
var cert = self.config.get('tls_cert.pem', 'binary');
```

amirkarimi:
Fixes #1892

### DIFF
--- a/plugins/auth/auth_proxy.js
+++ b/plugins/auth/auth_proxy.js
@@ -113,8 +113,8 @@ exports.try_auth_proxy = function (connection, hosts, user, passwd, cb) {
         response.push(rest);
         if (cont !== ' ') return;
 
-        const key = self.config.get(self.tls_cfg.main.key || 'tls_key.pem', 'binary');
-        const cert = self.config.get(self.tls_cfg.main.cert || 'tls_cert.pem', 'binary');
+        let key;
+        let cert;
 
         connection.logdebug(self, 'command state: ' + command);
         if (command === 'ehlo') {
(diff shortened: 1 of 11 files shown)


plugins/auth/auth_proxy.js
```diff
@@ -103,8 +113,8 @@ exports.try_auth_proxy = function (connection, hosts, user, passwd, cb) {
         response.push(rest);
         if (cont !== ' ') return;
 
-        let key;
-        let cert;
+        const key = self.config.get(self.tls_cfg.main.key || 'tls_key.pem', 'binary');
+        const cert = self.config.get(self.tls_cfg.main.cert || 'tls_cert.pem', 'binary');
```
msimerson: I'm kinda of torn on this change. The variable declarations need to be in this scope, but this change will load the (cached) TLS certificates, even if the next SMTP server never advertises `STARTTLS`. Is there some benefit to loading the certs up here?

amirkarimi: You're right I just wanted to prevent loading files several times in the loop. I'll return them back to the prev place as you mentioned they're cached. Right?

msimerson: Yep, the config loader caches the config files.

amirkarimi: Done

PR #2218, opened 2017-10-17
Header folding fix

atheken:
When headers are continued on new lines, the first leading space is not removed from the header content.

This PR correctly strips the first space on folded header lines so that they are not present in the decoded header.

Checklist:
- [ ] docs updated
- [X] tests updated

### DIFF
--- a/mailheader.js
+++ b/mailheader.js
@@ -56,7 +56,7 @@ class Header {
     decode_header (val) {
         // unfold continuations, strip the leading
         // space char after the first line
-        let values = val.split(/[\r\n]+/);
+        const values = val.split(/[\r\n]+/);
         val = values.shift();
         while (values.length > 0) {
             val += values.shift().replace(/^[\s]/, '');
--- a/mailheader.js
+++ b/mailheader.js
@@ -54,8 +54,13 @@ class Header {
     }
 
     decode_header (val) {
-        // Fold continuations
-        val = val.replace(/\r?\n/g, '');
+        // unfold continuations, strip the leading
+        // space char after the first line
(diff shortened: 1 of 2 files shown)


atheken:
OK, looks like some of the tests have "wrong" test data, so this is still failing, will follow up when it's all fixed up.

baudehlo:
I don’t think this is right at all. You can only fold headers on CFWS. It needs to be left as white space.

atheken:
Yes, agreed. I realized that after the fact when looking at some dialed tests.

There is still an issue of encoded-word and plain text header folding.. those extra leading spaces are not really intended for display in that case

//Andrew

baudehlo:
Yes the problem there is that each header gets treated differently so we need to customize the parser for different headers. We can do it “right” now for rfc2822 headers (we can but we don’t - for “reasons”). We just need more parsers.

atheken:
I definitely familiar with _reasons_. Many a plan been thwarted due to _reasons_. :-)

I'll review 2822 and 2231 (and 2047, 5322, ...). I know this is a **complicated** topic, so I'll do some more research and open an issue on it for discussion.

Maybe a first step would be not to unfold at all, and to allow the decoder(s) to process the headers before unfolding the lines.

baudehlo:
> Maybe a first step would be not to unfold at all, and to allow the
decoder(s) to process the headers before unfolding the lines.

Yes that would probably be wise - so headers_decoded would use an
appropriate parser (which unfolds), but undecoded it would still be
unfolded.

The only problem then is that X- headers can be folded, and there's no
pre-defined grammar for those (AFAIK).

PR #2231, opened 2017-10-19
hmail, todo read/write tests

msimerson:
- hmail: rename tl_reader -> length_stream
- hmail: rename td_reader -> header_stream
- replace [deprecated](https://nodejs.org/api/buffer.html) `new Buffer` with `Buffer.from` 
- hmail: before attempting to parse JSON, check that the last char is a `}`
    - avoid parsing exception w/o more expensive try/catch
- outbound: in hmail files, wrap the JSON header in \n chars (human friendly++)
- outbound.build_todo: replace new `Buffer.from(todo_json).length` with `Buffer.byteLength(todo_str)`
- outbound.build_todo: replace deprecated `new Buffer(size)` with `Buffer.alloc(size)`
- outbound.build_todo: replace `ws.write(buf);` with `ws.write(buf, 'utf8', write_more);` (use callback signature)
- add hmail tests

Checklist:
- [ ] docs updated
- [x] tests updated


outbound/hmail.js
```diff
@@ -87,29 +87,36 @@ class HMailItem extends events.EventEmitter {
...
+        length_stream.on('end', () => {
+            const todo_len = Buffer.from(todo_len_raw).readUInt32BE(0);
+            self.logdebug(`todo header length: ${todo_len}`);
+            const header_stream = fs.createReadStream(self.path, {encoding: 'utf8', start: 4, end: todo_len + 3});
             self.data_start = todo_len + 4;
 
             let todo_raw = '';
-            td_reader.on('data', (data) => { todo_raw += data; });
-            td_reader.on('end', () => {
+            header_stream.on('data', (data) => { todo_raw += data; });
```
baudehlo: We should come up with a better name - perhaps `json_data_stream`? Especially since `header` is such a loaded term in email.

outbound/hmail.js
```diff
@@ -87,29 +87,36 @@ class HMailItem extends events.EventEmitter {
...
+            header_stream.on('end', () => {
                 if (Buffer.byteLength(todo_raw) === todo_len) {
                     // we read everything
+                    todo_raw = todo_raw.trim()
```
baudehlo: Is there any situation this might happen or is it just belt and braces?

outbound/hmail.js
```diff
@@ -87,29 +87,36 @@ class HMailItem extends events.EventEmitter {
...
+                    const last_char = todo_raw.charAt(todo_raw.length - 1);
+                    if (last_char !== '}') {
+                        self.emit('error', `invalid todo header end char: ${last_char} at pos ${todo_len} of ${self.filename}`)
+                        return
+                    }
                     self.todo = JSON.parse(todo_raw);
-                    self.todo.rcpt_to = self.todo.rcpt_to.map(a => { return new Address (a); });
                     self.todo.mail_from = new Address (self.todo.mail_from);
+                    self.todo.rcpt_to = self.todo.rcpt_to.map(a => { return new Address (a); });
```
baudehlo: This seems moved unnecessarily. Also it's a "useless use of return" - if we're going to change that line at least change the afunc to `a => new Address(a)`.

outbound/index.js
```diff
@@ -313,29 +313,27 @@ exports.process_delivery = function (ok_paths, todo, hmails, cb) {
...
-    function exclude_from_json (key, value) {
-        switch (key) {
-            case 'message_stream':
-                return undefined;
-            default:
-                return value;
-        }
-    }
 
-    const todo_json = JSON.stringify(todo, exclude_from_json)
```
baudehlo: Let's just pass `'\t'` as the third param to JSON.stringify instead.

outbound/index.js
```diff
@@ -313,29 +313,27 @@ exports.process_delivery = function (ok_paths, todo, hmails, cb) {
...
+    const buf = Buffer.alloc(4 + todo_len);
     buf.writeUInt32BE(todo_len, 0);
-    buf.write(todo_json, 4);
-
-    const continue_writing = ws.write(buf);
-    if (continue_writing) return process.nextTick(write_more);
+    buf.write(todo_str, 4);
 
-    ws.once('drain', write_more);
+    ws.write(buf, 'utf8', write_more);
```
baudehlo: The semantics for `drain` are different to using the callback to write. This is an incorrect change. It's a very fringe condition where it happens, but using drain is the right thing to do (I've studied this code a lot to make sure it's correct).

baudehlo:
I know this is still WIP but I wanted to review this anyway :)

outbound/hmail.js
msimerson: You'd think that would work (I sure did) but when I tried that, it didn't! We do have a lot of "suboptimal" uses of return I'd like to see cleaned up (example: when return isn't returning a value but is being used solely for flow control, use a bare return on its own line), so that returns that are doing something different LOOK different.

baudehlo:
It has to be without the curlies - did you try that?

outbound/index.js
msimerson: Do you know of any conditions where that's required? I'd love to get it covered by tests.

outbound/hmail.js
msimerson: I don't love that name either. json and data are even more generic. For now, I've changed to todo_json_stream.

outbound/hmail.js
msimerson: It removes the leading \n characters (in a backwards compatible way) added here:

```js
const todo_str = `\n${JSON.stringify(todo, exclude_from_json)}\n`
```

outbound/index.js
msimerson: That doesn't do nearly the same thing. Although, passing in a `\t` as the third option does make the queue file that even more human friendly, so I'm happy to do that too. The `\n`'s put the first Received header onto a newline, after the JSON header.

msimerson:
Example queue file, after adding the leading and trailing `\n` and adding `\t` as the third arg to JSON.stringify (the first line is binary):

````
"
{
	"queue_time": 1507509981169,
	"domain": "redacteed.com",
	"rcpt_to": [
		{
			"original": "<postmaster@redacteed.com>",
			"original_host": "redacteed.com",
			"host": "redacteed.com",
			"user": "postmaster"
		}
	],
	"mail_from": {
		"original": "<matt@tnpi.net>",
		"original_host": "tnpi.net",
		"host": "tnpi.net",
		"user": "matt"
	},
	"notes": {
		"authentication_results": [
			"spf=pass smtp.mailfrom=tnpi.net"
		],
		"spf_mail_result": "Pass",
		"spf_mail_record": "v=spf1 a mx include:mx.theartfarm.com ?include:forwards._spf.tnpi.net include:lists._spf.tnpi.net -all",
		"attachment_count": 0,
		"attachments": [
			{
				"ctype": "application/pdf",
				"filename": "FileWîthÁccent Chars.pdf",
				"extension": ".pdf",
				"md5": "6c1d5f5c047cff3f6320b1210970bdf6"
			}
		],
		"attachment_ctypes": [
			"application/pdf",
			"multipart/mixed",
			"text/plain",
			"application/pdf"
		],
		"attachment_files": [
			"FileWîthÁccent Chars.pdf"
		],
		"attachment_archive_files": []
	},
	"uuid": "1D5483B0-3E00-4280-A961-3AFD2017B4FC.1"
}
Received: (Haraka outbound); Sun, 08 Oct 2017 20:46:21 -0400
Authentication-Results: qmail03.redacteed.com; iprev=pass; spf=pass smtp.mailfrom=tnpi.net
Received-SPF: Pass (qmail03.redacteed.com: domain of tnpi.net designates 66.128.51.165 as permitted sender) receiver=qmail03.redacteed.com; identity=mailfrom; client-ip=66.128.51.165; helo=mail.theartfarm.com; envelope-from=<matt@tnpi.net>
Received-SPF: Pass (qmail03.redacteed.com: domain of mail.theartfarm.com designates 66.128.51.165 as permitted sender) receiver=qmail03.redacteed.com; identity=helo; client-ip=66.128.51.165; helo=mail.theartfarm.com; envelope-from=<matt@tnpi.net>
Received: from mail.theartfarm.com (mail.theartfarm.com [66.128.51.165])
        by qmail03.redacteed.com (Haraka/2.8.16) with ESMTPS id 1D5483B0-3E00-4280-A961-3AFD2017B4FC.1
        envelope-from <matt@tnpi.net>
        (version=TLSv1/SSLv3 cipher=ECDHE-RSA-AES128-GCM-SHA256 verify=FAIL);
        Sun, 08 Oct 2017 20:46:12 -0400
Received: (Haraka outbound); Sun, 08 Oct 2017 17:46:06 -0700
Received-SPF: Pass (mail.theartfarm.com: domain of tnpi.net designates 174.24.224.63 as permitted sender) receiver=mail.theartfarm.com; identity=mailfrom; client-ip=66.128.51.165; helo=imac27.simerson.net; envelope-from=<matt@tnpi.net>
Received-SPF: None (mail.theartfarm.com: domain of imac27.simerson.net does not designate 174.24.224.63 as permitted sender) receiver=mail.theartfarm.com; identity=helo; client-ip=174.24.224.63; helo=imac27.simerson.net; envelope-from=<matt@tnpi.net>
Received: from imac27.simerson.net (174-24-224-63.tukw.qwest.net [174.24.224.63])
        by mail.theartfarm.com (Haraka/2.8.15) with ESMTPSA id 48774D25-5868-4812-B153-A2356C7785E1.1
        envelope-from <matt@tnpi.net> (authenticated bits=0)
        (version=TLSv1/SSLv3 cipher=ECDHE-RSA-AES256-GCM-SHA384 verify=FAIL);
        Sun, 08 Oct 2017 17:46:01 -0700
From: Matt Simerson <matt@tnpi.net>
Content-Type: multipart/mixed;
        boundary="Apple-Mail=_415A8F91-AD3D-4BD0-857C-B2B3EFCF68B5"
Subject: from matt to matt
Message-Id: <1F7209D5-782F-4EF2-92D7-94BFEEEFCCCA@tnpi.net>
Date: Sun, 8 Oct 2017 17:45:51 -0700
To: Post Mattster <postmaster@redacteed.com>
X-Mailer: Apple Mail (2.3445.1.7)
X-Haraka-ASN: 209
X-Haraka-Karma: score: 21, good: 8, bad: 2, connections: 10, history: 6, awards: 133,162,182, pass:relaying
DKIM-Signature: v=1;a=rsa-sha256;bh=bSMRYkUxvOcMB5ccXO41lmCTr6iJqFhhazaFWxo2hGE=;c=relaxed/simple;d=tnpi.net;h=from:subject:date:message-id:to:mime-version;s=mar2013;b=aA1NQDQvIunHNG6mY1m3p3fjiioLWTdM5MfgX/wZ9Nnj/Do9n0I+W9wQcs/kn7ES+f5TvbwZhnwigMaSX8mbjkyXiWf51gyxXVc0/4R+zDwoOFmmGNNWgs+ZVOKNtBLglNyoytTvD1cSfWgiiP+w5zIMNQjbhG9RUf5xLAx4zZI1wVbBZ1jvc1VeiApbKkRkKTftY/i1OwcFGjFXSvayiHZqZlWHnISewTR/t79Dr2juY2U0qwwtMgKkd5qVFGpBs6uyNd+SDsqQ1Pl4wArWJFajKa0Mg6kBiBEa4p9WheEr8FKqglHW/NfUGmFAy5h6+rdLomUcM42mMeM8P74eXQ==
Original-Authentication-Results: mail.theartfarm.com; iprev=pass; auth=pass (plain); spf=pass smtp.mailfrom=tnpi.net
X-Haraka-GeoIP: NA, US, TX, Southlake
X-Haraka-GeoIP-Received: 66.128.51.165:US,174.24.224.63:US


this is a sample message body.

````

outbound/hmail.js
msimerson: > This seems moved unnecessarily

True, it's not necessary at all. But it does arrange it in SMTP order, with feels better. :-)

baudehlo: This got lost by replying via email, but did you try without curlies? They change the semantics.

msimerson:
> It has to be without the curlies - did you try that?

Hadn't seen **that** syntax yet. It works. Thanks for the tip.

outbound/index.js
baudehlo: I know what it does, but I figured if we want human readable, the \t makes it so much better.

outbound/hmail.js
baudehlo: Oh - I was thinking "this isn't required for JSON.parse to work" but it's really for the last character check more than anything.

outbound/hmail.js
baudehlo: Yeah that works. I guess `todo` isn't a great name either (stolen from qmail, obv) but it's everywhere in the outbound code so ...

outbound/index.js
baudehlo: See highWaterMark in https://nodejs.org/api/stream.html#stream_constructor_new_stream_writable_options

We can lower the value to configure the event.

msimerson:
PS: the reason for this PR is that there's **still** an issue with the files getting written out in a way that's not getting read back in correctly, and then crashing Haraka. The tests are my way of wrapping my head around this code that until recently, I avoided using. The changes we committed recently (my last PR) helped in at least one case, but things still are not quite right since the PR that added *UInt32BE().

baudehlo:
Totally fair to make these changes - Outbound is badly tested and badly
maintained. I take full responsibility but it definitely needs more eyes.

msimerson:
I found the root cause of the issue: string concatenation is A Bad Idea[TM] when handling binary data. The fix is in [this commit](https://github.com/haraka/Haraka/pull/2231/commits/2ed8cccf98b96499ecceb9ae1a9e2edb981d9eee).

PR #2238, opened 2017-10-21
Hmail refactor

msimerson:
- encapsulate the streaming logic into  _stream_bytes_from
- with some safety rails

### DIFF
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -86,53 +86,68 @@ class HMailItem extends events.EventEmitter {
     }
 
     read_todo () {
-        const self = this;
-        const length_stream = fs.createReadStream(self.path, {start: 0, end: 3});
+        this._stream_bytes_from(this.path, {start: 0, end: 3}, (err, bytes) => {
+            if (err) {
+                const errMsg = `Error reading queue file ${this.filename}: ${err}`;
+                this.logerror(errMsg);
+                this.temp_fail(errMsg);
+                return
+            }
 
-        length_stream.on('error', (err) => {
-            const errMsg = `Error reading queue file ${self.filename}: ${err}`;
-            self.logerror(errMsg);
-            self.temp_fail(errMsg);
-        });
+            const todo_len = bytes.readUInt32BE(0);
+            this.logdebug(`todo header length: ${todo_len}`);
+            this.data_start = todo_len + 4;
 
-        let todo_len_raw = Buffer.alloc(0);
-        length_stream.on('data', (data) => {
-            todo_len_raw = Buffer.concat([todo_len_raw, data])
-        })
-        length_stream.on('end', () => {
-            const todo_len = todo_len_raw.readUInt32BE(0);
-            self.logdebug(`todo header length: ${todo_len}`);
-            const todo_json_stream = fs.createReadStream(self.path, {encoding: 'utf8', start: 4, end: todo_len + 3});
-            self.data_start = todo_len + 4;
-
-            let todo_raw = '';
-            todo_json_stream.on('data', (data) => { todo_raw += data; });
-            todo_json_stream.on('end', () => {
-                if (Buffer.byteLength(todo_raw) === todo_len) {
-                    // we read everything
-                    todo_raw = todo_raw.trim()
-                    const last_char = todo_raw.charAt(todo_raw.length - 1);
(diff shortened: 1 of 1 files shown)


msimerson:
This PR is kind of a question. Part of the reason I found `read_todo()` so difficult to reason about was all the semantics involving fs.createReadStream, Buffers, and encoding. And poring over the details of the docs for each. I feel like encapsulating the streaming in a consistent way (strictly as bytes) provides a simpler to reason about `read_todo()`. Thoughts?

msimerson:
PS: the diff is crap to look at, so there's the **after** version of those functions:

```js
    read_todo () {
        this._stream_bytes_from(this.path, {start: 0, end: 3}, (err, bytes) => {
            if (err) {
                const errMsg = `Error reading queue file ${this.filename}: ${err}`;
                this.logerror(errMsg);
                this.temp_fail(errMsg);
                return
            }

            const todo_len = bytes.readUInt32BE(0);
            this.logdebug(`todo header length: ${todo_len}`);
            this.data_start = todo_len + 4;

            this._stream_bytes_from(this.path, {start: 4, end: todo_len + 3}, (err2, todo_bytes) => {
                if (todo_bytes.length !== todo_len) {
                    const wrongLength = "Didn't find right amount of data in todo!"
                    this.logcrit(wrongLength);
                    fs.rename(this.path, path.join(queue_dir, "error." + this.filename), (err3) => {
                        if (err3) {
                            this.logerror("Error creating error file after todo read failure (" + this.filename + "): " + err);
                        }
                    });
                    this.emit('error', wrongLength); // Note nothing picks this up yet
                    return
                }

                // we read everything
                const todo_json = todo_bytes.toString().trim()
                const last_char = todo_json.charAt(todo_json.length - 1);
                if (last_char !== '}') {
                    this.emit('error', `invalid todo header end char: ${last_char} at pos ${todo_len} of ${this.filename}`)
                    return
                }
                this.todo = JSON.parse(todo_json);
                this.todo.mail_from = new Address (this.todo.mail_from);
                this.todo.rcpt_to = this.todo.rcpt_to.map(a => new Address (a));
                this.todo.notes = new Notes(this.todo.notes);
                this.emit('ready');
            });
        });
    }

    _stream_bytes_from (file_path, opts, done) {
        if (opts.encoding !== undefined) {
            // passing an encoding to fs.createReadStream will change the type of data returned
            // ex: instead of returning a buffer, it may return a String, which will cause
            // Buffer.concat to barf. There's a reason this function has 'bytes' in the name
            done(new Error("Thar be dragons here! Encode/decode on the result of this function"))
            return
        }

        const stream = fs.createReadStream(file_path, opts);

        stream.on('error', done)

        let raw_bytes = Buffer.alloc(0);
        stream.on('data', (data) => {
            raw_bytes = Buffer.concat([raw_bytes, data])
        })

        stream.on('end', () => {
            done(null, raw_bytes)
        })
    }
```

baudehlo:
One thing this refactor should be able to do: Instead of re-opening the
file every time maybe we can pass an FD to createReadStream instead?

I haven't read it in detail yet - will do.

msimerson:
> maybe we can pass an FD to createReadStream instead

It's not a bad idea, but I highly doubt it would make **any** difference at all. Once you open the file the first time, the FD is going to be in the OS and also likely the disk drive(s) cache. More than likely, the size and JSON header will be in the same disk block anyway. The only real reason I could see for doing that would be if it makes the code cleaner / more succinct, and that's something definitely doesn't do.

baudehlo:
The reason I thought of doing it was it helps reduce the number of used
file descriptors. Or maybe it doesn't actually because the file would be
closed before the next part is processed?

It also reduces syscalls.

msimerson:
> Or maybe it doesn't actually because the file would be closed before the next part is processed?

It doesn't because they are currently sequential. As is `data_stream()`. Managing FDs ourselves. It adds complexity to the code as we'd have to follow every code path and make sure they're all closed properly. If our outbound code was robust AND we had decent test coverage AND we tested and found managing the FDs ourself provided a performance boost, **then** I'd be +1 on passing in the FD. Until then, to me it feels like a premature optimization.

baudehlo:
Yep makes sense - I just wanted to discuss it.

PR #2246, opened 2017-10-23
Use template literals in logger.js

rhedshi:
Fixes #2129 

- Use template literals in logger.js

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -180,7 +180,7 @@ logger.set_loglevel = function (level) {
 
     const loglevel_num = parseInt(level);
     if (typeof level === 'string') {
-        this.log('INFO', 'loglevel: ' + level.toUpperCase());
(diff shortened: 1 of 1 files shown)


logger.js
```diff
@@ -304,9 +304,9 @@ logger.log_if_level = function (level, key, plugin) {
                 logger.log(
                     level,
                     [
-                        '[' + logobj.level + ']',
-                        '[' + logobj.uuid + ']',
-                        '[' + logobj.origin + ']',
+                        `[${logobj.level}]`,
+                        `[${logobj.uuid}]`,
+                        `[${logobj.origin}]`,
```
msimerson: This is good, but maybe better if we drop the array.join here and have just a single literal?

msimerson:
Doh, lint error. Run `npm run lint` or see Travis output.

PR #2249, opened 2017-10-23
Use template literals in outbound/index.js

m-ke:
#2129 

- Replace string concatenations with template literals in outbound/index.js

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/outbound/index.js
+++ b/outbound/index.js
@@ -220,14 +220,14 @@ exports.send_trans_email = function (transaction, next) {
     // add potentially missing headers
     if (!transaction.header.get_all('Message-Id').length) {
         logger.loginfo("[outbound] Adding missing Message-Id header");
-        transaction.add_header(`Message-Id<${transaction.uuid}@${config.get('me')}>`);
+        transaction.add_header('Message-Id', `<${transaction.uuid}@${config.get('me')}>`);
     }
     if (!transaction.header.get_all('Date').length) {
         logger.loginfo("[outbound] Adding missing Date header");
         transaction.add_header('Date', utils.date_to_str(new Date()));
     }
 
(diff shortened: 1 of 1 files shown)


outbound/index.js
```diff
@@ -220,14 +220,14 @@ exports.send_trans_email = function (transaction, next) {
     // add potentially missing headers
     if (!transaction.header.get_all('Message-Id').length) {
         logger.loginfo("[outbound] Adding missing Message-Id header");
-        transaction.add_header('Message-Id', '<' + transaction.uuid + '@' + config.get('me') + '>');
+        transaction.add_header(`Message-Id<${transaction.uuid}@${config.get('me')}>`);
```
msimerson: oops! The call signature for add_header requires two arguments. It needs to look like this instead:
```js
transaction.add_header('Message-Id', `<${transaction.uuid}@${config.get('me')}>`);
```

outbound/index.js
```diff
@@ -220,14 +220,14 @@ exports.send_trans_email = function (transaction, next) {
...
     }
     if (!transaction.header.get_all('Date').length) {
         logger.loginfo("[outbound] Adding missing Date header");
         transaction.add_header('Date', utils.date_to_str(new Date()));
     }
 
-    transaction.add_leading_header('Received', '('+cfg.received_header+'); ' + utils.date_to_str(new Date()));
+    transaction.add_leading_header(`Received(${cfg.received_header}); ${utils.date_to_str(new Date())}`);
```
msimerson: same here.

outbound/index.js
m-ke: Ah, I overlooked the comma. I'll push the fix in a bit.

PR #2259, opened 2017-10-24
Use template literals in outbound/hmail.js

m-ke:
There's one thing I wasn't sure about - long strings that were split into multiple lines. I joined them into one line since leaving string concatenation in seemed to defeat the purpose. Let me know if you'd prefer me to split them. 

Checklist:
- [ ] docs updated
- [ ] tests updated


msimerson:
+1 for long lines!

PR #2269, opened 2017-10-26
SMTPS port

### LINKED ISSUE #2268: Make the SMTPS port configurable
Currently, Haraka considers only port 465 as the SMTPS port. Since port 465 is reserved for root, this means that Haraka requires root permissions to provide SMTPS support. Making this port configurable will allow Haraka to provide SMTPS support when run as normal user.

gramakri:
Fixes #2268

- SMTPS port is now configurable using the `smtps_port` option in smtp.ini
- Fix get_smtp_server test (it is incorrectly passing now)

Checklist:
- [ X ] docs updated
- [ X ] tests updated
- [ X ] [Changes](https://github.com/haraka/Haraka/blob/master/Changes.md) updated

### DIFF
--- a/docs/CoreConfig.md
+++ b/docs/CoreConfig.md
@@ -23,8 +23,8 @@ The list of plugins to load
 
   Keys:
 
-  * port - the port to use (default: 25)
-  * listen\_address - default: 0.0.0.0 (i.e. all addresses)
+  * listen\_host, port - the host and port to listen on (default: ::0 and 25)
+  * listen - (default: [::0]:25) Comma separated IP:Port addresses to listen on
   * inactivity\_time - how long to let clients idle in seconds (default: 300)
   * nodes - if [cluster][1] is available, specifies how
     many processes to fork off. Can be the string "cpus" to fork off as many
--- a/server.js
+++ b/server.js
@@ -362,6 +362,7 @@ Server.get_smtp_server = function (host, port, inactivity_timeout, done) {
     }
     else {
         server = tls_socket.createServer(onConnect);
+        server.has_tls = false;
         Server.listeners.push(server);
         done(server);
     }
--- a/server.js
+++ b/server.js
@@ -41,6 +41,7 @@ Server.load_smtp_ini = function () {
         daemon_log_file: '/var/log/haraka.log',
         daemon_pid_file: '/var/run/haraka.pid',
         force_shutdown_timeout: 30,
+        smtps_port: 465,
     };
 
     for (const key in defaults) {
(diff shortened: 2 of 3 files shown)


msimerson:
SMTPS is deprecated since 1997.  No modern email clients use it by default (although several support it). SMTPS is rarely used in production and we have purposefully avoided documenting its support. It exists solely for backwards compatibility and **nobody** should be encouraged to use it. Adding docs and config settings to customize it feels like encouragement to me.

Do you use SMTPS?  What for?

gramakri:
@msimerson This is used internally in the Cloudron to appease some of our apps (that we have no control over).

Don't mean to confuse you with Cloudron specific details but I have added more information below. Would you be OK to simply leave this option undocumented? (I do agree with your stance on SMTPS that it's probably not a good idea to go out of the way to show "support" for it).

<cloudron specific>
We use Haraka as a mail relay for apps. Apps have to authenticate to relay email.

There is a bunch of apps/frameworks (e.g PHP based) that cannot work with self-signed certs. This led us to disabling starttls wholesale for all apps (this is why I added `no_tls_hosts` previously). We still need authentication. TLS is not required for us since the network is 'safe'.

This worked well so far until last month when [Go decided](https://github.com/golang/go/commit/ec3b6131de8f9c9c25283260c95c616c74f6d790#diff-7ec6f47619a11e762f94e4766283210d) not to send auth information without TLS. This means we have to enable TLS for some apps now.

One solution is to exclude specific apps from `no_tls_hosts` by IP but this gets complicated because we have to track IP of apps etc. The other easy solution was to simply expose a new SMTPS port and make Go apps use that. Another approach that I considered and did not like is to possibly create something like `no_tls_ports` and make Go apps connect to those ports for relaying. I don't like that option because over time, we will ideally migrate over all apps to use TLS and obsolete our use of `no_tls_hosts` as well.

Haraka also runs as non-root on Cloudron, so we have to use some other port for SMTPS.
</cloudron specific>

msimerson:
> Would you be OK to simply leave this option undocumented?

Yes

gramakri:
@msimerson Thanks, I have updated the PR to remove the docs and the note in Changes.md.

KingNoosh:
I don't understand the documentation changes currently. What was the reasoning? It looks like you've changed the naming of keys too in the docs?

server.js
```diff
@@ -362,6 +363,7 @@ Server.get_smtp_server = function (host, port, inactivity_timeout, done) {
     }
     else {
         server = tls_socket.createServer(onConnect);
+        server.has_tls = false;
```
KingNoosh: Is it still not a tls server at this point?

docs/CoreConfig.md
```diff
@@ -23,8 +23,8 @@ The list of plugins to load
 
   Keys:
 
-  * port - the port to use (default: 25)
-  * listen\_address - default: 0.0.0.0 (i.e. all addresses)
+  * listen\_host, port - the host and port to listen on (default: ::0 and 25)
```
KingNoosh: Where has listen_host and listen come from? Where has listen_address gone?

gramakri: @KingNoosh This change is fixing incorrect docs. It is fixing up what is in code with the docs.

The correct config is listen and not listen_address - https://github.com/haraka/Haraka/blob/master/server.js#L258. You seem to have missed the line that adds listen.

listen_host comes from https://github.com/haraka/Haraka/blob/master/server.js#L267. This is used in conjunction with port.

server.js
gramakri: tls_socket is a socket that can potentially upgrade to TLS (if and when starttls gets issued). doesn't mean it's already using TLS. in the above case, it's not upgrade to tls (yet).

PR #2275, opened 2017-10-29
Removed trailing spaces

lnedry:
Removed trailing spaces

### DIFF
--- a/plugins.js
+++ b/plugins.js
@@ -42,13 +42,13 @@ class Plugin {
         const plugin = this;
         /* From https://github.com/haraka/Haraka/pull/1278#issuecomment-168856528
         In Development mode, or install via a plain "git clone":
-    
+
             Plain plugin in plugins/ folder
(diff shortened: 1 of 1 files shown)


plugins.js
```diff
@@ -174,7 +174,7 @@ class Plugin {
                 return require(module);
             }
 
-            if (fs.existsSync(path.join(__dirname, module + '.js')) ||
+            if (fs.existsSync(path.join(`${__dirname, module}.js`)) ||
```
msimerson: this is incorrect.

plugins.js
```diff
@@ -492,7 +491,7 @@ plugins.run_next_hook = function (hook, object, params) {
         timeout_id = setTimeout(function () {
             timed_out = true;
             object.logcrit(`Plugin ${item[0].name} timed out on hook ${hook}` +
-                    ' - make sure it calls the callback');
+                    ` - make sure it calls the callback`);
```
msimerson: rather than concatenate two string literals, maybe condense? (above too)

PR #2298, opened 2017-11-13
Fix smtp_client error handling

smfreegard:
I found this issue when tracing a problem in some code where I was using smtp_client.

I had a case of an SMTP host that was down, but my `smtp_client.on('error')` code block was not being called and therefore `next()` wasn't being called, so the plugin always timed out instead of being correctly trapped.

The issue was that the `switch()` block would always call `client.destroy()` which would in turn set the state to `STATE.DESTROYED` so the `if (client.state === STATE.ACTIVE)` would never ever be true.

I've therefore re-ordered the statements and removed the return as calling `client.destroy()` should not do any other harm.

This fixed my issue.

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/smtp_client.js
+++ b/smtp_client.js
@@ -157,9 +157,12 @@ class SMTPClient extends events.EventEmitter {
                     error = '';
                 }
                 // msg is e.g. "errored" or "timed out"
-                // error is e.g. "Error: connect ECONNREFUSE"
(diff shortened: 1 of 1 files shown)


msimerson:
This looks fine to me, but I haven't found time to delve into the code and understand the change. If you're still running with this in production and it's worked with no side effects this long, it's probably the right fix and +1 from me.

PR #2310, opened 2017-12-11
Adding ENOTFOUND to also check A record

analogic:
I've found problem when working with Docker DNS system - please see https://github.com/docker/libnetwork/issues/2026 

Since docker implements only basic records, it responds to any MX request with NXDOMAIN (ENOTFOUND at node) so it is basicaly impossible to create two communicating containers with haraka. It is bug of Docker only (I have not seen any other DNS servers doing same thing) so please close if you think Haraka should remain strict at this...

Checklist:
- [ ] docs updated
- [ ] tests updated


smfreegard:
Thanks for the patch - this is absolutely the right thing to do anyway IMO.

PR #2314, opened 2017-12-28
Add "origin" option to outbound.send_mail for better bounce tracking

analogic:
Right now when bounce is generated there is no easy way to connect newly generated transaction with original so I've added new option to send_email that will be passed to log calls.

- new "origin" option at outbound.send_email
- updated HMailItem to generate bounce with this option

Checklist:
- [x] docs updated
- [ ] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -65,6 +65,7 @@
         * outbound/hmail #2197
     * Automatically set connection.remote.is_private when connection.remote.ip is set #2192
     * Add remove_msgid and remove_date options to outbound.send_email #2209
+    * Add origin option to outbound.send_mail #2314
 
 
 ## 2.8.16 - Sep 30, 2017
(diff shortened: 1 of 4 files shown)


analogic:
I am sorry I don't understand question. Do you mean it should be required?

msimerson:
Not required, my only question is if this should be "on by default." I think yes, but wanted to solicit additional opinions.

superman20:
I'd be a fan of having it on by default.  I have a personal goal in 2018 of making the entire mail delivery chain traceable in the logs and this would certainly help that effort.

smfreegard:
+1 for default from me.  I would go as far as saying it needn't be switchable.

analogic:
Understood, I have same goal as @superman20 also

PR #2317, opened 2018-01-18
Fix outbound pools

KingNoosh:
Fixes #

- Fix outbound pooling

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/outbound/client_pool.js
+++ b/outbound/client_pool.js
@@ -8,13 +8,18 @@ const logger       = require('../logger');
 
 const cfg          = require('./config');
 
-function _create_socket (port, host, local_addr, is_unix_socket, callback) {
+let socket_idx = 0;
+
+function _create_socket (pool_name, port, host, local_addr, is_unix_socket, callback) {
     const socket = is_unix_socket ? sock.connect({path: host}) :
         sock.connect({port: port, host: host, localAddress: local_addr});
+    socket.__pool_name = pool_name;
+    socket.__sock_idx = ++socket_idx;
     socket.setTimeout(cfg.connect_timeout * 1000);
     logger.logdebug(
         '[outbound] created',
         {
+            idx: socket_idx,
             host: host,
             port: port,
             pool_timeout: cfg.pool_timeout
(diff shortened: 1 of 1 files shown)


outbound/client_pool.js
```diff
@@ -8,28 +8,30 @@ const logger       = require('../logger');
...
 const cfg          = require('./config');
 
-function _create_socket (port, host, local_addr, is_unix_socket, callback) {
+let socket_idx = 0;
+
+function _create_socket (pool_name, port, host, local_addr, is_unix_socket, callback) {
     const socket = is_unix_socket ? sock.connect({path: host}) :
         sock.connect({port: port, host: host, localAddress: local_addr});
+    socket.__pool_name = pool_name;
+    socket.__sock_idx = ++socket_idx;
```
baudehlo: Since this is just for logging, is there any reason we shouldn't use a UUID for this instead?

msimerson: Wouldn't a UUID be confusing? If the outbound event emits a UUID, I'd expect that UUID to be correlated with whatever event landed the message in the queue. What do you gain from having the pool index number be a UUID?

KingNoosh: uuid would be useful when searching through the logs and looking at any other messages associated I believe

msimerson: Associated _messages_? My reading makes it seem like you'd only get associated delivery attempts by a particular pool socket. That association already exists with the host:port, so I'm not seeing what a UUID gains us.

anosh-plutoflume: Let's say a certain pool socket ends in a really funky way, you want to see what that particular socket did before the funky action happened. This is what we use it for. Just incase.

anosh-plutoflume: I'm fine w/ either way tbh, it makes no difference for me.

I don't like using an incrementing number but it works atm.

KingNoosh: Let's say a certain pool socket ends in a really funky way, you want to see what that particular socket did before the funky action happened. This is what we use it for. Just incase.

I'm fine w/ either way tbh, it makes no difference for me.
I don't like using an incrementing number but it works atm.

I have a bad habit from commenting from my company account. Sorry for spam.

msimerson: > Let's say a certain pool socket ends in a really funky way

That's the part I understand. We **already** have that association with host:port, adding the index number lets you determine that one particular socket is funky. What I'm not seeing is the advantage of sticking a really long UUID in the log messages adds.

anosh-plutoflume: Ah, so some people run their multiple instances of haraka in multiple containers and log to a central place, that host:port can be consistent across the instances but the UUID will always be unique.

KingNoosh: Ah, so some people run their multiple instances of haraka in multiple containers and log to a central place, that host:port can be consistent across the instances but the UUID will always be unique.

I'm going to keep this Firefox tab open now.

msimerson: :blink:, and his eyes were opened.

baudehlo: Also an integer could overflow into floating point :-)

I feel like an integer looks too magical, whereas a UUID makes people think "Ah this is just for tracking".

baudehlo:
Approved - @msimerson or @smfreegard any objections to merging it?

PR #2320, opened 2018-01-21
add openssl-wrapper as dependency

msimerson:
- used in tls_socket.js. Required by the TLS cert dir that extracts cert data from the .pem files.
- it should have come over with the TLS consolidation to tls_socket.js
- syntax cleanups in tls_socket.js


msimerson:
That’s another of those, “instead of assigning an anonymous function to a named variable, just create a named function” updates. We should be doing it that way most everywhere due to variable hoisting differences in const/let vs var.

PR #2323, opened 2018-01-22
add this.removeAllListeners('connection-error')

### LINKED ISSUE #2312: Memory leak from smtp_client
I'm working on a plugin that's modified based on smtp_forward, and haraka crashed under stress test. I believe it's caused by the smtp_client.release() method didn't remove listeners for event connection-error, resulting in existing transactions cannot be gc'ed. Lots of errors like this logged:
(node:7329) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 connection-error listeners added. Use emitter.setMaxListeners() to increase limit

After adding smtp_client.removeAllListeners('connection-error') after calling smtp_client.release() I no longer see memory leak.

msimerson:
- fixes #2312
- arrow functions

### DIFF
--- a/smtp_client.js
+++ b/smtp_client.js
@@ -50,7 +50,7 @@ class SMTPClient extends events.EventEmitter {
 
         const client = this;
 
-        client.socket.on('line', function (line) {
+        client.socket.on('line', (line) => {
             client.emit('server_protocol', line);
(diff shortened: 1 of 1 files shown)


Dexus:
LGTM

PR #2324, opened 2018-01-22
replace _ chars in hostnames with code points

### LINKED ISSUE #2225: Haraka throws exception when local hostname includes underscores
Haraka generates queue filenames based on [hostname](https://github.com/haraka/Haraka/blob/38d62f545d73653ede0abc950ae24757c928b8ea/outbound/qfile.js#L3) and joins separate parts with underscores. If the local hostname is configured to include underscore as well (eg "example_host" – not sure if it is even valid but it happens) then this underscore also gets included in the filename which results in a failed [filename check](https://github.com/haraka/Haraka/blob/38d62f545d73653ede0abc950ae24757c928b8ea/outbound/qfile.js#L53-L56) which in turn is causes an [exception to be thrown](https://github.com/haraka/Haraka/blob/38d62f545d73653ede0abc950ae24757c928b8ea/outbound/hmail.js#L52). Easy fix would be to replace underscore with a codepoint as is done with "/" and ":" but maybe something more thorough should be considered? Like replacing all non-alphanumeric chars with codepoints?

msimerson:
- fixes #2225
- also perform replacements on passed in host values
- add outbound/qfile tests

### DIFF
--- a/outbound/qfile.js
+++ b/outbound/qfile.js
@@ -1,12 +1,20 @@
-"use strict";
+'use strict';
 
-const my_hostname = require('os').hostname().replace(/\\/, '\\057').replace(/:/, '\\072');
+const os = require('os');
 const platform_dot = ((['win32','win64'].indexOf( process.platform ) !== -1) ? '' : '__tmp__') + '.';
 
 let QFILECOUNTER = 0;
 
 const _qfile = module.exports = {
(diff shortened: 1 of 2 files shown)


Dexus:
LGTM

qhxin:
i think that the replace pattern should use /g expression maybe?
image

msimerson:
Indeed. A PR is welcome.

PR #2329, opened 2018-02-01
Prevent data headers crit fail

KingNoosh:
Fixes #

- 
- 

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/plugins/data.headers.js
+++ b/plugins/data.headers.js
@@ -286,13 +286,13 @@ exports.from_match = function (next, connection) {
 
     try {
         const hdr_addr = (plugin.addrparser.parse(hdr_from))[0];
+        if (!hdr_addr) {
+            plugin.loginfo(`address at fault is: ${hdr_from}`);
+            connection.transaction.results.add(plugin, {fail: 'from_match(unparsable)'});
+            return next();
+        }
     } catch (error) {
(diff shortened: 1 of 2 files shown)


plugins/data.headers.js
```diff
@@ -284,8 +284,13 @@ exports.from_match = function (next, connection) {
         return next();
     }
 
-    const hdr_addr = (plugin.addrparser.parse(hdr_from))[0];
+    try {
+        const hdr_addr = (plugin.addrparser.parse(hdr_from))[0];
```
baudehlo: I don't think this works because hdr_addr now gets block-scoped here. You need to define it with `let` outside the try and assign inside.

msimerson: > I don't think this works

Correct, or he can move the code that uses `hdr_addr` into the try block.

msimerson:
Oops, there's two more uses of `hdr_addr` outside the scope of that try block. Eslint is reporting it:

````
/home/travis/build/haraka/Haraka/plugins/data.headers.js
  295:61  warning  'e' is not defined         no-undef
  298:46  warning  'hdr_addr' is not defined  no-undef
  305:52  warning  'hdr_addr' is not defined  no-undef
✖ 3 problems (0 errors, 3 warnings)
````

PR #2330, opened 2018-02-01
Fix address name quotes

KingNoosh:
Fixes #

- 
- 

Checklist:
- [ ] docs updated
- [ ] tests updated


msimerson:
You've got lint failures that need addressing. Also, a test case that exhibits:

> when names are decoded in address values, they must be quoted to ensure
> that extra spaces, commas, etc in the name don't cause address parsing problems.

seems like a prudent addition.

KingNoosh:
I'll see if I can source the example that was causing us issues but that'll be when I get back to the office on Monday

KingNoosh:
If I don't get it done this weekend

KingNoosh:
Going to need to redo this PR, I've somehow broken to many tests

PR #2331, opened 2018-02-01
Warn on max_header_lines

KingNoosh:
Fixes #

- 
- 

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -1584,8 +1584,7 @@ class Connection {
         }
 
         // Warn if we hit the maximum parsed header lines limit
-        const max_lines = trans.MAX_HEADER_LINES;
-        if (this.transaction.header_lines.length >= max_lines) {
+        if (this.transaction.header_lines.length >= trans.MAX_HEADER_LINES) {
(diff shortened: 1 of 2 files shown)


msimerson:
We've already got `const MAX_HEADER_LINES = config.get('max_header_lines') || 1000;` in `./transaction.js.` Does it make sense to export MAX_HEADER_LINES so you can reference it from `connection.js`?

KingNoosh:
I can take a further look.

connection.js
```diff
@@ -1583,6 +1583,12 @@ class Connection {
             return;
         }
 
+        // Warn if we hit the maximum parsed header lines limit
+        const max_lines = trans.MAX_HEADER_LINES;
+        if (this.transaction.header_lines.length >= max_lines) {
```
msimerson: What's the utility of creating max_lines?  You can now:

`if (this.transaction.header_lines.length >= trans.MAX_HEADER_LINES) {`

right?

KingNoosh: Makes sense since we're not using it more than once

PR #2335, opened 2018-02-06
smtp_forward: enable_outbound can be set per domain

gramakri:
Fixes #
enable_outbound can be set per domain

When haraka is managing multiple domains, it is sometimes desirable to use the mail relay only for specific domains. By making the enable_outbound flag per domain, we can disable the relay and use Haraka's internal MX based delivery.

Checklist:
[X] docs updated
[X] tests updated
[X ] Changes updated

### DIFF
--- a/plugins/queue/smtp_forward.js
+++ b/plugins/queue/smtp_forward.js
@@ -26,9 +26,7 @@ exports.register = function () {
 
     plugin.register_hook('queue', 'queue_forward');
 
-    if (plugin.cfg.main.enable_outbound) {
-        plugin.register_hook('queue_outbound', 'queue_forward');
-    }
+    plugin.register_hook('queue_outbound', 'queue_forward');
 };
 
 exports.load_smtp_forward_ini = function () {
@@ -42,6 +40,7 @@ exports.load_smtp_forward_ini = function () {
             '-main.check_sender',
             '-main.check_recipient',
             '*.enable_tls',
+            '*.enable_outbound'
         ],
     },
     function () {
@@ -60,6 +59,14 @@ exports.get_config = function (connection) {
     if (!plugin.cfg[dom]) return plugin.cfg.main;  // no specific route
 
     return plugin.cfg[dom];
+};
+
+exports.is_outbound_enabled = function (cfg) {
+    const plugin = this;
+
+    if ('enable_outbound' in cfg) return cfg.enable_outbound; // pick up per-domain flag if set
+
+    return plugin.cfg.main.enable_outbound; // follow the global configuration
 };
 
 exports.check_sender = function (next, connection, params) {
@@ -208,6 +215,11 @@ exports.queue_forward = function (next, connection) {
     }
 
     const cfg = plugin.get_config(connection);
+
+    if (connection.relaying && !plugin.is_outbound_enabled(cfg)) {
+        connection.logdebug(plugin, 'skipping, outbound disabled');
+        return next();
+    }
 
     smtp_client_mod.get_client_plugin(plugin, connection, cfg, function (err, smtp_client) {
         smtp_client.next = next;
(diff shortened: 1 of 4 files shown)


msimerson:
I'm +1 for making enable_outbound settings on a per-domain basis. This implementation seems excessively complicated and I'm curious why you implemented it this way.

I think the implementation would be simpler, consistent with how the rest of the per-domain options work, easier to reason about the config options (why does setting the main option to false disable the per-domain option?), and be more flexible if instead you used this logic:

• when per-domain enable_outbound is defined, honor it
• else use main.enable_outbound setting.

When implemented this way, you could have `enable_outbound` disabled by default (in main) and enable only for specific domains. This implementation excludes that use case.

gramakri:
@msimerson Indeed, I have over-complicated it. I have updated the PR per your suggestion. Thanks!

plugins/queue/smtp_forward.js
```diff
@@ -62,6 +60,14 @@ exports.get_config = function (connection) {
     return plugin.cfg[dom];
 };
 
+exports.is_outbound_enabled = function (cfg) {
+    const plugin = this;
+
+    if ('enable_outbound' in cfg) return cfg.enable_outbound === 'true'; // pick up per-domain flag if set
```
msimerson: Are you sure this will do what you want? The in operator will fire if enable_outbound is defined in the cfg object. So you're going to return the string "true", even if the value of enable_outbound is boolean false. Shouldn't we be consistently using JS booleans instead of strings?

tests/plugins/queue/smtp_forward.js
```diff
@@ -138,3 +138,37 @@ exports.get_mx = {
...
+        test.expect(1);
+        this.plugin.cfg['test.com'].enable_outbound = 'false';
+        this.connection.transaction.rcpt_to = [ new Address('<postmaster@test.com>') ];
+        const cfg = this.plugin.get_config(this.connection);
+        test.equal(this.plugin.is_outbound_enabled(cfg), false);
+        test.done();
+    },
+    'per-domain enable_outbound is true even if top level is false' : function (test) {
+        test.expect(1);
+        this.plugin.cfg.enable_outbound = false; // this will be ignored
```
msimerson: In the PR I sent, I had corrected this path to: `this.plugin.cfg.main.enable_outbound`

tests/plugins/queue/smtp_forward.js
```diff
@@ -138,3 +138,37 @@ exports.get_mx = {
...
+        this.plugin.cfg['test.com'].enable_outbound = 'true'; // this will be ignored
```
msimerson: Again, you should be using booleans here instead of a string.

tests/plugins/queue/smtp_forward.js
```diff
@@ -138,3 +138,37 @@ exports.get_mx = {
...
+    'per-domain enable_outbound is true by default' : function (test) {
+        test.expect(1);
+        this.connection.transaction.rcpt_to = [ new Address('<postmaster@test.com>') ];
+        const cfg = this.plugin.get_config(this.connection);
+        test.equal(this.plugin.is_outbound_enabled(cfg), true);
+        test.done();
+    },
+    'per-domain enable_outbound can be set to false' : function (test) {
+        test.expect(1);
+        this.plugin.cfg['test.com'].enable_outbound = 'false';
```
msimerson: again, use boolean instead of string.

tests/plugins/queue/smtp_forward.js
gramakri: Ah, I didn't see the PR at all and blindly force pushed. good catch.

plugins/queue/smtp_forward.js
gramakri: @msimerson The config values in subdomains are not parsed out as booleans. Is there a way to specify this in the `booleans` section when reading the config file?

gramakri: Oh, I see `*.enable_tls`. Maybe I can do this for enable_outbound. trying now.

PR #2339, opened 2018-02-07
line_socket: remove superfluous function

msimerson:
should be reviewed by @baudehlo

- Near as I can tell, despite this function being attached to the socket object, it is never called from outside `setup_line_processor()`. If my cursory check is representative, exporting those functions has served only one purpose over the years: unnecessary function calls on the stack.
- Note: there are functions with exactly the same names in connection.js and outbound/index.js. However, those are independently defined and accessed. A secondary use of these intermediate functions has been an additional namespace collision. I have therefore renamed them.

### DIFF
--- a/line_socket.js
+++ b/line_socket.js
@@ -16,7 +16,7 @@ class Socket extends net.Socket {
 function setup_line_processor (socket) {
     let current_data = '';
 
-    socket.on('data', function process_data (data) {
+    socket.on('data', function on_socket_data (data) {
         current_data += data;
         let results;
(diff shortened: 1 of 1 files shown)


msimerson:
Near as I can tell, nothing currently calls those functions outside that file (nor should anything ever have since they are event handlers), but you'd likely be the one to know if something else ever did.

baudehlo:
I already approved it on your repo I think. Anyway, lgtm.

KingNoosh:
@baudehlo should I just approve it on here on your behalf?

PR #2349, opened 2018-02-21
smtp_forward: add domain_selector

### LINKED ISSUE #2346: smtp_forward: support relaying based on MAIL FROM domain
The current plugin supports relaying based on looking the recipient domains. On the Cloudron, we allow setting a relay per domain. Any mail originating from the domain is sent via the configured relay.

gramakri:
this config variable can be set to rcpt_to or mail_from to decide how the domain configuration is chosen

Fixes #2346

- add domain_selector config variable that chooses domain configuration based on mail_from or rcpt_to domain

Checklist:
- [ X ] docs updated
- [ X ] tests updated
- [ X ] [Changes](https://github.com/haraka/Haraka/blob/master/Changes.md) updated

### DIFF
--- a/plugins/queue/smtp_forward.js
+++ b/plugins/queue/smtp_forward.js
@@ -52,8 +52,16 @@ exports.get_config = function (connection) {
     const plugin = this;
 
     if (!connection.transaction) return plugin.cfg.main;
-    if (!connection.transaction.rcpt_to[0]) return plugin.cfg.main;
-    const dom = connection.transaction.rcpt_to[0].host;
+
+    let dom;
+    if (plugin.cfg.main.domain_selector === 'mail_from') {
+        if (!connection.transaction.mail_from) return plugin.cfg.main;
+        dom = connection.transaction.mail_from.host;
+    }
+    else {
+        if (!connection.transaction.rcpt_to[0]) return plugin.cfg.main;
+        dom = connection.transaction.rcpt_to[0].host;
+    }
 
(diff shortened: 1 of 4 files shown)


plugins/queue/smtp_forward.js
```diff
@@ -52,8 +52,15 @@ exports.get_config = function (connection) {
     const plugin = this;
 
     if (!connection.transaction) return plugin.cfg.main;
-    if (!connection.transaction.rcpt_to[0]) return plugin.cfg.main;
-    const dom = connection.transaction.rcpt_to[0].host;
+
+    var dom;
```
KingNoosh: No `var`, `let` or `const` only please.

plugins/queue/smtp_forward.js
```diff
@@ -52,8 +52,15 @@ exports.get_config = function (connection) {
...
+    if (plugin.cfg.main.domain_selector === 'mail_from') {
+        if (!connection.transaction.mail_from) return plugin.cfg.main;
+        dom = connection.transaction.mail_from.host;
+    } else {
```
KingNoosh: No hugging elses, stroustrup brace style please

plugins/queue/smtp_forward.js
gramakri: Fixed

plugins/queue/smtp_forward.js
gramakri: Fixed this as well.

PR #2350, opened 2018-02-22
document force_shutdown_timeout and graceful_shutdown

luto:
This adds documentation for [`force_shutdown_timeout`](https://github.com/haraka/Haraka/blob/master/server.js#L45) as well as [`graceful_shutdown`](https://github.com/haraka/Haraka/blob/master/server.js#L116) found in the code, but not in the config. Please let me know if I got the meaning of those settings wrong.

### DIFF
--- a/docs/CoreConfig.md
+++ b/docs/CoreConfig.md
@@ -40,6 +40,8 @@ The list of plugins to load
   * spool\_dir - (default: none) directory to create temporary spool files in
   * spool\_after - (default: -1) if message exceeds this size in bytes, then spool the message to disk
     specify -1 to disable spooling completely or 0 to force all messages to be spooled to disk.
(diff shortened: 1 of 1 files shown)


smfreegard:
@baudehlo @msimerson 

Didn't we intentionally remove these?  Weren't we going to get rid of this functionality as it was too slow to be of any use?  Especially as if we drop the connection, the remote host will simply retry later anyway by design of the SMTP protocol.

luto:
After a short discussion on IRC I'm now aware that restarts aren't very needed or common in haraka. If those settings are going to be removed, please note that `force_shutdown_timeout` is are also present in the [default `smtp.ini`](https://github.com/haraka/Haraka/blob/master/config/smtp.ini#L41-L45).

Feel free to close this as needed. Thanks!

msimerson:
> Didn't we intentionally remove these?

That was your and my feelings, but @baudehlo wanted to keep them. So we just disabled them by default.

PR #2354, opened 2018-03-02
Ensure unrecognized_command hooks are invoked with params as an array

gramakri:
Plugins expect the last parameter cmd/params to be an array and not a string.
If an empty command is passed, tls plugin fails because it expects the argument
to be an array:

[CRIT] [3DD21EB3-057C-445B-9345-CDA7552D6BE0] [core] Plugin tls failed: TypeError: Cannot read property 'toUpperCase' of undefined
    at Plugin.exports.upgrade_connection (/app/haraka/plugins/tls.js:93:18)
    at Object.plugins.run_next_hook (/app/haraka/plugins.js:506:28)
    at callback (/app/haraka/plugins.js:464:32)
    at Plugin.exports.hook_unrecognized_command (/app/haraka/plugins/max_unrecognized_commands.js:27:12)

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -315,7 +315,7 @@ class Connection {
             const [ , methodMatch, , remainingMatch] = /^([^ ]*)( +(.*))?$/.exec(this.current_line);
             if (!methodMatch) {
                 return plugins.run_hooks('unrecognized_command',
-                    this, this.current_line);
+                    this, [this.current_line]);
             }
             const method = `cmd_${methodMatch.toLowerCase()}`;
             const remaining = remainingMatch || '';
--- a/plugins/max_unrecognized_commands.js
+++ b/plugins/max_unrecognized_commands.js
@@ -10,10 +10,10 @@ exports.hook_connect = function (next, connection) {
     return next();
 }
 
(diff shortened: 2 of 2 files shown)


msimerson:
I've looked and agree with this PR. However, in the `plugins/max_unrecognized_commands.js` the usage of the params arguments is `cmd` singular. It should be updated there too, so that it's plural everywhere. And that fail message as well. Agree?

gramakri:
@msimerson Ah, good catch. Agree, I will update this PR

gramakri:
I have verified that using an array in javascript string literal works :-) It does a join(',') it seems

smfreegard:
Isn't the pattern of the other commands (next, connection, cmd, params) ??

msimerson:
> Isn't the pattern of the other commands (next, connection, cmd, params) ??

If I understand your question correctly, no. I think the `cmd` you're thinking of is the `methodMatch` and when set, is the first argument in the params array. See also [line 342 in connections.js](https://github.com/haraka/Haraka/blob/master/connection.js#L342) as the only other place where the hook is invoked.

PR #2370, opened 2018-03-20
fix link to cluster in coreconfig docs

luto:
The link to "cluster" on http://haraka.github.io/manual/CoreConfig.html currently links to `http://learnboost.github.com/cluster/%20or%20node%20version%20%3E=%200.8`. This makes it link only to `http://learnboost.github.com/cluster/` as originally intended.


msimerson:
I think its safe to drop the "node 0.8" comment now. Agree?

smfreegard:
Definitely +1 for removing all references to 0.8 and the learnboost cluster repo

PR #2376, opened 2018-03-23
Fix handling of LMTP sockets

smfreegard:
As per IRC channel.  This fixes any MX list where a socket path is specified (e.g. LMTP).

### DIFF
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -270,6 +270,11 @@ class HMailItem extends events.EventEmitter {
             // duplicate each MX for each ip address family
             this.mxlist = [];
             for (const mx in mxlist) {
+                // Handle UNIX sockets for LMTP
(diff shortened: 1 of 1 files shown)


outbound/hmail.js
```diff
@@ -270,6 +270,11 @@ class HMailItem extends events.EventEmitter {
             // duplicate each MX for each ip address family
             this.mxlist = [];
             for (const mx in mxlist) {
+                // Handle UNIX sockets for LMTP
+                if (mxlist[mx].path) {
+                    this.mxlist.push(mxlist[mx]);
+                    continue;
```
baudehlo: This breaks with the style of the other part of the for loop. Just use an else if for the next line instead.

smfreegard: Ok - done.

PR #2387, opened 2018-04-13
Handle connection.relaying set on a transaction-level hook to prevent open-relay

smfreegard:
This PR is to address something I've noticed that multiple people on IRC have fallen foul of when writing their own plugins where they are setting `connection.relaying = true` in `hook_mail`, `hook_rcpt` or `hook_data` and do not track and unset it at the end of that transaction, thereby making their systems open-relays after a message has been sent that sets the flag.  It also saves a few lines of code in a plugin if you do need to set `connection.relaying` for a single transaction as you don't have to track and unset it - it's explicitly reset when the transaction is reset.

This adds `transaction.relaying` and includes code to put a warning into the logs if it is found that a plugin sets `connection.relaying` within a transaction-level hook telling them to use `transaction.relaying` instead:

`````
[INFO] [74FC27BB-2D17-42B8-A173-F8074C5C5DB8.1] [core]  hook=mail plugin=relay_warn_test function=hook_mail params=<> retval=OK msg=""
[WARN] [74FC27BB-2D17-42B8-A173-F8074C5C5DB8.1] [core] connection.relaying was set by a plugin during a transaction.  Use of transaction.relaying instead is recommended as it is much safer
`````

It also adds a `relay` flag to the log entry emitted when a message handled so the state of the flag is correctly shown.

It also makes `transaction.relaying` mirror the value of `connection.relaying` if it is set as a convenience to plugin authors, so that they only have to check `transaction.relaying` instead of both variables to check if the permission is allowed.

Checklist:
- [X] docs updated

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -677,8 +677,8 @@ class Connection {
     }
     warn_relay () {
         if (this.relay_warn !== this.relaying) {
-	    this.logwarn("connection.relaying was set by a plugin during a transaction.  Use of transaction.relaying instead is recommended as it is much safer");
-	}
+            this.logwarn("connection.relaying was set by a plugin during a transaction.  Use of transaction.relaying instead is recommended as it is much safer");
+        }
     }
     /////////////////////////////////////////////////////////////////////////////
     // SMTP Responses
@@ -1304,8 +1304,8 @@ class Connection {
         // Track connection.relaying state
         this.relay_warn = this.relaying;
         // Require authentication on connections to port 587 & 465
-        if (!(this.relaying || (this.transaction && this.transaction.relaying)) 
-            && [587,465].indexOf(this.local.port) !== -1) 
+        if (!(this.relaying || (this.transaction && this.transaction.relaying))
+            && [587,465].indexOf(this.local.port) !== -1)
         {
             this.errors++;
             return this.respond(550, 'Authentication required');
--- a/connection.js
+++ b/connection.js
@@ -105,6 +105,7 @@ class Connection {
         this.early_talker = false;
         this.pipelining = false;
         this.relaying = false;
+        this.relay_warn = false;
         this.esmtp = false;
         this.last_response = null;
         this.hooks_to_run = [];
@@ -644,6 +645,9 @@ class Connection {
                 });
             });
             self.transaction.results = new ResultStore(self);
+            // Convenince for plugin authors so they only need to check
+            // c.t.relaying instead of both c.relaying and c.t.relaying.
+            self.transaction.relaying = self.relaying;
             if (cb) cb();
         });
     }
@@ -670,6 +674,11 @@ class Connection {
             self.prev_state = null;
         }
         setImmediate(() => self._process_data());
+    }
+    warn_relay () {
+        if (this.relay_warn !== this.relaying) {
+	    this.logwarn("connection.relaying was set by a plugin during a transaction.  Use of transaction.relaying instead is recommended as it is much safer");
+	}
     }
     /////////////////////////////////////////////////////////////////////////////
     // SMTP Responses
@@ -950,6 +959,7 @@ class Connection {
             this.logerror("mail_respond found no transaction!");
             return;
         }
+        this.warn_relay();
         const sender = this.transaction.mail_from;
         const dmsg   = `sender ${sender.format()}`;
         this.lognotice(
@@ -1076,7 +1086,7 @@ class Connection {
         }
     }
     rcpt_respond (retval, msg) {
-        if (retval === constants.cont && this.relaying) {
+        if (retval === constants.cont && (this.relaying || this.transaction.relaying)) {
             retval = constants.ok;
         }
 
@@ -1085,6 +1095,9 @@ class Connection {
             this.results.add(this, {err: 'rcpt_respond found no transaction'});
             return;
         }
+
+        this.warn_relay();
+
         const rcpt = this.transaction.rcpt_to[this.transaction.rcpt_to.length - 1];
         const dmsg = `recipient ${rcpt.format()}`;
         if (retval !== constants.ok) {
@@ -1288,8 +1301,12 @@ class Connection {
             this.errors++;
             return this.respond(503, 'Use EHLO/HELO before MAIL');
         }
+        // Track connection.relaying state
+        this.relay_warn = this.relaying;
         // Require authentication on connections to port 587 & 465
-        if (!this.relaying && [587,465].indexOf(this.local.port) !== -1) {
+        if (!(this.relaying || (this.transaction && this.transaction.relaying)) 
+            && [587,465].indexOf(this.local.port) !== -1) 
+        {
             this.errors++;
             return this.respond(550, 'Authentication required');
         }
@@ -1297,7 +1314,7 @@ class Connection {
         let from;
         try {
             results = rfc1869.parse("mail", line, config.get('strict_rfc1869') &&
-                      !this.relaying);
+                      !(this.relaying || (this.transaction && this.transaction.relaying)));
             from    = new Address (results.shift());
         }
         catch (err) {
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by smfreegard** [comment]
Ok - I realised that I could make a significant improvement based on the original code added to detect relaying added by transaction-level hooks and actually catch and prevent cases of a plugin making Haraka an open-relay.

`````
[DEBUG] [D83CEE66-315A-4C6A-80EF-D5BC833E1AC8.1] [core]  hook=reset_transaction plugin=relay_warn_test function=hook_reset_transaction params="" retval=CONT msg=""
[CRIT] [D83CEE66-315A-4C6A-80EF-D5BC833E1AC8.1] [core] Relaying explicitly disabled on this connection to prevent allowing open-relay
`````

I also removed the logwarn regarding using `transaction.relaying` instead of `connection.relaying` as it's probably too annoying and this logcrit will instead catch the really bad cases.

**Comment by msimerson** [comment]
This is a good idea. However, there may be some issues with the implementation.

- there's a bunch of plugins that use only `connection.relaying` as their data source, including queue plugins.
- those plugins currently have no means to detect `transaction.relaying`.  Should they?  If so, adding code like `(connection.relaying || (connection.transaction && connection.transaction.relaying)))` to them is rather unattractive.
- this is a good idea: `// Convenince for plugin authors so they only need to check c.t.relaying instead of both c.relaying and c.t.relaying` but not a good implementation as it adds more places in the code where we have to guard against `if (!this.transaction)`, which is a long-standing source of crashing bugs. 

### potential ideas

- connection already has get/set functions, so instead of checking `connection.relaying`, we could deprecate `if (connection.relaying)` in favor of `if (connection.get('relaying'))`. Then we just modify the getter to also check the property on the transaction.
- Add a function like `can_relay()` or `has_relay()` or just `relay()` that checks `connection.relaying` and `connection.transaction.relaying`. This would be a slightly more succinct calling pattern: `if (connection.relay())`
- rather than a boolean, `connection.relaying` and `connection.transaction.relaying` could be the name of the plugin, specific rule, or the hook that granted relaying.
- deprecate `connection.relaying` and force everything through get/set. Then if something calls `set('relaying', someValue)` while a transaction is present, set only `transaction.relaying`. Plugins like `relay` and `auth_*` will still set the value on the connection and everything that sets relaying while in a transaction will see that privilege disappear upon that transaction's conclusion.

The implementation  of get/set would be easy:

```js
index d9bb1ff9..437072c7 100644
--- a/connection.js
+++ b/connection.js
@@ -238,6 +238,15 @@ class Connection {
         }
     }
     set (obj, prop, val) {
+        if (obj === 'relaying') {
+            if (this.transaction) {
+                this.transaction.relaying = prop;
+            }
+            else {
+                this.relaying = prop;
+            }
+        }
+
         if (!this[obj]) this.obj = {};   // initialize
 
         this[obj][prop] = val;  // normalized property location
@@ -263,6 +272,10 @@ class Connection {
         // /sunset
     }
     get (prop_str) {
+        if (prop_str === 'relaying') {
+            if (this.transaction && this.transaction.relaying) return this.transaction.relaying;
+            return this.relaying;
+        }
         return prop_str.split('.').reduce((prev, curr) => {
             return prev ? prev[curr] : undefined
         }, this)
```

smfreegard:
> there's a bunch of plugins that use only connection.relaying as their data source, including queue plugins.

This makes absolutely no difference to existing plugins - as it shouldn't.  Nothing changes except we have a means to set `transaction.relaying=true` if someone finds it necessary to set relay in a transactional command and that guarantees a higher level of safety as the permission is removed when the transaction is reset.

All of the other suggestions are fine - but they would require changes to plugins or require us to deprecate stuff and if you want to do that - go right ahead and do it in the v3 branch, this fixes a clearly identified need and will stop bad plugins from accidentally making Haraka an open-relay, so is an immediate security win.  

I'm personally not inclined to do anything above this.

msimerson:
> Nothing changes except we have a means to set transaction.relaying=true

Yeah, I get that. And you're encouraging folks to use `transaction.relaying` as an alternative to `connection.relaying`. So far, so good. But the majority of the existing code doesn't ever check `transaction.relaying`. Plugin authors who follow your advice and set only `transaction.relaying` while in the transaction hooks are likely to find that they need to set relaying in both places to get other subsequent plugins to behave as they expect. That *seems* like a problem.

smfreegard:
>  And you're encouraging folks to use transaction.relaying as an alternative to connection.relaying

No - I'm absolutely not.  

Allowing relay permissions in a transaction is a very rare case indeed, but if you need to do it then this provides much better safety than doing it using `connection.relaying` for the instance.

Here's a real-world use case in a plugin that I wrote for someone on the IRC channel the other day - it demonstrates the issue here perfectly: https://gist.github.com/smfreegard/f7a11a2245c3c176b20fe07755ffe7ff

My PR today is because someone else that I helped on IRC with a plugin earlier fell into this exactly trap by setting `connection.relaying` in hook_rcpt based on whether the recipient was local or not, yet didn't set `connection.relaying=false` after transaction end, therefore made themselves an open-relay for subsequent transactions on the same connection.   Not ideal at all.

> Plugin authors who follow your advice and set only transaction.relaying while in the transaction hooks are likely to find that they need to set relaying in both places to get other subsequent plugins to behave as they expect. That seems like a problem.

It isn't.  I can't think of a single use-case for setting `connection.relaying` based on hook_mail, hook_rcpt or hook_data attributes where using `transaction.relaying` wouldn't be sufficient and safer - we certainly don't have any plugins in core that do this.

msimerson:
> Allowing relay permissions in a transaction is a very rare case indeed

Completely agree, but since it is a use case, why not provide a clean API that does the right thing?

If I were to ever set transaction.relaying, I *would* expect subsequent plugins (like spamassassin and smtp_forward) to honor it.

>  I can't think of a single use-case for setting connection.relaying based on hook_mail, hook_rcpt or hook_data attributes where using transaction.relaying wouldn't be sufficient and safer

No arguments there at all. What I'm suggesting is that instead of the state tracking, we just modify connection.get/set to automatically set the more specific property. Then update all our code to use get/set, and also encourage everyone to use the `connection.set('relaying', value)` syntax to get the safer behavior. I'm not asking you to do work, just whether that might be a suitable way to solve this problem.

smfreegard:
> If I were to ever set transaction.relaying, I would expect subsequent plugins (like spamassassin and smtp_forward) to honor it.

Ok - I see what you mean now, you're right - those plugins do check for relaying and change their behaviour accordingly and wouldn't see the transaction relaying property without modifications, so yeah - this wasn't the right way to do it.

> No arguments there at all. What I'm suggesting is that instead of the state tracking, we just modify connection.get/set to automatically set the more specific property. Then update all our code to use get/set, and also encourage everyone to use the connection.set('relaying', value) syntax to get the safer behavior. I'm not asking you to do work, just whether that might be a suitable way to solve this problem.

I don't however agree with this.  Any time you catch yourself saying "Then update all our code", then automatically think v3-branch because whilst we can update core and core plugins, that leaves anyone else with plugins out in the cold and makes Haraka v2 a moving target for plugin developers which I'm not cool with.

I've completely changed this PR to remove `transaction.relaying` completely, but left in the code I used to detect when `connection.relaying` was turning Haraka into an open-relay and automatically preventing this, because basically this doesn't break anything, isn't a big change and keeps everything simple and ultimately achieves what I wanted to do in the first place.

connection.js
```diff
@@ -1351,12 +1373,12 @@ class Connection {
             this.errors++;
             return this.respond(503, "Use MAIL before RCPT");
         }
-
+        // Track relay state
+        this.relay_warn = this.relaying;
```
baudehlo: Don't you have to do this check in the _respond functions?

smfreegard: No - I'm making sure this.relay_warn is set to whatever this.relaying is set to *before* the plugins are run as I'm checking to see if they changed the value of this.relaying during the hooks.

smfreegard: I guess I should have called it this.relay_saved ...

PR #2390, opened 2018-04-13
connection.set() for any depth

msimerson:
Currently, `connection.set()` has a three argument call signature: [first, second, value] and does this `connection[first][second] = value` with a little bit of backwards compatibility. Example:
- `connection.set('local', 'ip', '1.2.3.4');

This change modifies that to a two argument signature with the first value being a dot delimited string of the target connection path. Example:
- `connection.set('local.ip', '1.2.3.4');

This change permits using set() for settings of any depth including one:
- `connection.set('relaying', true);

Even if approved, this PR should not be merged until after haraka-test-fixtures is updated with the updated set function.

Checklist:
- [x] docs updated
- [x] tests updated
- [ ] haraka-test-fixtures updated


smfreegard:
I know we have get() and set() in connection already, but what's the ultimate point of this other than adding bloat e.g. accessing a variable .vs. running a boadload of functions to ultimately return the value of a variable that was already in scope given that Javascript has no concept of protected/private class variables.

msimerson:
> given that Javascript has no concept of protected/private class variables.

Using get/set is the standard method by which we can have protected/private class variables in JS. Granted, this PR alone doesn't provide them, but it is a step in that direction.

smfreegard:
You mean a kludgy version of protected/private class variables because it isn't actually a standard JS feature right?

Sorry - but this absolutely screams v3-branch to me.

e.g. if you're grabbed Haraka and a bunch of plugins and you're reading the code of said plugins, then the concepts in the plugin are not going to match the docs, so they're going to pop-up on IRC etc. with questions like "What does connection.relaying = true; do?", "What isn't X documented", "What's the difference between connection.set("relaying", "true"); and connection.relaying = true; and why does one plugin do it this way .vs. ...".

smfreegard:
> It does the same thing as get/set under the hood but with cleaner syntax and then we could make connection.relaying a private property.

Like I said before - we can't do that - it breaks all the plugins unless you're going to allow both ways to work for backwards compatibility, at which point - is it really worth it in the first place?  Do it in v3 and be done with it as we can break whatever we want there.

smfreegard:
Sorry - ignore me.  I just read that link to the ES2015 syntax.  It will maintain backwards compatibility, so should work just fine.

However being as I've just removed transaction.relaying as on reflection, there's not that much point in adding it when I had other code to detect and handle misuse of connection.relaying, this probably still isn't needed.

msimerson:
> it breaks all the plugins

No Steve. It doesn't break any plugins. The syntax for calling would be identical to what exists now, so plugins would get the new smarter behavior (setting txn.relaying when in a transaction, setting conn.relaying otherwise) and checking `if (connection.relaying)` would be smart enough to check the txn.relaying and conn.relaying.

msimerson:
I have removed all the code that touched relaying in this PR and also removed the discussion that was focused upon it. This PR is about providing a nicer syntax for setting connection properties, as well as supporting connection properties at depths != 2. This set syntax now matches the get syntax and it also matches the get/set syntax used in [haraka-notes](haraka/haraka-notes).

PR #2393, opened 2018-04-14
connection: make setting relaying in a transaction safe

msimerson:
This is an alternative implementation of Steve's excellent idea in #2387, which is making sure that when relaying is set in a transaction, that it's automatically unset when the transaction ends.

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -104,7 +104,7 @@ class Connection {
         this.deny_includes_uuid = config.get('deny_includes_uuid') || null;
         this.early_talker = false;
         this.pipelining = false;
-        this.relaying = false;
+        this._relaying = false;
         this.esmtp = false;
         this.last_response = null;
         this.hooks_to_run = [];
@@ -266,6 +266,18 @@ class Connection {
         return prop_str.split('.').reduce((prev, curr) => {
             return prev ? prev[curr] : undefined
         }, this)
+    }
+    set relaying (val) {
+        if (this.transaction) {
+            this.transaction._relaying = val;
+        }
+        else {
+            this._relaying = val;
+        }
+    }
+    get relaying () {
+        if (this.transaction && this.transaction._relaying) return this.transaction._relaying;
+        return this._relaying;
     }
     process_line (line) {
         const self = this;
--- a/connection.js
+++ b/connection.js
@@ -276,7 +276,7 @@ class Connection {
         }
     }
     get relaying () {
-        if (this.transaction && this.transaction._relaying !== undefined) return this.transaction._relaying;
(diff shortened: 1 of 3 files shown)


connection.js
```diff
@@ -267,6 +267,18 @@ class Connection {
...
+    set relaying (val) {
+        if (this.transaction) {
+            this.transaction._relaying = val;
+        }
+        else {
+            this._relaying = val;
+        }
+    }
+    get relaying () {
+        if (this.transaction && this.transaction._relaying) return this.transaction._relaying;
```
baudehlo: What if you explicitly set relaying to false in a transaction hook, when it was true in the connection?

Maybe just do `if ('_relaying' in this.transaction)` instead?

msimerson: Then you've just set relaying=false in **only** the transaction. Which is basically what you want, isn't it?

If the connection was granted relaying permissions **outside** of a transaction, it should remain set to that for the duration of the connection. If it's modified for a particular transaction (a rare occurrence), shouldn't that change end with the transaction?

msimerson: Perhaps the conditional there should be !== undefined.

baudehlo: Yeah that's exactly the same as using `in` fwiw.

msimerson: Oooooohhh, right. That's nicer syntax, I'll update.

baudehlo:
Not sure why Steve closed #2393 - doesn't this solve the issue with that ticket?

msimerson:
Actually, this is #2393. I think you're thinking about his PR #2387. Both PRs scratched the same itch.

baudehlo:
Right but I’m suggesting that both have value because Steve’s change warns plugin developers that they created an open relay.

msimerson:
> Right but I’m suggesting that both have value

I initially expected we'd end up with a little bit of both. However, once the conversation led us to the ES2015 get/set that automatically does exactly The Right Thing (preventing open relays when a transaction sets `relaying=true`) without modifying any plugins, without any need for state tracking, without any need for documentation, and without any plugin author effort, it was obvious that this implementation is better than either of our first ideas. By automatically preventing the open relay, there's no longer a need for the warning.

baudehlo:
Oh I now understand (I have baby brain now - takes me a while to comprehend PRs these days). Thanks. Nice feature. I wonder if there's other things we should do something like this for?

msimerson:
> I wonder if there's other things we should do something like this for?

It would be a more complex, but making results.[add|get|etc..] and/or notes.* automatically Do The Right Thing would make them considerably more pleasant to work with.

PR #2395, opened 2018-04-15
Fix dotfile cleanup to consider platform-based prefix.

soncodi:
On non-win platforms, dotfiles begin with `__tmp__` (per qfile.js). This was not matched by the regexp.


baudehlo:
Merged.

PR #2403, opened 2018-04-30
Add denysoft/denysoftdisconnect actions to unrecognized command hook

smfreegard:
Currently if you send a DENYSOFT or DENYSOFTDISCONNECT, then you'll just get the regular 500 response (and no disconnect in the case of DENYSOFTDISCONNECT), so this just treats DENYSOFT as the equivalent DENY and also improves the logging by recording what command was unrecognized (e.g. it will be stored in the lr="" value in hook_disconnect).


smfreegard:
Only if you're doing something with the input.  We're just echoing back what was sent.  I'm unaware of any issues with that.

msimerson:
Maybe I'm reading this wrong but in this case, you're both returning the user supplied input via the network and logging it?  I haven't thought of any other but this at least seems like a vector for log injection attacks. This could affect using Haraka with a log-processor (like fail2ban, etc.). 

OTOH, that risk already exists if folks enable LOGPROTOCOL, but at least there we can legitimately say, "don't run it like that in production."

msimerson:
I'm ready to be convinced that echoing back and logging the unrecognized command is safe. If @smfreegard wants, this PR could be reopened. I can readily see the value of logging the unrec. command for troubleshooting. I'm reluctant to approve this PR until I get the time to review it in detail (ENOTIME error), or someone else concludes, "in this case, it's safe because ... and it doesn't open the door to code injection vulnerabilities or log injection exploits because..."

PR #2407, opened 2018-05-09
CritFix: ensure ClamAV recognises input as a MIME message

smfreegard:
I've just discovered this morning that if the input message does not start with 'Received: ', then ClamAV will not correctly handle the input message as a MIME message and will therefore fail to detect attachments accordingly.

This is particularly bad if you have plugins that add Received-SPF, Auth-Results headers or any plugin that runs c.t.add_leading_header().

I considered a workaround by using `send_pristine: true` in the MessageStream options, but again, this would fail if the message was sent direct-to-MX by an MUA, so I decided to always add a fake Received header for maximum safety.

IMO - this is probably serious enough to warrant an immediate release.

### DIFF
--- a/plugins/clamd.js
+++ b/plugins/clamd.js
@@ -232,7 +232,15 @@ exports.hook_data_post = function (next, connection) {
             const addressInfo = hp === null ? '' : ` ${hp.address}:${hp.port}`;
             connection.logdebug(plugin, `connected to host${addressInfo}`);
             socket.write("zINSTREAM\0", function () {
-                txn.message_stream.pipe(socket, { clamd_style: true });
+                // We *MUST* ensure that the first line is a Received: header
+                // otherwise ClamAV will not interpret the input as MIME...
+                var fake_rcvd = "Received: from fake\r\n";
+                var buf = new Buffer(fake_rcvd.length+4);
+                buf.writeUInt32BE(fake_rcvd.length, 0);
+                buf.write(fake_rcvd, 4);
+                socket.write(buf, function () {
+                    txn.message_stream.pipe(socket, { clamd_style: true });
+                });
             });
         });
 


superman20:
Is this technically a bug in ClamAV that should be fixed upstream?...or do they require that by design? Also, a long time ago, I had a plugin in ClamAV that looked at the received headers.  I know that's an absurd place to look at received headers (and have since moved that logic into SpamAssassin), but I wonder if it would be better to conditionally create the fake received header only if it is necessary so as not to change normal functionality?

smfreegard:
@superman20 - it's not really a bug in ClamAV, there's no flag that we can sent to ClamAV to tell it what we're asking it to scan, so it has to work out what the input is for itself.   I haven't read the code, but I suspect that it's looking for a Received: header within N bytes from the start and that is what causes it to determine that it's looking at message/rfc822 or maildir file.

I chose to always add the fake Received header for safety and for speed as doing it conditionally would require us to read and parse the first few chunks of the MessageStream which would make it more complicated and slower.

I looked at doing it in the MessageStream itself, but I chose the simplest approach which was what I submitted here.

msimerson:
First off: great catch!  And yes, issuing a release soon is prudent.

The idea of pre-pending the header is good, but...

1. you're using [deprecated Buffer syntax](https://nodejs.org/dist/latest-v8.x/docs/api/buffer.html). Sometime soon(ish) that code will need updating.
2. There's no tests, to make sure that after updating the Buffer syntax, it continues to work properly.
3. Adding a test that covers this code in the clamd plugin is non-trivial.

There are two easy ways to make the header prepending code more test-friendly:

- encapsulate it in an exported function
- prepend the header in `./messagestream.js` as you mentioned.

smfreegard:
> you're using deprecated Buffer syntax. Sometime soon(ish) that code will need updating.

Which is used all over MessageStream already.  This can get fixed once all the rest of the places are updated.   I do not have the time to do this now.

> There's no tests, to make sure that after updating the Buffer syntax, it continues to work properly.
> Adding a test that covers this code in the clamd plugin is non-trivial.

My time to do this currently is 0.  I've already fixed my version of Haraka  - I'm providing this here merely as a courtesy for other Haraka users as IMO this is an important fix for those people that rely on Haraka for AV scanning at their gateway.

msimerson:
Thanks. I have a lot less "computer time" than normal right now too but within a few days I'll poke around at it and see about adding some test coverage for this.

smfreegard:
> I have a lot less "computer time" than normal right now too but within a few days I'll poke around at it and see about adding some test coverage for this.

Thanks - I'm in DC from tomorrow and then Montreal over the next 10 days.   If I get time to look then I will.

PR #2380, opened 2018-04-06
Add fallback to config.get('me')

olsonpm:
* Replace all instances of `config.get('me')` with a wrapper which adds the
  expected fallback to `os.hostname()`.

Fixes #2344

Specifically [implements this comment](https://github.com/haraka/Haraka/issues/2344#issuecomment-379316126)

---

**Feedback wanted**
- I didn't know what to do about testing because the code affected isn't currently tested.  If I were to properly implement tests then I'd be spending the next couple days learning the code base and adding the currently missing tests.  I'm willing to add tests but would like some direction so I can narrow my focus.

- I didn't know a good spot for the wrapper function so I placed it [in its own file](https://github.com/haraka/Haraka/compare/master...olsonpm:fix_config-get-me#diff-3a13bfb0cf6129de5b217e13d82cf635).  If there is a more preferable spot then let me know

---

Checklist:
- [] docs updated
    *Instead this updates the code to match the docs*

- [] tests updated
    *Addressed above*

- [✔] [Changes](https://github.com/haraka/Haraka/blob/master/Changes.md) updated

PR #2413, opened 2018-05-15
create config/me if not present

msimerson:
closes #2380 

- create config/me if not present
- use platform dependent os.EOL instead of \n (in config/me and config/host_list)
- grammar fixes in docs/CoreConfig.md
- use config.local.host when connection in scope

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/bin/haraka
+++ b/bin/haraka
@@ -157,12 +157,10 @@ function copyDir (srcPath, dstPath) {
     mkDir(dstPath);
     const files = fs.readdirSync(srcPath);
 
-    for (let i = 0; i < files.length; i++) {
+    for (let i=0; i < files.length; i++) {
 
         // Ignore ".*"
-        if (/^\./.test(files[i])) {
-            continue;
-        }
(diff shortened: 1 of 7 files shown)


bin/haraka
```diff
@@ -208,20 +208,19 @@ function copyFile (srcFile, dstFile) {
...
-    }
-    catch (e) {
-        warning("Unable to write config/host_list file: " + e);
-    }
+    const hostname = `${os.hostname()}${os.EOL}`;
+
+    ['me','host_list'].forEach(f => {
+        const cfPath = path.join(confPath, f);
+        try {
+            if (fs.statSync(cfPath).isFile()) return;
```
baudehlo: Won't this fail if cfPath doesn't exist (because fs.statSync returns null, and then you try and do .isFile on it?) I think that was the point of the original version.

msimerson: Ahh, right you are. Fixed.

PR #2417, opened 2018-05-16
mods to SPF plugin

celesteking:
Part of yuri's patch series.

3780ab47 Allow bypassing SPF check for relaying or auth'ed clients.
5c5dbbab SPF plugin: don't bypass in helo as no info is available.
7982db3e Log excemption reason to Results in SPF plugin
01365392 Add to results if skipping private IP.
0bcba6b7 Introduce configurable spf plugin lookup timeout.
5cd7f445 enable openspf WHY page text support in DENY
968f19bf spf: don't wrap in <> the sender in CB so that we get nice spf text

Essentially, it introduces timeout, openspf explanation DENY text `http://www.openspf.org/Why?s=${scope}&id=${sender_id}&ip=${connection.remote_ip}`, populates results with more info.

Following settings added to `spf.ini`:
```ini
; in secs
lookup_timeout = 15

; bypass hosts that match these conditions
[bypass]
; hosts that relay through us
relaying = false
; hosts that are SMTP AUTH'ed
auth = true

[deny]
openspf_text = true
```

Checklist:
- [x] docs updated
- [ ] tests updated

### DIFF
--- a/plugins/spf.js
+++ b/plugins/spf.js
@@ -73,7 +73,7 @@ exports.load_config = function () {
         plugin.cfg.relay = { context: 'sender' };  // default/legacy
     }
 
-    plugin.cfg.lookup_timeout = plugin.cfg.main.lookup_timeout || 30;
+    plugin.cfg.lookup_timeout = plugin.cfg.main.lookup_timeout || plugin.timeout - 1;
 };
 
 exports.hook_helo = exports.hook_ehlo = function (next, connection, helo) {
--- a/plugins/spf.js
+++ b/plugins/spf.js
@@ -292,7 +292,7 @@ exports.save_to_header = function (connection, spf, result, mfrom, host, id, ip)
     if (!connection) return;
     if (!connection.transaction) return;
     connection.transaction.add_leading_header('Received-SPF',
-        `${spf.result(result)} (${plugin.config.get('me')}: domain of ${host}${result === spf.SPF_PASS ? ' designates ' : ' does not designate '}${connection.remote.ip} as permitted sender) receiver=${plugin.config.get('me')}; identity=${id}; client-ip=${ip ? ip : connection.remote.ip}; helo=${connection.hello.host}; envelope-from=<${mfrom}>`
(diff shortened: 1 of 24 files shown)


plugins/spf.js
```diff
@@ -42,6 +44,9 @@ exports.load_config = function () {
             '-deny_relay.mfrom_softfail',
             '-deny_relay.mfrom_fail',
             '-deny_relay.mfrom_permerror',
+
+            '-bypass.relaying',
+            '-bypass.auth',
```
msimerson: Instead of 'bypass', we have an established pattern of using the term 'skip' instead. Consistency is preferable.

plugins/spf.js
```diff
@@ -66,13 +71,18 @@ exports.load_config = function () {
...
+
+    plugin.cfg.lookup_timeout = plugin.cfg.main.lookup_timeout || 30;
+};
 
 exports.hook_helo = exports.hook_ehlo = function (next, connection, helo) {
     const plugin = this;
 
     // Bypass private IPs
-    if (connection.remote.is_private) { return next(); }
+    if (net_utils.is_private_ip(connection.remote_ip)) {
```
msimerson: Why switch this to using a more expensive function call?  The precomputed property is always set and accessing it will always be faster than a function call.

celesteking: Take it easy, the patch is from 2016. I'll adjust the stuff, don't worry.

plugins/spf.js
```diff
@@ -270,4 +293,16 @@ exports.save_to_header = function (connection, spf, result, mfrom, host, id, ip)
     connection.transaction.add_leading_header('Received-SPF',
         `${spf.result(result)} (${plugin.config.get('me')}: domain of ${host}${result === spf.SPF_PASS ? ' designates ' : ' does not designate '}${connection.remote.ip} as permitted sender) receiver=${plugin.config.get('me')}; identity=${id}; client-ip=${ip ? ip : connection.remote.ip}; helo=${connection.hello.host}; envelope-from=<${mfrom}>`
     );
-}
+};
+
+exports.bypass_hosts = function(connection) {
+    var plugin = this;
```
msimerson: use of `var` is deprecated.

celesteking:
Sorry -- somehow, a couple of upstream commits got lost, that's why you kept seeing strange stuff. I have now adjusted accordingly and it should be ready to roll. Maybe I shoudl squash those 11 commits?

As re: openspf_text, I'd give user more knobs to tune.

PR #2418, opened 2018-05-16
mods to SA plugin

celesteking:
Part of yuri's patch series.

04edc47 Put SA plugin results into Transaction Results object. Use Results logging functionality to replace regular log text.
d500901 Make SA X-Haraka-Relay header name configurable.

spamassassin.ini
```ini
; the header that is sent to spamc
spamc_auth_header = X-Haraka-Auth
```
Checklist:
- [ ] docs updated
- [x] tests updated

### DIFF
--- a/plugins/spamassassin.js
+++ b/plugins/spamassassin.js
@@ -258,9 +258,9 @@ exports.get_spamd_headers = function (connection, username) {
         `X-Envelope-From: ${connection.transaction.mail_from.address()}`,
         `X-Haraka-UUID: ${connection.transaction.uuid}`,
     ];
-    if (connection.notes.auth_user) {
+    if (connection.relaying) {
         const header_name = plugin.cfg.main['spamc_auth_header'] || 'X-Haraka-Relay';
-        headers.push(header_name + ': true');
+        headers.push(`${header_name}: true`);
     }
 
     return headers;
--- a/plugins/spamassassin.js
+++ b/plugins/spamassassin.js
@@ -44,6 +44,9 @@ exports.hook_data_post = function (next, connection) {
     const plugin = this;
     if (plugin.msg_too_big(connection)) return next();
 
+    const header_name = plugin.cfg.main['spamc_auth_header'] || 'X-Haraka-Relay';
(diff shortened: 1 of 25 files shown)


plugins/spamassassin.js
```diff
@@ -269,10 +275,13 @@ exports.get_spamd_socket = function (next, connection, headers) {
     const results_timeout = parseInt(plugin.cfg.main.results_timeout) || 300;
 
     socket.on('connect', function () {
+        // Abort if the transaction is gone
         if (!connection.transaction) {
+            plugin.logwarn(connection, 'Transaction gone, cancelling SPAMD connection');
             socket.end();
-            return;
+            return false;  // next() is called in socket.on('end')
```
msimerson: Why is this not a code smell?  Values should not be returned unless something is checking them.

celesteking:
btw, proto upped to 1.4 because 
> HEADERS       --  Same as PROCESS, but return only modified headers, not body (new in protocol 1.4)

plugins/spamassassin.js
```diff
@@ -44,6 +44,9 @@ exports.hook_data_post = function (next, connection) {
     const plugin = this;
     if (plugin.msg_too_big(connection)) return next();
 
+    const header_name = plugin.cfg.main.spamc_auth_header || 'X-Haraka-Relay';
+    connection.transaction.remove_header(header_name); // just to be safe
+
```
msimerson: Instead of having duplicate code for the auth_header default name, include it instead in the config loader so the value is always set. There's already a block in `load_spamassassin_ini()` called `defaults` to insert it into:

```js
$ git diff plugins/spamassassin.js 
diff --git a/plugins/spamassassin.js b/plugins/spamassassin.js
index 87d2704a..abcde956 100644
--- a/plugins/spamassassin.js
+++ b/plugins/spamassassin.js
@@ -24,6 +24,7 @@ exports.load_spamassassin_ini = function () {
         max_size:     500000,
         old_headers_action: "rename",
         subject_prefix: "*** SPAM ***",
+        spamc_auth_header: 'X-Haraka-Relay',
     };
 
     for (const key in defaults) {
```js

PR #2421, opened 2018-05-16
mods to auth_base plugin

celesteking:
Allow supplying custom message and status codes to passcheck callback routine. We "hijack" `message` param for that, maintaining backwards compatibility in case it's a plain string message.

Checklist:
- [ ] docs updated
- [x] tests updated

### DIFF
--- a/plugins/auth/auth_base.js
+++ b/plugins/auth/auth_base.js
@@ -101,20 +101,28 @@ exports.check_user = function (next, connection, credentials, method) {
         return;
     }
 
-    const passwd_ok = function (valid, message) {
+    // valid: (true|false)
+    // opts: ({ message, code }|String)
+    var passwd_ok = function (valid, opts) {
+        var status_code = (typeof(opts) == 'object' && opts['code']) || (valid ? 235 : 535);
+        var status_message = (typeof(opts) == 'object' ? opts['message'] : opts) ||
+                (valid  ? '2.7.0 Authentication successful' : '5.7.8 Authentication failed');
+
         if (valid) {
             connection.relaying = true;
             connection.results.add({name:'relay'}, {pass: plugin.name});
+            connection.results.add(plugin, {pass: method});
+
             connection.results.add({name:'auth'}, {
                 pass: plugin.name,
                 method: method,
-                user: credentials[0],
+                user: credentials[0]
             });
-            connection.respond(235, ((message) ? message : 'Authentication successful'), function () {
+
+            connection.respond(status_code, status_message, function () {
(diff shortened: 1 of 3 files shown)


plugins/auth/auth_base.js
```diff
@@ -101,20 +101,28 @@ exports.check_user = function (next, connection, credentials, method) {
         return;
     }
 
-    const passwd_ok = function (valid, message) {
+    // valid: (true|false)
+    // opts: ({ message, code }|String)
+    var passwd_ok = function (valid, opts) {
```
msimerson: Switching from `const` to `var`  is part of why your lint tests are failing. Better still is shedding the function assignment and creating a function declaration: `function passwd_ok (valid, opts) {`.

plugins/auth/auth_base.js
```diff
@@ -101,20 +101,28 @@ exports.check_user = function (next, connection, credentials, method) {
...
         if (valid) {
             connection.relaying = true;
             connection.results.add({name:'relay'}, {pass: plugin.name});
+            connection.results.add(plugin, {pass: method});
+
             connection.results.add({name:'auth'}, {
                 pass: plugin.name,
                 method: method,
-                user: credentials[0],
+                user: credentials[0]
```
msimerson: You don't need to agree that trailing commas in lists are preferable, and contributions aren't required to have them, but also please don't remove them.

plugins/auth/auth_base.js
```diff
@@ -101,20 +101,28 @@ exports.check_user = function (next, connection, credentials, method) {
...
                 method: method,
                 user: credentials[0],
             });
-            connection.respond(235, ((message) ? message : 'Authentication successful'), function () {
+
+            connection.respond(status_code, status_message, function () {
                 connection.authheader = "(authenticated bits=0)\n";
                 connection.auth_results(`auth=pass (${method.toLowerCase()})`);
                 connection.notes.auth_user = credentials[0];
-                connection.notes.auth_passwd = credentials[1];
```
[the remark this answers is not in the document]
celesteking: Yeah, that's why it adds in Changes.md:
```diff
+    * Breaking: auth_base.js: don't set `connection.notes.auth_passwd`
```
Anyway, I'll make it configurable, but I think we already talked about this -- that this is a bad practice. Main danger lurks in further plugins, possibly user-written, that would somehow allow attacker to execute malicious code. Once he's able to get access to notes object, user's info is compromised.

I'd recommend switching the default to REMOVE rather than LEAVE.

PR #2423, opened 2018-05-16
lower log level for inbound TLS timeout: way too common occurrance

celesteking:
What do you think?


msimerson:
I just checked and I see about 1-3 of these per day, primarily from iPhones on AT&T's network. Whether it's logged as error or notice makes no difference to me.

PR #2427, opened 2018-05-17
mods to transaction.js & connection.js

celesteking:
* Basically, this patch adds more info to notes object, like:
  * `connection.notes.proxy` , for remotes that use PROXY proto
  * `transaction.msg_status.(rejected|tempfailed|accepted) = true`, for ex., for consequent logging into ES 
* Don't add anything to queue results if queue hook returned CONT (that probably means we want next queue hook to fire, which will populate the results with delivery info)
* Fix "queued" messages, so that we don't get ($uuid)($uuid) duplicates and such
* fixes #2403 , where we don't just ignore denysoftdisconnect on unrecognized_command

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -1665,28 +1665,28 @@ class Connection {
             case constants.deny:
                 this.respond(550, msg || "Message denied", () => {
                     self.msg_count.reject++;
-                    self.transaction.msg_status.rejected = true;
+                    self.transaction.msg_status = 'rejected';
                     self.reset_transaction(() => self.resume());
                 });
                 break;
             case constants.denydisconnect:
                 this.respond(550, msg || "Message denied",() => {
                     self.msg_count.reject++;
-                    self.transaction.msg_status.rejected = true;
+                    self.transaction.msg_status = 'rejected';
                     self.disconnect();
                 });
                 break;
             case constants.denysoft:
                 this.respond(450, msg || "Message denied temporarily", () =>  {
                     self.msg_count.tempfail++;
-                    self.transaction.msg_status.tempfailed = true;
+                    self.transaction.msg_status = 'deferred';
                     self.reset_transaction(() => self.resume());
                 });
                 break;
             case constants.denysoftdisconnect:
                 this.respond(450, msg || "Message denied temporarily",() => {
                     self.msg_count.tempfail++;
-                    self.transaction.msg_status.tempfailed = true;
+                    self.transaction.msg_status = 'deferred';
                     self.disconnect();
                 });
                 break;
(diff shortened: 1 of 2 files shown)


msimerson:
This `msg_status` part of this PR strikes me as odd. For starters, you've already got the queue status which tells what happened to the message. If that's insufficient, why 3 booleans on an object instead of a single property named `msg_status` with three values: [accepted, deferred, rejected]?

msimerson:
>  I can tell you it's only used in my elasticsearch logging plugin

That part makes sense to me.

> Maybe it was related to the way ES query were formed, where just checking on Boolean was easier than checking/comparing to a string.

That's the part that doesn't make a lot of sense to me. For an ES query is no easier to check `transaction.msg_status_accepted: true` than `transaction.msg_status: accepted`. The latter is shorter and it makes it much easier and more fun to browse and graph (in Kibana) aggregated versions of the data using the latter syntax.

> Elaborate, please, where's that status stored?

If a message is "accepted" then the transaction will have a msg.queue.pass property (`_exists_: msg.queue.pass`) in ES. If that property doesn't exist, it was: deferred, rejected, or never tried (because sender or recipient attempts failed). And now I see the issue, you can't tell why without inspecting other plugin results.

So really, your `transaction.msg_status` would have 4 indicative values:
- undefined: when sender or recipient validation failed (never made it to DATA)
- accepted: (2XX value)
- rejected: (5XX value)
- deferred: (4XX value)

celesteking:
Alright.

transaction.js
```diff
@@ -38,6 +38,7 @@ class Transaction {
             tempfail: 0,
             reject: 0,
         }
+        this.msg_status = 'undefined';
```
msimerson: 'undefined' shouldn't be quoted. It should be one of: `undefined`, `null`, or an empty string `''`.

celesteking: The reason for a string is because faulty Kibana/ES didn't know how to output "missing" count aggregate. They've added that in https://www.elastic.co/blog/kibana-6-2-0-released , so we should be golden. We could also pre-process that in ES plugin, substituting missing/undefined value with a string.

connection.js
```diff
@@ -772,6 +772,7 @@ class Connection {
                 this.respond(500, msg || "Unrecognized command");
                 break;
             case constants.denydisconnect:
+            case constants.denysoftdisconnect:
```
baudehlo: Shouldn't `denysoftdisconnect` disconnect after 421 not 521?

baudehlo: Also no handler for plain `denysoft`.

connection.js
```diff
@@ -1750,30 +1762,34 @@ class Connection {
...
                 this.respond(450, msg, () => {
                     self.msg_count.tempfail++;
+                    self.transaction.msg_status = 'deferred';
                     self.disconnect();
                 });
                 break;
             default:
                 outbound.send_email(this.transaction, (retval2, msg2) => {
-                    if (!msg2) msg2 = self.queue_msg(retval2, msg2);
+                    if (!msg2) msg2 = self.queue_msg(retval2, msg);
```
baudehlo: Can you explain the purpose of this change?

baudehlo: And shouldn't it also do the `|| 'Message Queued'` from above?

transaction.js
```diff
@@ -38,6 +38,7 @@ class Transaction {
             tempfail: 0,
             reject: 0,
         }
+        this.msg_status = undefined;
```
baudehlo: Why not `null` like other fields?

connection.js
celesteking: @smfreegard  said "The RFC isn't really very clear on unrecognized commands resulting in anything other than a 5xx". Thus, 5xx and not 4xx. Also, sending unrec commands is bad, especially when a plugin knows about it and explicitly requests client to disconnect.
Plain denysoft handled by `default` case.

connection.js
celesteking: I'm a bit perplexed. 1. Because `msg2` is `null` in that case and we'd like to use the message that queue plugin returned. 2. It shouldn't do because that's already done by `msg`.

transaction.js
celesteking: Ask the other person of this discussion. But basically, because it's a good English word, it completely reflects the status.

msimerson: FTR, I didn't choose `undefined` over `null`, I suggested that `undefined` the special JS value was better than `'undefined'` as a string. If `null` is more consistent with the surrounding code, and there's no compelling reason to prefer `undefined`, then null is preferable. In this case, I think it doesn't matter to Kibana/ES because the ES plugin prunes both `null` and `undefined`.

baudehlo:
The issue with undefined is the slightly different behaviour in JS. To check if a key in an object has never been assigned to you’ll often see `if (obj.key === undefined)` (I think it’s clearer to use `’key’ in obj` but whatever) hence the use of null instead.

celesteking:
yeah, it's of very much irrelevant nature, let's concentrate on something more important guys. ;)

PR #2428, opened 2018-05-17
list pools feature

celesteking:
You have introduced pools, but you haven't supplied a way to look into them.
This patch adds `--oplist` feature that lists currently registered pools. Initially, I thought that would list only outbound pool, but I was mistaken as all pools (smtp_client) seem to end up in `server.notes.pool`

The output format is:
> `[${instance.inUseObjectsCount()}/${instance.getPoolSize()}] ${name}`;

Checklist:
- [ ] docs updated
- [ ] tests updated


connection.js
```diff
@@ -1255,8 +1255,13 @@ class Connection {
             return this.respond(501, "No command given");
         }
 
-        require('./server').sendToMaster(command, results);
-        return this.respond(250, "Command sent for execution. Check Haraka logs for results.");
+        const obj = require('./server').sendToMaster(command, results);
+        if (obj && obj.error)
```
msimerson: Sorry to nitpick, but a long time agree we settled on a single line or braces. Your choice.

celesteking:
No problem. Note, that feature doesn't work in cluster mode (not using it here). 

We should also think about introducing a framework that would allow exchanging internal commands. So that you don't employ `--qlist`, `--qstat` hacks. 
Just send an internal command to server and output a reply. The exchange could be in JSON and then processed by frontend bin/haraka into human text.

baudehlo:
That exists. There's an INTERNALCMD verb used for that.

celesteking:
It is used right by this patch, but there's no defined framework for it. There's no defined API and no format conventions. Initially, you could just submit a command to the server and that's all (this patch allows returning text data back). We need something more advanced.

msimerson:
> There's no defined API ... We need something more advanced.

Why not just choose JSON? That's our default.

baudehlo:
The problem with INTERNALCMD is it was designed to send the same command to all the children, and because we didn't know if they could respond (e.g. the graceful restart) no expected response.

What we should probably do though is still support that, but change your patch a bit to return a list of objects up the chain. Ordering hopefully doesn't matter.

analogic:
+1 I've started coding same cmd until I found this PR

msimerson:
the only reason I haven't merged this is because of @baudehlo 's suggestions.

baudehlo:
Yes you can communicate between master/children in a plugin. See how the
process title plugin does it.

msimerson:
Closing as abandoned.

PR #2431, opened 2018-05-18
mods to outbound

celesteking:
This is a collection of small "improvements" over `outbound` framework.
* rework send_email_respond switch so that it would accept `DENY` and send bounce with passed `params`
* add hack to allow passing params into bounce creation routine
* add trans uuid to Received header
* include mx.bind IP address in log message
* note outbound client UUID in `hmail`.todo
* pass `extra` param to `deferred` hook

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -295,7 +295,7 @@ class HMailItem extends events.EventEmitter {
         }
     }
 
-    try_deliver () {
+    try_deliver (previous_mx = null) {
         const self = this;
 
         // check if there are any MXs left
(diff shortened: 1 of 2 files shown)


outbound/hmail.js
```diff
@@ -1267,7 +1281,10 @@ class HMailItem extends events.EventEmitter {
             'fails': this.num_failures,
             'rcpts': `${ok_recips.length}/${fail_recips.length}/${bounce_recips.length}`
         });
-        plugins.run_hooks("delivered", this, [host, ip, response, delay, port, mode, ok_recips, secured, authenticated]);
+
+        const hook_params = [host, ip, response, delay, port, mode, ok_recips, secured, authenticated];
+        hook_params.bind_ip = mx.bind;
```
baudehlo: You're adding a key to an array. I don't think that'll work.

celesteking: It's Javascript, it works. As noted it's a hack and it's better to change the entire hook args list. Changing the hook arg list means a breaking change I guess. Suggestions are welcome.

msimerson: You can make changes without them being breaking. For example, you can check the arg list length. Better still, pass in objects instead and then test what's being passed in, with an unwrapper function for legacy invocations.

msimerson:
Closing as abandoned.

PR #2432, opened 2018-05-18
mods to smtp_client.js

celesteking:
* Support specifying max mails per pool instance.  Once it reaches that number, it shutdowns and, generally, next one is picked/created. Good for avoiding mem leaks and long running connections, for whatever reason.
  * Allow configurable next() action on server error via `error_action`.
  * We introduce a shutdown state. We trigger it on destroy action. (So many years have passed that I'm not sure what was the reason for introducing, but I can tell you for sure there **was** a reason for that. I saw client misbehaving in certain circumstances and additional state was needed to get it to transition nicely into void state. Also, it might've been used for self-destroy on max mails, I'm not sure)

* don't log periodic pool availability messages

Oh yeah, all this works nicely in tandem with your beloved smtp_forward/proxy plugin, assuming it knows how to use the provided features.

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/smtp_client.js
+++ b/smtp_client.js
@@ -21,16 +21,20 @@ const line_socket = require('./line_socket');
 const logger      = require('./logger');
 const HostPool    = require('./host_pool');
 
+const DSN = require('haraka-dsn');
+const constants = require('haraka-constants');
+
 const smtp_regexp = /^([0-9]{3})([ -])(.*)/;
 const STATE = {
     IDLE: 1,
     ACTIVE: 2,
     RELEASED: 3,
     DESTROYED: 4,
+    SHUTDOWN: 5,
 }
 
 class SMTPClient extends events.EventEmitter {
-    constructor (port, host, connect_timeout, idle_timeout, socket) {
+    constructor (port, host, connect_timeout, idle_timeout, socket, max_mails) {
         super();
         this.uuid = utils.uuid();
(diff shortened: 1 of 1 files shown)


smtp_client.js
```diff
@@ -256,6 +273,52 @@ class SMTPClient extends events.EventEmitter {
...
+        this.removeAllListeners('helo');
+        this.removeAllListeners('mail');
+        this.removeAllListeners('rcpt');
+        this.removeAllListeners('data');
+        this.removeAllListeners('dot');
+        this.removeAllListeners('rset');
+        this.removeAllListeners('auth');
+        this.removeAllListeners('error');
+        this.removeAllListeners('bad_code');
+        this.removeAllListeners('rset');
```
msimerson: most of this `shutdown()` is a copy/paste of `release()`. Except the errors, like listing 'rset' twice. Not only should it be refactored to avoid duplicate code, but it should also be refactored for hygiene. Something like this is a fair bit cleaner IMO:

```js
['greeting', 'capabilities', 'xclient', 'helo', 'mail', 'rcpt', 'data', 'dot', 'rset', 'auth', 'error', 'bad_code'].forEach(l => {
    this.removeAllListeners(l);
})
```

msimerson:
Closing as abandoned. Can be reopened if/when the author responds.

Issue #2439, opened 2018-06-02
Mails not sent to some domains, SSL23_GET_SERVER_HELLO:sslv3 alert handshake Error


vincowl:
While having installed Poste.io using the latest free docker image (analogic/poste.io) which relies on Haraka, I set up my domain, SSL certs through given Let'sEncrypt option. Everything runs fine, I can send and receive mails to and from gmail.com for instance. But as I want to send mails to free.fr (french provider), I have the following error log, and the mail is not sent. I can receive mails from free.fr.

I tested my domain using mxtoolbox.com : everything's fine.... Any idea ?? I am a bit lost : is this an error coming from my container or from free.fr servers ? 

Here is the log from log/delivery/del/...

```
2018-06-02T14:56:13.875Z [1] [DEBUG] [outbound] running send_email hooks
2018-06-02T14:56:13.876Z [1] [DEBUG] [outbound] Sending mail: 1527951373843_1527951373843_0_611_5hphaG_7_<myhostname>
2018-06-02T14:56:13.876Z [1] [DEBUG] [outbound] running get_mx hooks
2018-06-02T14:56:13.876Z [1] [DEBUG] [outbound] running get_mx hook in relay plugin
2018-06-02T14:56:13.876Z [1] [DEBUG] [outbound] hook=get_mx plugin=relay function=force_routing params=free.fr retval=CONT msg=""
2018-06-02T14:56:13.876Z [1] [DEBUG] [outbound] running get_mx hook in rcpt_database plugin
2018-06-02T14:56:13.878Z [1] [DEBUG] [outbound] hook=get_mx plugin=rcpt_database function=hook_get_mx params=free.fr retval=CONT msg=""
2018-06-02T14:56:13.878Z [1] [DEBUG] [outbound] running get_mx hook in queue/lmtp plugin
2018-06-02T14:56:13.878Z [1] [DEBUG] [outbound] hook=get_mx plugin=queue/lmtp function=hook_get_mx params=free.fr retval=CONT msg=""
2018-06-02T14:56:13.912Z [1] [INFO] [outbound] Looking up A records for: mx1.free.fr
2018-06-02T14:56:13.914Z [1] [INFO] [outbound] Attempting to deliver to: 212.27.48.7:25 (0) (6)
2018-06-02T14:56:13.981Z [1] [PROTOCOL] [outbound] S: 220 mx1-g20.free.fr\r\n
[...]

PR #2441, opened 2018-06-05
Slightly changing example tls.ini to more "compatible" to real world

analogic:
Ref #2439

### DIFF
--- a/config/tls.ini
+++ b/config/tls.ini
@@ -1,5 +1,6 @@
 ; See 'haraka -h tls'
 
+[inbound]
 ; key=tls_key.pem
 ; cert=tls_cert.pem
 ; dhparam=dhparams.pem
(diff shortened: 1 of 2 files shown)


config/tls.ini
```diff
@@ -1,5 +1,6 @@
 ; See 'haraka -h tls'
 
+[inbound]
```
msimerson: Won't this change the following settings paths? IE, instead of `main.key`, they'll be `inbound.key`, and thus break them?

analogic: Previously my *whole* tls.ini looked like this:

```
ciphers=...
```

Now I am running this:
```
[inbound]
ciphers=...
[outbound]
ciphers=DEFAULT
```
Default keys are still loaded from same place at least in my case. I looked at tls_socket.js and it seems that section name has no influence to paths also

analogic:
Closing since It seems I don't fully understand how fallback to defaults work.

I would rather set only outbound and leave main as is, but [main]+[outbound] is somewhat not working contrary code:
https://github.com/haraka/Haraka/blob/d52c99a80b52ac8dbef4585a4915e8777b7a6501/outbound/tls.js#L19
...and this should set defaults from:
https://github.com/haraka/Haraka/blob/d52c99a80b52ac8dbef4585a4915e8777b7a6501/tls_socket.js#L310

If I set [inbound]+[outbound] default keys are still loaded without error.

PR #2446, opened 2018-06-13
Escape values in html bounce properly

analogic:
Fixes:
- "hidden" rcpt_tos in html bounce due < brackets >
- XSS via email is AFAIK impossible (clients will parse out scripts) but bounce content can be damaged with some chars

- move bounce values to top and for html template replace with escaped variables

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -995,7 +995,7 @@ class HMailItem extends events.EventEmitter {
         bounce_msg_html_.forEach(function (line) {
             line = line.replace(/\{(\w+)\}/g, function (i, word) {
                 if (word in values) {
-                    return String(values[word]).replace(/[&<>]/g, function(m) { return escaped_chars[m]; });
+                    return String(values[word]).replace(/[&<>]/g, function (m) { return escaped_chars[m]; });
                 } else {
                     return '?';
                 }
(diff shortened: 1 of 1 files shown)


outbound/hmail.js
```diff
@@ -966,8 +986,22 @@ class HMailItem extends events.EventEmitter {
             }
         });
 
+        const escaped_chars = {
+            "&": "&amp;",
+            "<": "&lt;",
+            ">": "&gt;"
+        };
+
```
baudehlo: I'd suggest going a bit further:

```javascript
	const escaped_chars = {
		'<': 'lt',
		'>': 'gt',
		'"': 'quot',
		'\'': 'apos',
		'&': 'amp',
		'\r': '#10',
		'\n': '#13'
	};
```
(I left out the `&` and `;` since you can add them once in the replacement).

analogic: I can't imagine why anyone would use variables for tag parameters and need escaping quotes or apostrophes but I see no problem with that. I am slightly opposed with escaping \r\n - itself it don't break anything, even after escaping pre tag is required to properly view lines...

analogic: ...or can it break email itself? Something like \r\n\r\n.\r\n?

analogic:
Do you mind merging this or is there any problem? These bounces looks silly
eh

PR #2456, opened 2018-06-28
Removed invalid opt from load_tls_ini()

### LINKED ISSUE #2455: tls plugin isn't loading correctly on 2.8.19
### system info

echo "Haraka | $(haraka -v)"; echo " --- | :--- "; echo "Node | $(node -v)"; echo "OS | $(uname -a)"; 
```
Haraka | Haraka.js — Version: 2.8.19
 --- | :--- 
Node | v6.14.2
OS | Linux mail.example.com 3.10.0-862.3.3.el7.x86_64 #1 SMP Fri Jun 15 04:15:27 UTC 2018 x86_64 
x86_64 x86_64 GNU/Linux
```

echo "openssl | $(openssl version)"
`openssl | OpenSSL 1.0.2k-fips  26 Jan 2017`

### Expected behavior
After updating from 2.8.18, server should continue to work as normal.

### Observed behavior
TLS plugin failes to load correctly with this on the logs:
```
Jun 26 23:34:47 mail.example.com haraka[20396]: [ERROR] [-] [core] tls key /haraka/config/etc/letsencrypt/live/mail.example.com/privkey.pem could not be loaded.
Jun 26 23:34:47 mail.example.com haraka[20396]: [ERROR] [-] [core] tls cert /haraka/config/etc/letsencrypt/live/mail.example.com/fullchain.pem could not be loaded.
[...]
Jun 27 00:10:36 mail.example.com haraka[22101]: [ERROR] [-] [tls] no valid TLS config
```

### Steps to reproduce
Upgrade to 2.8.19 with a valid TLS config.

tls.ini as follows:
# cat tls.ini
```
cert=/etc/letsencrypt/live/mail.example.com/fullchain.pem
key=/etc/letsencrypt/live/mail.example.com/privkey.pem
ciphers=EECDH+AESGCM:EDH+aRSA+AESGCM:EECDH+AES256:EDH+aRSA+AES256:EECDH+AES128:EDH+aRSA+AES128:RSA+AES:RSA+3DES
dhparam=tls/dhparam.pem
```

analogic:
Fixes #2455

- removed invalid argument from load_tls_ini which prevents properly load defaults due change at #2414 https://github.com/haraka/Haraka/pull/2414/files#diff-7b009c1421c1a91ef0875d92e1159e1aR259

### DIFF
--- a/plugins/tls.js
+++ b/plugins/tls.js
@@ -23,9 +23,7 @@ exports.shutdown = function () {
 exports.load_tls_ini = function () {
     const plugin = this;
 
-    plugin.cfg = tls_socket.load_tls_ini(function () {
-        plugin.load_tls_ini();
-    });
+    plugin.cfg = tls_socket.load_tls_ini();
 }
 
 exports.advertise_starttls = function (next, connection) {


analogic:
It should be merged asap, TLS loading is broken by this bug

baudehlo:
Does this break reloading tls.ini when it changes?

analogic:
Yes, this callback was never called even before...

msimerson:
> I'm just wondering if we want to support a callback for that purpose?

`tls_socket.load_tls_ini();` already has the callback within it, so that's already done.

baudehlo:
I disagree that it's already done. It won't change `plugin.cfg` on reload
because it'll be a new object. It will change tlss.cfg:
https://github.com/msimerson/Haraka/blob/bd1f3380a7d6ffc3fb5e6a3754dd328b0cd0e823/tls_socket.js#L257
but not plugin.cfg because it will be a new object.

msimerson:
We should test that. I'm expecting that plugin.cfg will be a reference to tlss.cfg, so when `tlss.cfg` gets updated, so too does `plugin.cfg`. You're expecting that plugin.cfg will remain a reference to the first `tlss.cfg` object. I think you're likely right. I'll address this in a way, that fits in with my "get rid of plugins/tls entirely" plan.

baudehlo:
It's a new object:
https://github.com/haraka/haraka-config/blob/master/config.js#L34

msimerson:
@baudehlo, do you agree that https://github.com/haraka/Haraka/pull/2459 suitable addresses the issue?

baudehlo:
I think so.

PR #2459, opened 2018-06-28
remove tls.ini loading from plugins/tls

msimerson:


msimerson:
> Does this still cope fine with the [inbound] section?

No change. This only affects config loading, it doesn't affect whether other things look for specific sections in tls.ini.

baudehlo:
Since this seems to be breaking things, we should release again.

msimerson:
completely agree. I've a couple more bug fix PRs awaiting review. Once those merge I'll be inclined to create a release PR.

PR #2477, opened 2018-07-19
Clear timeout on cancel

smfreegard:
Finally tracked down a long standing annoyance that I've had with TLS.   If the client drops the connection after TLS negotiation fails, then the hook timer is never cancelled and you end up with a load of 'plugin tls timed out on hook unrecognized_command' messages in the logs.

### DIFF
--- a/plugins.js
+++ b/plugins.js
@@ -433,7 +433,10 @@ plugins.run_next_hook = function (hook, object, params) {
     let cancelled = false;
     let item;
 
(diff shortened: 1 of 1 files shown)


baudehlo:
Good catch. Looks like we have a few critical enough bugs to do a release. Thoughts?

msimerson:
Definitely, I've been wanting to get a deployment from @HEAD into production testing and just got that done today.

PR #2482, opened 2018-07-21
Delete qfiles if outbound queue timeout occurs

### LINKED ISSUE #2325: Performance problem with Outbound: large emails with lots of recipients cause major problems
smf pointed this out on irc.

It’s a weakness of the current outbound code that the email is split by domain into individual queue files. If the mail is large and to a mailing list (eg 1MB email with 100 recipients) you do a LOT of I/O.

This can actually cause a catastrophic failure: the mail starts to get written, the queue_outbound hook times out and the sender gets a temp fail. But the outbound work has started so you get huge duplicates.

Options: 

- write the mail once and use locking to update the todo part to remove “done” recipients.
- write a separate todo file like qmail does (qmail writes 3 files - we should research what they are)
- use a local storage db (level db?) instead which can take care of the nitty gritty details. Wild duck uses Mongo but I don’t want to go that far.

Furthermore we need to deal with that timeout sanely by nuking the outbound files (remote end could timeout so we need to cope there too).

baudehlo:
Fixes #2325 - maybe - hard to test

- Delete qfiles on timeout

There might be other ways this can timeout than the ways specified by smf - so this is just posted for discussion first.

Checklist:
- [ ] docs updated
- [ ] tests updated


outbound/index.js
```diff
@@ -281,6 +281,7 @@ exports.send_trans_email = function (transaction, next) {
                 delivery_queue.push(hmail);
             }
 
+            transaction.qfiles = qfiles;
```
baudehlo: Sorry to take so long to get to this review...

Anyway we need to push onto this array every time process_delivery finishes, rather than do it at the end. Otherwise we have to wait for them all to finish that function.

celesteking:
2 years later...

msimerson:
4 years. Moved to [Outbound Improvementsi](https://github.com/haraka/Haraka/wiki/Outbound-Improvements) wiki page.

Issue #1532, opened 2016-07-08
Custom HELO/EHLO message


xpepermint:
As I can see from the source, Haraka v2.8.7 currently doesn't have support for EHLO response message customization. I don't want to tell a hacker what SMTP server I'm using so I would like to hide that information by replacing embeded `Haraka` messages (e.g. [Haraka is at your service.](https://github.com/haraka/Haraka/blob/master/connection.js#L787)) with my custom messages. Is there a specific reason that these messages are not configurable? Can we add this option please?

baudehlo:
Patches for this are welcome. I would welcome this being configurable in a flexible manner.

msimerson:
moved to [https://github.com/haraka/Haraka/wiki#feature-requests](https://github.com/haraka/Haraka/wiki#feature-requests)

xpepermint:
ok

Bramzor:
Reopening to create a pull request for this one

Dexus:
Create a PR you don't need to reopen this here...

PR #2498, opened 2018-08-21
#1532 Possibility to hide Haraka is at your service from HELO/EHLO for security reasons.

Bramzor:
Possibility to hide Haraka is at your service from HELO/EHLO for security reasons.
Added the option to just hide it instead of adding your own message because this might break more than it solves. 

Excluding the "Haraka is at your service" message from HELO/EHLO message is as easy as creating a file config/ehlo_hello_exclude_haraka with the value "true" which is a standard method in this file.

Fixes #

- #1532 

Checklist:
- [na] docs updated
- [na] tests updated

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -100,7 +100,7 @@ class Connection {
         this.transaction = null;
         this.tran_count = 0;
         this.capabilities = null;
-        this.ehlo_hello_exclude_haraka = config.get('ehlo_hello_exclude_haraka') ? true : false;
+        this.ehlo_hello_message = config.get('ehlo_hello_message') || null;
         this.banner_includes_uuid = config.get('banner_includes_uuid') ? true : false;
         this.deny_includes_uuid = config.get('deny_includes_uuid') || null;
         this.early_talker = false;
@@ -868,7 +868,7 @@ class Connection {
             default:
                 // RFC5321 section 4.1.1.1
                 // Hostname/domain should appear after 250
-                this.respond(250, `${this.local.host} Hello ${this.get_remote('host')}${this.ehlo_hello_exclude_haraka ? '': ', Haraka is at your service.'}`);
+                this.respond(250, `${this.local.host} Hello ${this.get_remote('host')}${this.ehlo_hello_message ? this.ehlo_hello_message: ', Haraka is at your service.'}`);
         }
     }
     ehlo_respond (retval, msg) {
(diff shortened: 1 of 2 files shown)


Bramzor:
@baudehlo Security best practices... No reason to show what software is used or what version. I understand why you want to there, therefore it has been made optional.

msimerson:
Instead of Yet Another Config File that lacks documentation, how about adding a property to smtp.ini?

```ini
; smtp_announce = Haraka is at your service
smtp_announce = Bramzor's Secure SMTPd is ready
```

If it's not present in smtp.ini (for legacy installs), set it to 'Haraka is at your service' in server.js so its always set. Then in connection.js, reference `server.cfg.smtp_greeting` instead of the static string.

Bramzor:
Im not using smtp.ini so I would not put it in that file. Why not starting a generic config file? config/config or something where you define these settings. I agree that you dont want another config file but I just used the method that already existed.

So if we agree on which config file I can do the update. We should actually end up with one file with a lot of settings (and all commented) so you can basically just read through the config file and uncomment what you need.

baudehlo:
No I'm saying why do you need this extra way of doing this vs adding
something to config/smtpgreeting

msimerson:
@baudehlo , they are different places where we leak the SMTPd identity.

````
 $ nc mail.theartfarm.com 25
220 mail.theartfarm.com ESMTP Haraka/2.8.21 ready  <-- smtpgreeting
EHLO home.simerson.net
250-mail.theartfarm.com Hello [2602:47:...:175c], Haraka is at your service.  <-- HELO response
250-PIPELINING
250-8BITMIME
250-SMTPUTF8
250-SIZE 45000000
250 STARTTLS
quit
221 mail.theartfarm.com closing connection. Have a jolly good day.
````

**Comment by Bramzor** [comment]
Didnt notice the smtp greeting yet. So there are indeed 2 different parameters. So I think 2 files are acceptable or we need to start 1 bigger configuration file in a specific structure. I will make a change that instead of just hiding that message, you can update the response.

File: connection.js [hunk 1]
```diff
@@ -100,6 +100,7 @@ class Connection {
         this.transaction = null;
         this.tran_count = 0;
         this.capabilities = null;
+        this.ehlo_hello_message = config.get('ehlo_hello_message') || null;
```
msimerson: This is an improvement, but wouldn't this be better?

```js
this.ehlo_hello_message = config.get('ehlo_hello_message') || 'Haraka is at your service.'
```

Then the two ternaries below are simple variable references.

connection.js
```diff
@@ -100,6 +100,7 @@ class Connection {
         this.transaction = null;
         this.tran_count = 0;
         this.capabilities = null;
+        this.ehlo_hello_message = config.get('ehlo_hello_message') || 'Haraka is at your service.'
```
spathon: The new default Haraka message is missing `, ` from the previous.
`Hello [127.0.0.1]Haraka is at your service.`

Bramzor: What you want to do is check if ehlo_hello_message is set and if so, add ', ' to the message. Although in that case someone is forced to use that ', ' as it will be static. Other solution is to add ', ' to this to the default text.

spathon: In this case it would probably be easiest with just add it to the default message?

PR #2522, opened 2018-10-21
relay: update port 465 doc

msimerson:
Checklist:
- [x] docs updated
- [ ] tests updated


msimerson:
@haraka/core , still need a review.

PR #2524, opened 2018-10-21
ob/tls: use obtls (vs plugin) for "this" name

msimerson:
ob/tls: more concise flow control

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/outbound/tls.js
+++ b/outbound/tls.js
@@ -13,7 +13,7 @@ const inheritable_opts = [
 class OutboundTLS {
     constructor () {
         this.config = config;
-        this.name = "OutboundTLS";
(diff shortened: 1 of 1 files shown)


baudehlo:
Shouldn’t we just be using the fact that `this` is still valid inside of arrow functions now we use them?

msimerson:
> Shouldn’t we just be using the fact that `this` is still valid inside of arrow functions

So long as it's unambiguous, maybe. If you are two or three levels deep, you're going to have to read a fair bit of surrounding code to be sure you know exactly what `this` is. Within class files, there's some intrinsic value in having a named `this` that is used consistently.

PR #2525, opened 2018-10-24
Update license

AuspeXeu:


smfreegard:
"Haraka" isn't an entity - so I don't see how it can hold copyright.

AuspeXeu:
Fair point. Fixed.

PR #2529, opened 2018-11-02
move max_mime_parts config load to connection init

### LINKED ISSUE #2528: Low performance when accepting large mails?
I've hit this problem by sending larger email (like 20mb attachments) with roundcube which timeouted. 

Debug with empty email with 16MB attachment:

```
2018-11-02 09:49:34  ⭠ DATA
2018-11-02 09:49:34  [core] [DEBUG] - running data hooks
2018-11-02 09:49:34  ⭢ 354 go ahead, make my day
... 18 seconds of upload and node at 100% CPU ...
2018-11-02 09:49:52  [core] [DEBUG] - running data_post hooks
```
There are no plugins which can limit "upload" rate enabled and there should not be any parsing happening afaik. So it gives me that email is uploaded at rate < 1MB/s from localhost to localhost with enough ram and i7 cpu. Is there anything what I should look for to tweak or I have simply unreal expectations?

analogic:
Fixes #2528

removing huge penalty per email line

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -112,6 +112,7 @@ class Connection {
         this.start_time = Date.now();
         this.last_reject = '';
         this.max_bytes = parseInt(config.get('databytes')) || 0;
+        this.max_mime_parts = config.get('max_mime_parts') || 1000;
         this.totalbytes = 0;
         this.rcpt_count = {
             accept:   0,
@@ -1597,8 +1598,7 @@ class Connection {
             return;
         }
 
-        const max_mime_parts = config.get('max_mime_parts') || 1000;
-        if (this.transaction.mime_part_count >= max_mime_parts) {
+        if (this.transaction.mime_part_count >= this.max_mime_parts) {
             this.logcrit("Possible DoS attempt - too many MIME parts");
             this.respond(554, "Transaction failed due to too many MIME parts", function () {
                 self.disconnect();
--- a/connection.js
+++ b/connection.js
@@ -112,7 +112,7 @@ class Connection {
         this.start_time = Date.now();
         this.last_reject = '';
         this.max_bytes = parseInt(config.get('databytes')) || 0;
-        this.max_mime_parts = config.get('max_mime_parts') || 1000;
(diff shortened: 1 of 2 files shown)


connection.js
```diff
@@ -112,6 +112,7 @@ class Connection {
         this.start_time = Date.now();
         this.last_reject = '';
         this.max_bytes = parseInt(config.get('databytes')) || 0;
+        this.max_mime_parts = config.get('max_mime_parts') || 1000;
```
smfreegard: Can you make this a parseInt() like above.

smfreegard:
Great catch!

connection.js
smfreegard: Oh - and can you update the CHANGES document with this too please.

connection.js
```diff
@@ -112,6 +112,7 @@ class Connection {
         this.start_time = Date.now();
         this.last_reject = '';
         this.max_bytes = parseInt(config.get('databytes')) || 0;
+        this.max_mime_parts = parseInt(config.get('max_mime_parts')) || 1000;
```
baudehlo: If `parseInt` produces `NaN` this will satisfy the `||` - so move the `|| 1000` into the parseInt.

Also never use parseInt raw like that - always use it with the `, 10` parameter.

And yes I'm aware code you haven't touched doesn't follow either of these rules. Maybe you could fix it all?

msimerson: Perhaps also add the [eslint rule](https://eslint.org/docs/rules/radix)?

baudehlo:
> Oh, I think it would be better to just create something like
config.getInt(name) or config.getInt(name, default_val) to move
responsibility to config module

100%

connection.js
msimerson: See [code](https://github.com/haraka/haraka-config/pull/44/files#diff-e22b50e740e102c8de5e858dbcd7753dR52) and [tests](https://github.com/haraka/haraka-config/pull/44/files#diff-447140de3e1b662ff6e9201e7352b529R406) in [haraka-config](
https://github.com/haraka/haraka-config).

PR #2531, opened 2018-11-02
hmail: log the correct err message

msimerson:


outbound/hmail.js
```diff
@@ -102,9 +102,9 @@ class HMailItem extends events.EventEmitter {
                 if (todo_bytes.length !== todo_len) {
                     const wrongLength = "Didn't find right amount of data in todo!"
                     this.logcrit(wrongLength);
```
baudehlo: While you're in there - not sure why this: a) is a variable, and b) doesn't log err2.

msimerson: It's a variable because a few lines down it's also emitted as an error. I've added logging of err2.

PR #2532, opened 2018-11-03
Add connection.remote.is_local

analogic:
- add  connection.remote.is_local flag for detecting loopback and link_local addresses

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -11,6 +11,7 @@
 
 * ehlo_hello_message: config/ehlo_hello_message can be used to overwrite the EHLO/HELO msg replacing `, Haraka is at your service` #2498
 
+* connection: add connection.remote.is_local flag for detecting loopback and link local IPs
 
 ## 2.8.21 - Jul 20, 2018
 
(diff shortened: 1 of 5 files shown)


msimerson:
I can see the value of having a `connection.remote.is_local` populated. However, the way you've done it here means the code that tests for localhost gets called twice (once for `is_local` and again for `is_private`). Since one of the things this PR does is makes Haraka connections more efficient (by not scanning localhost mail), we should probably not add additional costs to every other connection.

I [created a PR](https://github.com/haraka/haraka-net-utils/pull/39) in haraka-net-utils that lets you call `net_utils.is_local_ip()`. For backwards compatibility, the existing `is_private_ip()` needs to continue checking both, but we can update the code in `connection.js` to first call `is_local_ip`, and then:

```js
if (this.remote.is_local) {
    this.remote.is_private = true;
}
else {
    if net.isIPv4(ip)) this.remote.is_private = net_utils.is_private_ipv4(ip);
}
```

That avoids the redundant checks for loopback IPs.

Do you concur?

analogic:
Yep, totally makes sense!

PR #2533, opened 2018-11-03
add clamd check options for skipping

analogic:
- added check.authenticated, check.private_ip, check.local_ip for ability to skip some connections (same as rspamd plugin)
- depends on https://github.com/haraka/Haraka/pull/2532

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/plugins/clamd.js
+++ b/plugins/clamd.js
@@ -172,10 +172,7 @@ exports.hook_data_post = function (next, connection) {
     const txn = connection.transaction;
     const cfg = plugin.cfg;
 
-    if (!plugin.should_check(connection)) {
-        txn.results.add(plugin, {skip: 'criteria match'});
-        return next();
-    }
+    if (!plugin.should_check(connection)) return next();
 
     // Do we need to run?
     if (cfg.main.only_with_attachments && !txn.notes.clamd_found_attachment) {
--- a/plugins/clamd.js
+++ b/plugins/clamd.js
@@ -90,6 +90,10 @@ exports.load_clamd_ini = function () {
             // clamd.conf options enabled by default, but prone to false
             // positives.
             '-reject.Phishing',
+
+            '+check.authenticated',
+            '+check.private_ip',
+            '+check.local_ip'
         ],
     }, function () {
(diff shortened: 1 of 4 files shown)


tests/plugins/clamd.js
```diff
@@ -96,6 +96,36 @@ exports.hook_data_post = {
         }.bind(this);
         this.plugin.hook_data_post(next, this.connection);
     },
+    'skip authenticated': function (test) {
+        this.connection.notes.auth_user = 'user';
+        this.plugin.cfg.check.authenticated = false;
+        test.expect(1);
+        const next = function () {
+            test.ok(this.connection.transaction.results.get('clamd').skip);
```
msimerson: IIRC, skip is an array, so don't we have to check length?

```js
> const foo = [];
> if (foo) console.log('true')
true
```

plugins/clamd.js
```diff
@@ -315,6 +327,18 @@ exports.hook_data_post = function (next, connection) {
     try_next_host();
 }
 
+exports.wants_skip = function (connection) {
+    const plugin = this;
+
+    if (!plugin.cfg.check.authenticated && connection.notes.auth_user) return true;
+
+    if (connection.remote.is_local) return !plugin.cfg.check.local_ip;
```
msimerson: Same here, lets stay consistent with the code surrounding this line. The biggest reason is future maintainability. Someone else might add another condition after this and be surprised when it sometimes doesn't run.

```suggestion
    if (plugin.cfg.check.local_ip && connection.remote.is_local) return true;
```

I'm half tempted to say, move the `txn.results.add(plugin, {skip: 'criteria match'});` code into this function, so that the skip message can be a more precise reason (authed, localhost, private).

msimerson:
Hello @analogic, I'm pushing towards a release. This still has the "hard to reason about" code that we discussed in the rspamd repo. Maybe that same approach that we (I think) found better should be used here too?

analogic:
Sorry, I was bit busy lately. I have applied almost same changes we have done at rspamd plugins.

PR #2543, opened 2018-11-19
Add tls authorizationRequired option

iczero:
- Add tls `authorizationRequired` option which will allow specifying only certain ports on which to require a valid certificate. This allows smtp clients to send emails outbound without a valid certificate but still ensure a valid certificate for incoming mail.

Checklist:
- [x] docs updated
- [x] tests updated


iczero:
Oh huh. I blame npm, sorry about that

msimerson:
I'm not sure changing the name to `validCertRequired` is an improvement. This PR and the new config setting overlap with `rejectUnauthorized` and understanding how the two interact is more complex with this PR.

A different option would be permitting `rejectUnauthorized` to be a boolean or an Array. When it's an array, do port number matching.

Another option is deprecating `rejectUnauthorized` entirely, and using it as the basis to populate a newer config setting: `requireAuthorized` or `requireValidated`. Positive assertions tend to be easier to reason about than negative ones.

msimerson:
Hey @iczero , have a look at #2554 and post your thoughts.

PR #2547, opened 2018-11-25
Implement SIGTERM graceful shutdown if pid is 1

jkroepke:
Related: https://blog.ghaiklor.com/avoid-running-nodejs-as-pid-1-under-docker-images-when-running-them-on-mesosphere-kubernetes-or-b7bd505657f9

Like in the blog described node has some signal issues if it runs as pid 1 (usually inside a container).
Currently on docker stop docker sends SIGTERM to node which will be ignored because not handled somewhere.

Haraka on docker or even on kubernetes has to be killed unless you use workarounds like `tini`. I would prefer that haraka initiate a graceful shutdown on docker stop.

- Implement SIGTERM graceful shutdown if pid is 1

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/haraka.js
+++ b/haraka.js
@@ -38,7 +38,13 @@ process.on('uncaughtException', err => {
 });
 
 let shutting_down = false;
-['SIGINT'].forEach((sig) => {
+const signals = ['SIGINT'];
+
+if (process.pid === 1) {
+    signals.push('SIGTERM')
+}
+
+signals.forEach((sig) => {
     process.on(sig, () => {
         if (shutting_down) return process.exit(1);
         shutting_down = true;
--- a/Changes.md
+++ b/Changes.md
@@ -1,6 +1,7 @@
 
 ## 2.8.23 - Mmm DD, 201Y
 ### Changes
+* Implement SIGTERM graceful shutdown if pid is 1 #2547
 ### New Features
 ### Fixes
 


jkroepke:
Did I something wrong? -- tests are broken.

msimerson:
If you follow the links to the test output, you'll find the issue:

````
/home/travis/build/haraka/Haraka/haraka.js
  41:5  error  'signals' is never reassigned. Use 'const' instead  prefer-const
  43:1  error  Expected space(s) after "if"                        keyword-spacing
````

File: haraka.js [hunk 1]
```diff
@@ -38,7 +38,13 @@ process.on('uncaughtException', err => {
 });
 
 let shutting_down = false;
-['SIGINT'].forEach((sig) => {
+let signals = ['SIGINT'];
```
msimerson:
```suggestion
const signals = ['SIGINT'];
```

haraka.js
```diff
@@ -38,7 +38,13 @@ process.on('uncaughtException', err => {
...
+
+if(process.pid === 1) {
```
msimerson:
```suggestion
if (process.pid === 1) {
```

jkroepke:
I force-pushed into the branch with lint fixes applied. I also add a notice at the Changes.md

PR #2554, opened 2018-12-07
support requiring verified TLS certs on specific ports

msimerson:
Fixes #2543 

This is largely the same as #2543 except:

- includes tests (which surfaced a couple issues)
- config setting named `requireAuthorized` instead of `authorizationRequired`
- no boolean option (simpler, easier to document)
- has support for SMTPS server (was only for STARTTLS)
- more GFM formatting in tls.ini

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/config/tls.ini
+++ b/config/tls.ini
@@ -15,6 +15,12 @@
 ; requestCert=true
 ; requestOCSP=false
 
+; rejectUnauthorized above requires verified TLS certs on EVERY TLS connection. When
+; rejectUnauthorized=false (default), you can require verified TLS certs on only the
+; ports you specify.
+; requireAuthorized[]=465
+; requireAuthorized[]=587
+
 
 [redis]
 ; options in this block require redis to be enabled in config/plugins.
--- a/plugins/queue/discard.js
+++ b/plugins/queue/discard.js
@@ -1,7 +1,7 @@
 // discard
 
 exports.register = function () {
-    this.register_hook('queue', 'discard');
+    this.register_hook('queue',          'discard');
     this.register_hook('queue_outbound', 'discard');
 }
 
@@ -24,5 +24,5 @@ exports.discard = function (next, connection) {
     if (process.env.YES_REALLY_DO_DISCARD) return discard();
 
     // Allow other queue plugins to deliver
-    return next();
+    next();
 }
--- a/Changes.md
+++ b/Changes.md
@@ -12,13 +12,17 @@
 
 * Implement SIGTERM graceful shutdown if pid is 1 #2547
 
+### New Features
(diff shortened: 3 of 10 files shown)


plugins/queue/discard.js
```diff
@@ -1,7 +1,7 @@
 // discard
 
 exports.register = function () {
-    this.register_hook('queue', 'discard');
+    this.register_hook('queue',          'discard');
```
AuspeXeu: Is this whitespace needed?

config/tls.ini
```diff
@@ -15,6 +15,12 @@
 ; requestCert=true
 ; requestOCSP=false
 
+; rejectUnauthorized above requires verified TLS certs on EVERY TLS connection. When
+; rejectUnauthorized=false (default), you can require verified TLS certs on only the
```
AuspeXeu: Wouldn't it be more secure to have this default to true?

msimerson: Yes, and unplugging the network cable would also be more secure. Neither one is very useful for MTA operators wanting to receive emails.

plugins/queue/discard.js
msimerson: Definitely not needed. It makes no difference at all to the JS engine. So sometimes we optimize code for humans to read, and vertical alignment of identical items makes it easier to see at-a-glance that both registered hooks call the same function.

config/tls.ini
msimerson: Typical connections that send mail to a MTA will not present a valid TLS client certificate. If you set `rejectUnauthorized=true`, then you'll block the vast majority of mail. This is really an "edge case" feature, where you have clients that will authenticate to Haraka with a signed certificate instead of user/pass.

AuspeXeu: Yeah, I shot a bit quickly and was only thinking about using Haraka as an outbound mail server. In this case it would only send emails to other MTAs that support TLS with this option, no?

msimerson: Likely the best way to make this point is to refer you to your Haraka mail logs with grep. Here's a few of my connections:

````
# grep verified=false /var/log/maillog | wc -l             
    1165
# grep verified=true /var/log/maillog | wc -l
      11
````

msimerson: > In this case it would only send emails to other MTAs that support TLS with this option, no?

Not quite. Haraka outbound would only be able to send mail to remote MTAs that present valid (is, verified by a CA that's included in the CAs that are bundled with Node.js) TLS certificates. This would be a higher than for inbound, but I think you'd still end up with lots of undelivered mail in your outbound queue.

PR #2559, opened 2018-12-09
release client when successfully finished delivery

### LINKED ISSUE #2556: LMTP pool blocks under stress
I have noticed strange behaviour on one installation where nagios generated hunderds of emails at one time which are sent through haraka to local LMTP server. When that happend haraka after ~70 emails stopped delivering to LMTP with

```
[outbound] [INFO] - Attempting to deliver to: 127.0.0.1:24 using LMTP (0) (3)
[outbound] [INFO] - Temp failing 1544181767545_1544181767545_0_651_dgzLB6_41_mail.example.com for 64 seconds: Tried all MXs
```
It seems ok - LMTP was exhausted at the time - but it contiued even after load was essentialy zero
```
[outbound] [INFO] - Temp failing 1544181767545_1544181831613_1_651_dgzLB6_46_mail.example.com for 128 seconds: Tried all MXs
...
[outbound] [INFO] - Temp failing 1544181767545_1544181960330_2_651_dgzLB6_2901_mail.example.com for 256 seconds: Tried all MXs
...
[outbound] [INFO] - Temp failing 1544181767545_1544182216353_3_651_dgzLB6_1116_mail.example.com for 512 seconds: Tried all MXs
```
Important is that Dovecots LMTP was working ok, another instance of haraka had no problem with delivery.

Second thing is that pool seems to be blocked forever I guess - we have noticed issue after 8 hours.

It can be fairly easily reproduced with something like
```
#!/bin/bash

for IT in {0..999}
do
    swaks --to rec1@example.com,rec2@example.com --server localhost > /dev/null &
    swaks --to rec1@example.com,rec2@example.com --server localhost > /dev/null &
    swaks --to rec1@example.com,rec2@example.com --server localhost > /dev/null &
    swaks --to rec1@example.com,rec2@example.com --server localhost > /dev/null &
    swaks --to rec1@example.com,rec2@example.com --server localhost > /dev/null
done
```

analogic:
Fixes #2556

There is question with this PR: do we need to send RSET after finished email? Most servers I've tested have no problem with that, RFC is clear about using it, LMTP doesn't have rset at all. But still there is maybe corner case which I don't know.

- RSET is not sent when SMTP|LMTP delivery is finished with success and we release client immediately
- if delivery ends without success then we send QUIT to LMTP and RSET to SMTP

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -618,10 +618,12 @@ class HMailItem extends events.EventEmitter {
             else {
                 self.discard();
             }
-            if (cfg.pool_concurrency_max && !mx.using_lmtp) {
+
+            if (cfg.pool_concurrency_max && success) {
+                client_pool.release_client(socket, port, host, mx.bind, fin_sent);
+            } else if (cfg.pool_concurrency_max && !mx.using_lmtp) {
                 send_command('RSET');
-            }
-            else {
+            } else {
                 send_command('QUIT');
             }
         }
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -621,9 +621,11 @@ class HMailItem extends events.EventEmitter {
 
             if (cfg.pool_concurrency_max && success) {
                 client_pool.release_client(socket, port, host, mx.bind, fin_sent);
-            } else if (cfg.pool_concurrency_max && !mx.using_lmtp) {
+            }
+            else if (cfg.pool_concurrency_max && !mx.using_lmtp) {
                 send_command('RSET');
-            } else {
+            }
+            else {
                 send_command('QUIT');
             }
         }
--- a/Changes.md
+++ b/Changes.md
@@ -15,6 +15,7 @@
 ### Fixes
 
 * mf.resolvable: reduce timeout by one second (so < plugin.timeout) #2544
+* LMTP blocks under stress #2556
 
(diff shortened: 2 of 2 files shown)


outbound/hmail.js
```diff
@@ -618,10 +618,12 @@ class HMailItem extends events.EventEmitter {
...
             }
-            if (cfg.pool_concurrency_max && !mx.using_lmtp) {
+
+            if (cfg.pool_concurrency_max && success) {
+                client_pool.release_client(socket, port, host, mx.bind, fin_sent);
+            } else if (cfg.pool_concurrency_max && !mx.using_lmtp) {
                 send_command('RSET');
-            }
-            else {
+            } else {
```
baudehlo: Don't do that.

msimerson: @baudehlo , see https://github.com/haraka/haraka-eslint/pull/16

analogic: Should be ok by now I guess?

msimerson: Yes, cuddled else's have shown up in several PRs in recent memory, so rather than having humans point it out, we can let eslint do the job.

baudehlo:
Yeah I don't know for certain it's the right thing, but let's try it.

PR #2564, opened 2018-12-12
adjust should_check logic, extend to spamassassin

msimerson:
I need a mechanism to skip spamassassin checks for IP based relaying clients. Haraka already has similar mechanisms for this in rspamd and clamd via their `check` config settings and the `should_check()` function. This PR (and [rspamd #13](haraka/haraka-plugin-rspamd#13)) extends `should_check()` to support relay clients.

Aside: auth clients **are** relaying clients, but relay might also be set by other plugins, such as IP lists in the `relay` plugin.

- adjust should_check logic to be flow-through
    - easier to read/parse/grok and extend
    - shows if a connection was skipped for more than one reason
- copy `should_check()` to spamassassin
- early_talker had similar inline checks which I refactored into `should_check()`

related to #2551  (skip early_talker if +karma)
related to haraka/haraka-plugin-rspamd#13
related to #2533 (add clamd skip options)

It might be useful to abstract `should_check()` into plugin.js and have a "standard" set of config settings that each plugin can skip processing for. At present, we have:

```ini
[check]
;authenticated=false
;relay=false
;private_ip=false
;local_ip=false
```

The `early_talker` plugin also has *karma* as a criteria, which could be useful elsewhere as well. 

Checklist:
- [ ] docs updated
- [x] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -11,10 +11,10 @@
 ### New Features
 
 * Implement SIGTERM graceful shutdown if pid is 1 #2547
-
-### New Features
-
 * tls: require validated certs on some ports with requireAuthorized
+* spamassassin: disable checks when requested
+* early_talker: extend reasons to skip checking
+* clamd: permit skipping for relay clients
 
 ### Fixes
 
--- a/Changes.md
+++ b/Changes.md
@@ -7,14 +7,14 @@
 * early_talker: skip if sender has good karma #2551
 * dockerfile: update to node 10 #2552
 * Update deprecated usages of Buffer #2553
+* early_talker: extend reasons to skip checking #2564
 
 ### New Features
 
 * Implement SIGTERM graceful shutdown if pid is 1 #2547
-* tls: require validated certs on some ports with requireAuthorized
-* spamassassin: disable checks when requested
-* early_talker: extend reasons to skip checking
-* clamd: permit skipping for relay clients
+* tls: require validated certs on some ports with requireAuthorized #2554
+* spamassassin: disable checks when requested #2564
+* clamd: permit skipping for relay clients #2564
 
 ### Fixes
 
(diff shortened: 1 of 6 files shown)


msimerson:
@analogic , do you have an opinion on the switch to this "flow through" style of logic for `should_skip()`?   Specifically, does it make it easier to reason about what will happen? Is it easier to maintain (alter/extend)? 

```js
exports.should_check = function (connection) {
     const plugin = this;

     let result = true;  // default

     if (plugin.cfg.check.authenticated == false && connection.notes.auth_user) {
         connection.transaction.results.add(plugin, { skip: 'authed'});
         result = false;
     }

     if (plugin.cfg.check.relay == false && connection.relaying) {
         connection.transaction.results.add(plugin, { skip: 'relay'});
         result = false;
     }

     if (plugin.cfg.check.local_ip == false && connection.remote.is_local) {
         connection.transaction.results.add(plugin, { skip: 'local_ip'});
         result = false;
     }

     if (plugin.cfg.check.private_ip == false && connection.remote.is_private) {
         if (plugin.cfg.check.local_ip == true && connection.remote.is_local) {
             // local IPs are included in private IPs
         }
         else {
             connection.transaction.results.add(plugin, { skip: 'private_ip'});
             result = false;
         }
     }

     return result;
 }
```

analogic:
LGTM!

To plugin.js abstraction - In future It would be nice to have some generic flag which would control skipping from outside. Like another plugin deciding by IP/RCPT/FROM/AUTH/user profile. (I like what https://github.com/creamfinance/haraka-plugin-plugin_manager can do but I don't like mixing yaml and js logic)

msimerson:
> In future It would be nice to have some generic flag which would control skipping from outside

Exactly. In briefly thinking about it, what I see is that there's already a half dozen implementations of varying degrees of skipping / whitelisting. Before assembling this PR I reviewed them but none of them are a one sized fits all. The conditions under which you might want to whitelist a connection from early_talker or greylisting are considerably different than for rspamd or clamd. But this should_check logic does cover the majority of cases for most plugins. In another evolution or two maybe we'll have something that's general purpose enough to work everywhere.

PR #2573, opened 2018-12-23
server: warn when nodes=undefined

### LINKED ISSUE #2388: Do a logwarn() when smtp.ini nodes= is unset
Most people miss this setting, and it's important for keeping the server up (and for scalability). We should issue a logwarn() on startup if it's unset.

Not that anyone reads the logs though...

msimerson:
Fixes #2388 

- server: warn when nodes=undefined
- default to nodes=1

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -9,6 +9,7 @@
 * early_talker: extend reasons to skip checking #2564
 * tls: add 'ca' option (for CA root file) #2571
 * outbound: little cleanups #2572
+* server: default to nodes=1 (was undefined) #2573
 
 ### New Features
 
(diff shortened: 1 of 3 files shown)


analogic:
So cluster is now on by default? I like to handle fatalities at supervisor (for alerts at least) so I hope that single thread process will not vanish from code eventually :)

msimerson:
> So cluster is now on by default?

Yes, that would be the result of this PR. This PR doesn’t change what you’re able to do, it just changes the default. You can always disable cluster by setting nodes=0 in smtp.ini.

PR #2577, opened 2018-12-26
Introduce status plugin

analogic:
Plugin which makes queues and pools accessible for inspection through SMTP commands. Please see the code and https://github.com/analogic/Haraka/blob/3f11ebd3a95716ce266f9d011f5e0713c3f9aef1/docs/plugins/status.md

- no changes in current code

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/config/plugins
+++ b/config/plugins
@@ -7,6 +7,7 @@
 #
 # To see the help docs for a particular plugin, run 'haraka -h plugin.name'
 
+status
 #process_title
 # Log to syslog (see 'haraka -h syslog')
 # syslog
--- a/config/plugins
+++ b/config/plugins
@@ -7,7 +7,7 @@
 #
 # To see the help docs for a particular plugin, run 'haraka -h plugin.name'
 
-status
+#status
 #process_title
 # Log to syslog (see 'haraka -h syslog')
 # syslog
--- a/plugins/status.js
+++ b/plugins/status.js
@@ -1,0 +1,245 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const async = require('async');
+
+exports.register = function () {
+    this.outbound = require('./../outbound');
+    this.queue_dir = require('./../outbound/queue').queue_dir;
+}
+
+exports.hook_capabilities = function (next, connection) {
+
+    if (connection.remote.ip === '127.0.0.1' || connection.remote.ip === '::1') {
+        connection.capabilities.push('STATUS');
+    }
+
+    next();
+}
+
+exports.hook_unrecognized_command = function (next, connection, params) {
+    const self = this;
+
+    const result_process = function (err, result) {
+        if (err) return next(DENY, err.message);
+        return connection.respond(211, result ? JSON.stringify(result) : "null", () => next(OK));
+    };
+
+    if (params[0] !== 'STATUS') {
+        next();
+    }
+    else if (connection.remote.ip !== '127.0.0.1' && connection.remote.ip !== '::1') {
+        return next(DENY, "STATUS not allowed remotely");
+    }
+    else {
+        if (server.cluster && !params[1].match(/^QUEUE LIST/)) {
+            self.call_master(params[1], result_process);
+        }
+        else {
+            self.command_action(params[1], result_process);
+        }
+    }
+}
+
+exports.command_action = function (cmd, cb) {
+    const self = this;
+    const params = cmd.split(' ');
+
+    switch (params.shift()) {
+        case 'POOL':
+            return self.pool_action(params, cb);
+        case 'QUEUE':
+            return self.queue_action(params, cb);
+        default:
+            cb("unknown STATUS command")
+    }
+}
+
+exports.pool_action = function (params, cb) {
+    const self = this;
+
+    switch (params.shift()) {
+        case 'LIST':
+            return self.pool_list(cb);
+        default:
+            cb("unknown POOL command")
+    }
+}
+
+exports.queue_action = function (params, cb) {
+    const self = this;
+
+    switch (params.shift()) {
+        case 'LIST':
+            return self.queue_list(cb);
+        case 'STATS':
+            return self.queue_stats(cb);
+        case 'INSPECT':
+            return self.queue_inspect(cb);
+        case 'DISCARD':
+            return self.queue_discard(params.shift(), cb);
+        case 'PUSH':
+            return self.queue_push(params.shift(), cb);
+        default:
+            cb("unknown QUEUE command")
+    }
+}
+
+exports.pool_list = function (cb) {
+    const result = {};
+
+    if (server.notes.pool) {
+        Object.keys(server.notes.pool).forEach(function (name) {
+            const instance = server.notes.pool[name];
+
+            result[name] = {inUse: instance.inUseObjectsCount(), size: instance.getPoolSize()};
+        });
+    }
+
+    cb(null, result);
+}
+
+exports.queue_list = function (cb) {
+    this.outbound.list_queue(function (err, qlist) {
+        if (err) cb(err);
+
+        const result = [];
+
+        if (qlist) {
+            qlist.forEach(function (todo) {
+                result.push({
+                    file: todo.file,
+                    uuid: todo.uuid,
+                    queue_time: todo.queue_time,
+                    domain: todo.domain,
+                    from: todo.mail_from.toString(),
+                    to: todo.rcpt_to.map((r) => r.toString())
+                });
+            });
+        }
+
+        cb(err, result);
+    });
+}
+
+exports.queue_stats = function (cb) {
+    cb(null, this.outbound.get_stats());
+}
+
+exports.queue_inspect = function (cb) {
+    cb(null, {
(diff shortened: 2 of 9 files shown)


plugins/status.js
```diff
@@ -0,0 +1,245 @@
...
+const async = require('async');
+
+exports.register = function () {
+    this.outbound = require('./../outbound');
+    this.queue_dir = require('./../outbound/queue').queue_dir;
+}
+
+exports.hook_capabilities = function (next, connection) {
+
+    if (connection.remote.ip === '127.0.0.1' || connection.remote.ip === '::1') {
```
msimerson: Any reason not to use the already computed `is_local` flag?

```suggestion
    if (connect.remote.is_local) {
```

plugins/status.js
```diff
@@ -0,0 +1,245 @@
...
+        connection.capabilities.push('STATUS');
+    }
+
+    next();
+}
+
+exports.hook_unrecognized_command = function (next, connection, params) {
+    const self = this;
+
+    const result_process = function (err, result) {
```
msimerson: Now that we're in an es6 world with `const` and `let` not doing variable hoisting, we've deprecated variable expressions in favor of variable declarations (which are still hoisted).

```suggestion
    function result_process (err, result) {
```

plugins/status.js
```diff
@@ -0,0 +1,245 @@
...
+        if (err) return next(DENY, err.message);
+        return connection.respond(211, result ? JSON.stringify(result) : "null", () => next(OK));
+    };
+
```
msimerson: This might be a dumb question. You've got 3 `if` conditions here and only two of them call the `next` callback. That doesn't "look" right. Is it?

plugins/status.js
```diff
@@ -0,0 +1,245 @@
...
+    else {
+        if (server.cluster && !params[1].match(/^QUEUE LIST/)) {
+            self.call_master(params[1], result_process);
+        }
+        else {
+            self.command_action(params[1], result_process);
+        }
+    }
+}
+
```
msimerson: Just an observation (no need to change): in the next 3 functions, you've done `const self = this`, but in none of the cases is it necessary. You can just use `this` within the switch blocks and the correct `this` will be resolved. Generally we only create references to `this` when a) an enclosing callback alters `this` or b) when providing a more obvious name to `this` makes it easier to humans to parse/grok what `this` is. Ex: within plugins we often do: `const plugin = this`.

plugins/status.js
```diff
@@ -0,0 +1,245 @@
...
+    const result_process = function (err, result) {
+        if (err) return next(DENY, err.message);
+        return connection.respond(211, result ? JSON.stringify(result) : "null", () => next(OK));
+    };
+
+    if (params[0] !== 'STATUS') {
+        next();
+    }
+    else if (connection.remote.ip !== '127.0.0.1' && connection.remote.ip !== '::1') {
```
msimerson:
```suggestion
    else if (connection.remote.is_local) {
```

plugins/status.js
msimerson: FYI: the `is_local` property also takes into account the link-local (IPv4 and 6) and IPv6 unique local assignments, so it's a fair bit more robust.

msimerson:
Very nice work @analogic , this looks very very good.

plugins/status.js
analogic: Yep, comming from "different" languages I am trying to use this right. Well, these are just leftover from debuging, updated...

plugins/status.js
analogic: make sense

plugins/status.js
```diff
@@ -0,0 +1,241 @@
...
+        return next(DENY, "STATUS not allowed remotely");
+    }
+    else {
+        return this.run(params[1], function (err, result) {
+
+            if (err) return next(DENY, err.message);
+            return connection.respond(211, result ? JSON.stringify(result) : "null", () => next(OK));
+        })
+    }
+}
```
msimerson: Within this function you have inconsistent use of `return`. Return stops the execution of a function and returns a value. We generally use `return` for two purposes: flow control and/or returning something (a value, a promise, etc.) to the calling function.

Since nothing is checking the return values, it's obvious that you're using `return` for flow control. But you're also using the `if` blocks for flow control. Your first `if` condition doesn't have a `return`, but the 2nd and 3rd do. You should choose one or the other for flow control. The result would look like one of:

## return for flow control

```js
exports.hook_unrecognized_command = function (next, connection, params) {
    if (params[0] !== 'STATUS')      return next();
    if (!connection.remote.is_local) return next(DENY, "STATUS not allowed remotely");

    this.run(params[1], function (err, result) {

        if (err) return next(DENY, err.message);
        connection.respond(211, result ? JSON.stringify(result) : "null", () => next(OK));
    })
}
```

## conditional blocks for flow control

```js
exports.hook_unrecognized_command = function (next, connection, params) {
    if (params[0] !== 'STATUS') {
        next();
    }
    else if (!connection.remote.is_local) {
        next(DENY, "STATUS not allowed remotely");
    }
    else {
        this.run(params[1], function (err, result) {

            if (err) return next(DENY, err.message);
            connection.respond(211, result ? JSON.stringify(result) : "null", () => next(OK));
        })
    }
}
```

Choose your favorite. I prefer the former because I find it easier to read, nicer to extend (over time, the stuff inside the if test(s) tends to grow, making them hairy), and since JS often forces indents on us, I like to avoid them when I can.

plugins/status.js
```diff
@@ -0,0 +1,241 @@
...
+
+exports.run = function (cmd, cb) {
+    if (server.cluster && !cmd.match(/^QUEUE LIST/)) {
```
msimerson:
```suggestion
    if (server.cluster && !/^QUEUE LIST/.test(cmd)) {
```

Match works here as you expect, but it creates an array and returns it with your match results. If you don't need the match results, `test` will return a boolean instead, which is more efficient.

tests/plugins/status.js
```diff
@@ -0,0 +1,138 @@
+'use strict';
+
+const fixtures = require('haraka-test-fixtures');
+const outbound = require('./../../outbound');
```
msimerson:
```suggestion
const outbound = require('../../outbound');
```

tests/plugins/status.js
```diff
@@ -0,0 +1,138 @@
...
+        test.done();
+    },
+}
+
+exports.access = {
+    setUp : _set_up,
+    'remote': function (test) {
+
+        test.expect(1);
+        const cb = function (code) {
```
msimerson:
```suggestion
        function cb (code) {
```

tests/plugins/status.js
```diff
@@ -0,0 +1,138 @@
...
+        test.expect(1);
+
+        outbound.temp_fail_queue = new TimerQueue(10);
+        outbound.temp_fail_queue.add("file1", 10, function () {
+            test.ok(false, "This callback should not be called");
+            test.done();
+        });
+        outbound.temp_fail_queue.add("file2", 2000, function () {});
+
+        const res = function () {
```
msimerson:
```suggestion
        function res () {
```

plugins/status.js
analogic: Yep, these two ifs are just validation before actual work. I don't know why I mixed this, I like former way too...

plugins/status.js
```diff
@@ -0,0 +1,244 @@
...
+}
+
+exports.hook_unrecognized_command = function (next, connection, params) {
+    if (params[0] !== 'STATUS') return next();
+    if (!connection.remote.is_local) return next(DENY, "STATUS not allowed remotely");
+
+    this.run(params[1], function (err, result) {
+        if (err) return next(DENY, err.message);
+
+        connection.respond(211, result ? JSON.stringify(result) : "null", () => next(OK));
```
msimerson: You're returning the string "null" here. Typically one would pass a js `null` (no quotes), or an empty string `''` (empty quotes). This string won't evaluate to false like an actual `null` would.

config/plugins
```diff
@@ -7,6 +7,7 @@
 #
 # To see the help docs for a particular plugin, run 'haraka -h plugin.name'
 
+status
```
msimerson:
```suggestion
# status
```

Open question: should this plugin be enabled by default? We purposefully keep the enabled-by-default plugins to a minimum. I lean towards not enabled by default.

I like and expect to use this plugin, and there's a good chance that someone extends it with web hooks. But that's not a good enough reason to have it enabled by default. Shucks, we don't even have TLS enabled by default.

plugins/status.js
analogic: It will come out as "211 null" message and thus json parser on other end will convert to `null` also. Due some glitch sending empty string response with `connection.respond(211, '')` is not working and nothing happens.

config/plugins
analogic: Definitely no need to enable this by default, this is reason why this is done as plugin (vs #2428). Apache's mod_status or nginx status page is not enabled by default also

I am going to implement web hooks too eventually. Also extending plugin (and haraka) with hooks like command_help and command_exec to just move "--qlist, --qstat, --qunstick..." commands inside plugin and remove logic from core would be nice.

PR #2581, opened 2019-01-08
Status plugin: use queued item's callback instead of setting fire_time to -1

analogic:
Fixes problem where pushed emails are not actually processed due waiting items earlier at queue, see https://github.com/haraka/Haraka/blob/e9128d6dea02f3509abcb5104f796e3a9a73313e/outbound/timer_queue.js#L60-L69

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/plugins/status.js
+++ b/plugins/status.js
@@ -153,7 +153,7 @@ exports.queue_push = function (file, cb) {
     for (let i = 0; i < queue.length; i++) {
         if (queue[i].id !== file) continue;
 
-        let item = queue.splice(i, 1)[0];
+        const item = queue.splice(i, 1)[0];
         item.cb();
         
         break;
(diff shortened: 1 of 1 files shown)


msimerson:
There are lint errors that need attention.

analogic:
Travis working and now appveyor is failing. × SPF - mod_redirect, false, AssertionError: 3 == 6

It seems that it is same issue that I've hit recently with ubuntu's systemd-resolved https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1804487 - resolver didn't return all txt records due length

msimerson:
I clicked the "Rebuilt PR" button in AppVeyor and now it passed. Good enough.

PR #2582, opened 2019-01-09
correct MIME parsing when charset: utf8 and encoding: 8bit

### LINKED ISSUE #2548: Problem non-latin symbols
### system info

Haraka | Haraka.js — Version: 2.8.21
 --- | :--- 
Node | v7.10.1
OS | Linux hostname 4.4.0-83-generic #106-Ubuntu SMP Mon Jun 26 17:54:43 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
openssl | OpenSSL 1.0.2g  1 Mar 2016

### Expected behavior
Content Transfer Encoding 8bit lose non-ASCII symbols

### Observed behavior
Emails from many of our clients containing Content-Type: utf-8 and Content-Transfer-Encoding 8bit removes non-ASCII characters. For example, German ß turns into �.

### Steps to reproduce
Create a plugin from this code

```javascript
var plugin;

    plugin = this;
};

exports.hook_rcpt_ok = function (next, connection, params) {
    connection.transaction.parse_body = true;
    return next();
};

exports.hook_data_post = function (next, connection, params) {
    connection.logwarn(connection.transaction.body);
    return next();
};
```

and send this email

```
Subject: Test
To: to@example.com
From: from@example.com
Date: Wed, 27 Nov 2018 18:09:23 +0100
MIME-version: 1.0
Content-type: multipart/alternative;
 boundary=------------D0A00162984CC178E2583417
Content-language: en-US

This is a multi-part message in MIME format.
--------------D0A00162984CC178E2583417
Content-Type: text/plain; charset=utf-8; format=flowed
Content-Transfer-Encoding: 8bit

Grüße,

Buß

--------------D0A00162984CC178E2583417
Content-Type: text/html; charset=utf-8
Content-Transfer-Encoding: 8bit

<html>
  <head>
    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
  </head>
  <body text="#000000" bgcolor="#FFFFFF">
    <p>Grüße,

Buß<br>
  </body>
</html>

--------------D0A00162984CC178E2583417--
```
via swaks with command
```swaks --to to@example.com --from from@example.com -d test.eml```

As a result we see the following problem
```
       bodytext: 'Gr��e,\n\n Bu�\n\n',
       body_text_encoded: 'Grüße,\n\nBuß\n\n',
       body_encoding: 'utf-8',
```

It is not a problem with console. This will be displayed in DB and other places.

We were trying to find out what the problem was and found 

> 367 to 369 in [master](/haraka/Haraka/blob/master/mailbody.js)

`new Buffer(line, 'binary');` will lose all non ascii symbols, when 'binary' is removed everything is ok.

analogic:
Fixes #2548

Originaly proposed solution from @Zerstoren would damage all others encodings other than utf8. There might be better place to handle this - maybe at `decode_8bit` itself. I've also added tests so it should be safe to improve this decoding logic.

- for 8bit utf-8  we use `buffer.toString('binary')` insteand of `buffer.toString('utf8')`

Problem might be another multibyte charsets (utf-16, utf-32) also, but I don't think they are seriously used anyway

Checklist:
- [ ] docs updated
- [x] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -1,5 +1,5 @@
 
-## 2.8.24 - Mmm DD, 201Y
+## 2.8.24 - Jan DD, 2019
 
 ### Changes
 
@@ -11,6 +11,7 @@
 * outbound: little cleanups #2572
 * smtp_client: pass pool_timeout to new SMTPClient #2574
 * server: default to nodes=1 (was undefined) #2573
+* test/server: use IPv4 127.0.0.1 instead of localhost #2584
 
 ### New Features
 
--- a/tests/server.js
+++ b/tests/server.js
@@ -250,7 +250,7 @@ exports.smtp_client = {
 
 exports.nodemailer = {
     setUp : function (done) {
-        _setupServer(this, 'localhost:2503', done);
+        _setupServer(this, '127.0.0.1:2503', done);
     },
     tearDown: _tearDownServer,
     'accepts SMTP message': function (test) {
@@ -258,7 +258,7 @@ exports.nodemailer = {
         test.expect(1);
         const nodemailer = require('nodemailer');
         const transporter = nodemailer.createTransport({
-            host: 'localhost',
+            host: '127.0.0.1',
             port: 2503,
             tls: {
                 // do not fail on invalid certs
@@ -292,7 +292,7 @@ exports.nodemailer = {
         test.expect(1);
         const nodemailer = require('nodemailer');
         const transporter = nodemailer.createTransport({
-            host: 'localhost',
+            host: '127.0.0.1',
             port: 2503,
             auth: {
                 user: 'matt',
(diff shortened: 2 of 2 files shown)


baudehlo:
Tests are failing. I can't figure out why. Can you?

analogic:
It seems like travis only problem. I have no problem localy

baudehlo:
OK I'll hit restart a few times and see if it goes through. The branch
travis passed, it's only the PR one failing.

msimerson:
It's a "test" problem that I haven't had time to fix yet. 

````
✖ nodemailer - rejects invalid auth
AssertionError: 'ECONNECTION' == 'EAUTH'
````

It's on my list of things to get to.

PR #2583, opened 2019-01-09
Fixes related to Check flags

gramakri:
- Update SpamAssassin plugin docs to have check flags
- check flags are defaulted to true

Checklist:
- [X] docs updated

### DIFF
--- a/docs/plugins/clamd.md
+++ b/docs/plugins/clamd.md
@@ -113,6 +113,12 @@ meeting following criteria.
 
     If true, messages from localhost will be scanned.
 
+- relay
+
+    Default: true
+
+    If true, messages that are to be relayed will be scanned.
+
 ## clamd.excludes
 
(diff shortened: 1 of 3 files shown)


gramakri:
Looks like the CI failures are unrelated. `nodemailer - rejects invalid auth`

msimerson:
Correct, CI changes are due to #2578 which I haven't had time to fix. It's a small issue, the error message has changed and so the test for it needs to be adapted.

This PR changes the defaults for all the `check` settings. While that's personally favorable (that's how I configure it), plenty of Haraka users opt to not scan mail that originates locally (auth, relay, etc.) and for good reason(s). We generally avoid making changes to default behavior unless we're certain that the new default is an improvement for everybody.

msimerson:
Ahh, I recall now. Thanks for the reminder.

PR #2587, opened 2019-01-10
spf: use the skip config for helo/ehlo checks

gramakri:
- This extends the use of the skip flags in the ehlo hook


gramakri:
Just as a node, it's unlikely that auth is true before ehlo but the acl plugin might allow relaying.

PR #2590, opened 2019-01-11
spf: add option to disable mfrom and/or ehlo checks

gramakri:
SPF checks on the EHLO string is not really required (http://www.openspf.org/FAQ/Envelope_from_scope). This option allow us to turn it off.

Checklist:
- [X] docs updated
- [X] tests updated

### DIFF
--- a/docs/plugins/spf.md
+++ b/docs/plugins/spf.md
@@ -27,6 +27,15 @@ In that use case, Haraka should use context=myself.
     * context=sender    evaluate SPF based on the sender (connection.remote.ip)
     * context=myself    evaluate SPF based on Haraka's public IP
 
+By default, the plugin does SPF checks for the domain in EHLO/HELO string as
+well as the Envelope From. This can be selectively disabled:
+
+    [check]
+    ; SPF check the domain in helo string
+    helo=true
+    ; SPF check the domain in MAIL FROM
+    mfrom=true
+
 The rest of the optional settings (disabled by default) permit deferring or
 denying mail from senders whose SPF fails the checks.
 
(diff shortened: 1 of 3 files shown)


gramakri:
We found that we saw 3 `Received-SPF` headers in our emails. The reason was because the SPF plugin was running 3 times - 2 EHLO (one before and another after TLS) and then once for MAIL FROM. This patch helped us disable the EHLO time checks.

msimerson:
This doesn't feel right to me. 

Not checking both HELO/EHLO and MFROM in the evaluation isn't actually performing a valid SPF test. I need to look into that deeper, or get @smfreegard's blessing before I'll approve this. Something that *would* make sense to me though, is skipping SPF evaluation of EHLO the 2nd time, if the EHLO hostname is unchanged. Something like this should do the trick:

```js
@@ -107,6 +110,10 @@ exports.hook_helo = exports.hook_ehlo = function (next, connection, helo) {
         return next();
     }, plugin.cfg.lookup_timeout * 1000);
 
+    // avoid 2nd EHLO evaluation if EHLO host is identical
+    const results = connection.results.get(plugin);
+    if (results && results.domain === helo) return next();
+
     spf.check_host(connection.remote.ip, helo, null, function (err, result) {
         if (timer) clearTimeout(timer);
         if (timeout) return;
```

gramakri:
@msimerson Thanks for reviewing.

I checked http://www.openspf.org/RFC_4408. In 2.1, it says `It is RECOMMENDED that SPF clients not only check the "MAIL FROM" identity, but also separately check the "HELO" identity by applying the check_host() function (Section 4) to the "HELO" identity as the <sender>."`

Mmmm, so the RFC actually recommends it. I submitted this PR because I thought it was incorrect to check in the first place. I think I like your idea of just making sure that we don't check the EHLO identity twice. Let me rework this change.

I added @smfreegard  to review/comment in any case.

gramakri:
I will create a new PR with the fix suggested by @msimerson

PR #2609, opened 2019-03-12
transaction: dont replace CRLF with newlines

### LINKED ISSUE #2594: Some emails get invalid DKIM body hash
When examining DMARC reports I've found that small percentage of emails fails with DKIM. Later I've found that the problem is that "body hash did not verify" and also it seems that it happens almost only when email is "generated" and attachment included.

Steps to reproduce:
1. enable dkim_sign plugin & generate keys
2. install and enable rspamd
3. turn DMARC reporting on with sending through haraka
4. send some emails (from DMARC enabled domains) back and forth
...wait aprox 1 day...
5. rspamd will send out DMARC reports with invalid DKIM body hash

I am sure that there is more simple method but this is the only way I've identified yet. (I will post examples later if I don't get this bug on my own)

### LINKED ISSUE #2595: inactivity_timeout is 600, not 300
### system info

Haraka | Haraka.js — Version: 2.8.23
 --- | :---
Node | v8.10.0
OS | Linux testmaschine01 4.15.0-43-generic #46-Ubuntu SMP Thu Dec 6 14:45:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
openssl | OpenSSL 1.1.0g  2 Nov 2017

### Expected behavior

Documentation says:
https://haraka.github.io/manual/CoreConfig.html
`inactivity_time - how long to let clients idle in seconds (default: 300)`

smtp.ini comment says:
```
; Time in seconds to let sockets be idle with no activity
;inactivity_timeout=300
```

### Observed behavior

When running Haraka, the timeout is 600 seconds (10 minutes):
```
Jan 25 20:55:13 testmaschine01 haraka[915]: [NOTICE] [39BB1D39-138D-45D5-95DA-E2EB638460AD] [core] connect ip=1.2.3.4 port=40920 local_ip=:: local_port=2525
Jan 25 21:05:13 testmaschine01 haraka[915]: [WARN] [39BB1D39-138D-45D5-95DA-E2EB638460AD] [core] client [1.2.3.4] connection timed out
```

### ToDo:

`inactivity_time` on documentation page (https://haraka.github.io/manual/CoreConfig.html) is wrong, it's called `inactivity_timeout` in smtp.ini and in the code.

Searching for `inactivity_timeout` in this repository on Github is set to 600 in one place, and 300 in another place:
grafik

msimerson:
patch from #2594

Fixes #2594 

- transaction: dont replace CRLF with newlines because it breaks DKIM signatures

Checklist:
- [ ] docs updated
- [ ] tests updated


msimerson:
I've tested this patch now in production. Using the crlf-bug.txt from #2594, I can replicate the DKIM signature breaking. With this patch applied, I can still send the message through my server (using `swaks -d crlf-bug.txt`) and it gets delivered to the next hop (using Gmail) but then it doesn't show up in my gmail box. Yet.

PR #2610, opened 2019-03-12
tls: load CA certs from file when specified

msimerson:
This is useful on distros that get slow updates and have a package available that provides a (sometimes MUCH) newer CA list.


msimerson:
apparently I already got this merged. (after updating the branch, there's nothing left to commit)

PR #2643, opened 2019-05-14
Actually enforce using key for INTERNALCMD

d--j:
Right now you can issue `INTERNALCMD` SMTP commands without a key – even when the `internalcmd_key` file exists. This renders having an `internalcmd_key` file moot.

This pull requests rejects `INTERNALCMD` commands without a key when a key is configured.


baudehlo:
God it was always my intention to auto create this file on install. Shame I never got around to it.

PR #2647, opened 2019-05-21
mailheader: do not remove partial matches from header_list

### LINKED ISSUE #2641: transaction.remove_header removes too much
The following test plugin should remove the header `X-Test` but do not affect any other headers.

```js
    const plugin = this;
    plugin.inherits('queue/quarantine')
    plugin.load_quarantine_ini();
    plugin.register_hook('queue',          'quarantine');
    plugin.register_hook('queue_outbound', 'quarantine');
}

exports.hook_rcpt = function (next, connection) {
    const plugin = this
    connection.notes.quarantine = true
    next(OK)
}

exports.hook_data_post = function (next, connection) {
    const plugin = this
    connection.transaction.remove_header('X-Test')
    next()
}

```

But when I call
```shell
swaks -s $HARAKA_HOST -p $HARAKA_PORT -t test@example.com --add-header "X-Test: remove-me" --add-header "X-Test-1: do-not-remove-me"
```

The `X-Test-1` header gets spuriously removed, too.

Looks like `_remove_more` in `mailheader.js` is the culprit:
https://github.com/haraka/Haraka/blob/4aea44d56834a1a4be78c56b646a6c266aa7d7b5/mailheader.js#L127-L135
(only comparing the header key where it should compare key + ':')

msimerson:
Fixes #2641

- mailheader: add test for remove()
- mailheader: do not remove partial matches from header_list

Checklist:
- [ ] docs updated
- [x] tests updated

### DIFF
--- a/mailheader.js
+++ b/mailheader.js
@@ -126,7 +126,7 @@ class Header {
     _remove_more (key) {
         const key_len = key.length;
         for (let i=0, l=this.header_list.length; i < l; i++) {
-            if (this.header_list[i].substring(0, key_len).toLowerCase() === key) {
+            if (this.header_list[i].substring(0, key_len + 1).toLowerCase() === `${key}:`) {
                 this.header_list.splice(i, 1);
                 return this._remove_more(key);
             }
@@ -164,7 +164,8 @@ class Header {
         if (/[^\x00-\x7f]/.test(value)) {
             // Need to QP encode this header value and assume UTF-8
             value = '=?UTF-8?q?' + utils.encode_qp(value) + '?=';
-            value = value.replace(/=\n/g, ''); // remove wraps - headers can only wrap at whitespace (with continuations)
+            // remove wraps - headers can only wrap at whitespace (with continuations)
+            value = value.replace(/=\n/g, '');
         }
         this._add_header(key.toLowerCase(), value, "push");
         this._add_header_decode(key.toLowerCase(), value, "push");
--- a/mailheader.js
+++ b/mailheader.js
@@ -21,8 +21,7 @@ class Header {
     parse (lines) {
         const self = this;
 
-        for (let i=0,l=lines.length; i < l; i++) {
-            const line = lines[i];
+        for (const line of lines) {
             if (/^[ \t]/.test(line)) {
                 // continuation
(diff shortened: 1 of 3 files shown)


mailheader.js
```diff
@@ -126,8 +125,8 @@ class Header {
 
     _remove_more (key) {
         const key_len = key.length;
-        for (let i=0,l=this.header_list.length; i < l; i++) {
-            if (this.header_list[i].substring(0, key_len).toLowerCase() === key) {
+        for (let i=0, l=this.header_list.length; i < l; i++) {
```
KingNoosh: Can't we do `const header of this.header_list` here too?

mailheader.js
```diff
@@ -126,8 +125,8 @@ class Header {
...
+            if (this.header_list[i].substring(0, key_len + 1).toLowerCase() === `${key}:`) {
```
KingNoosh: Remember to update this to `header`.

tests/mailheader.js
```diff
@@ -26,62 +26,75 @@ for (let i=0; i<lines.length; i++) {
...
-            h.get_decoded('content-type'),
+            this.h.get_decoded('content-type'),
             'multipart/alternative;   boundary=Apple-Mail-F2C5DAD3-7EB3-409D-9FE0-135C9FD43B69'
         );
-        test.equal(h.get_decoded('fromUTF8'), 'Kohl’s <Kohls@s.kohls.com>');
+        test.equal(this.h.get_decoded('fromUTF8'), 'Kohl’s <Kohls@s.kohls.com>');
         test.done();
     },
-    'content type w/parens': test => {
+    'content type w/parens': function (test) {
```
KingNoosh: What benefit does this change have?

msimerson: It allows `this` within the setup function to also refer to `this` within the test.

KingNoosh: Ah scoping, cool no problem.

mailheader.js
msimerson: If you notice, `i` gets used within the loop (in the splice). Kinda important...

KingNoosh: I should probably look into getting new glasses.

PR #2699, opened 2019-09-14
Fix mx hook in smtp_forward

gramakri:
-  Register the mx_hook in smtp_forward plugin
- Get the domain config based on the `domain_selector` setting

### DIFF
--- a/plugins/queue/smtp_forward.js
+++ b/plugins/queue/smtp_forward.js
@@ -341,9 +341,10 @@ exports.get_mx = function (next, hmail, domain) {
         return next(OK, plugin.get_mx_next_hop(hmail.todo.notes.next_hop));
     }
 
-    if (domain !== domain.toLowerCase()) domain = domain.toLowerCase();
+    const dom = plugin.cfg.main.domain_selector === 'mail_from' ? hmail.todo.mail_from.host.toLowerCase() : domain.toLowerCase();
+    const cfg = plugin.cfg[dom];
 
-    if (plugin.cfg[domain] === undefined) {
+    if (cfg === undefined) {
         plugin.logdebug('using DNS MX for: ' + domain);
         return next();
     }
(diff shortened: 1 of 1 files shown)


gramakri:
It seems the get_mx function is unused in smtp_forward since it is never registered.

With this change, any mails queued in haraka will be sent via the smtp_forward settings (as intended by the original code).

plugins/queue/smtp_forward.js
```diff
@@ -339,9 +341,10 @@ exports.get_mx = function (next, hmail, domain) {
         return next(OK, plugin.get_mx_next_hop(hmail.todo.notes.next_hop));
     }
 
-    if (domain !== domain.toLowerCase()) domain = domain.toLowerCase();
+    const dom = plugin.cfg.main.domain_selector === 'mail_from' ? hmail.todo.mail_from.host.toLowerCase() : domain.toLowerCase();
```
msimerson:
```suggestion
    let dom = domain.toLowerCase();
    if (plugin.cfg.main.domain_selector && plugin.cfg.main.domain_selector === 'mail_from') hmail.todo.mail_from.host.toLowerCase();
```

Should we check that `domain_selector` is defined first?

baudehlo: FWIW that suggestion is broken.

msimerson: Not that it matters, but I've fixed it.

lobovkin:
Now smtp_forward use DNS MX for unlisted domains. Shouldn't it use the main host from smtp_forward.ini?

msimerson:
If you want unlisted domains to get smtp forwarded, set `enable_outbound=false` in the main section. Then the message gets delivered by smtp_forward and the get_mx hook won't ever get called by outbound. I think.

lobovkin:
> If you want unlisted domains to get smtp forwarded, set `enable_outbound=false` in the main section. Then the message gets delivered by smtp_forward and the get_mx hook won't ever get called by outbound. I think.

get_mx hook is still used for queued messages from disk (if the initial delivery fails).

PR #2706, opened 2019-09-30
mailheader: avoid variable assignments, es6 interp. strings

msimerson:
- mailheader: avoid two variable assignments
- mailheader: convert a few + concats with interpolated es6 strings
- disable github actions/workflows testing on windows node 8 & 10 (currently broken)

### DIFF
--- a/mailheader.js
+++ b/mailheader.js
@@ -86,12 +86,11 @@ class Header {
             return val;
         }
 
-        //strip whitespace between encoded-words, rfc 2047 6.2
-        val = val.replace(/(=\?.+?\?=)\s+(?==\?.+?\?=)/g,"$1");
-
-        val = val.replace(/=\?([\w_-]+)(\*[\w_-]+)?\?([bqBQ])\?([\s\S]*?)\?=/g, _decode_header);
-
-        return val;
+        return val
+            // strip whitespace between encoded-words, rfc 2047 6.2
+            .replace(/(=\?.+?\?=)\s+(?==\?.+?\?=)/g,"$1")
+            // decode each encoded match
+            .replace(/=\?([\w_-]+)(\*[\w_-]+)?\?([bqBQ])\?([\s\S]*?)\?=/g, _decode_header);
     }
 
     get (key) {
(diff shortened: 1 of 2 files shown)


mailheader.js
```diff
@@ -86,12 +86,11 @@ class Header {
...
-        val = val.replace(/(=\?.+?\?=)\s+(?==\?.+?\?=)/g,"$1");
-
-        val = val.replace(/=\?([\w_-]+)(\*[\w_-]+)?\?([bqBQ])\?([\s\S]*?)\?=/g, _decode_header);
-
-        return val;
+        return val
+            // strip whitespace between encoded-words, rfc 2047 6.2
+            .replace(/(=\?.+?\?=)\s+(?==\?.+?\?=)/g,"$1")
+            // decode each encoded match
+            .replace(/=\?([\w_-]+)(\*[\w_-]+)?\?([bqBQ])\?([\s\S]*?)\?=/g, _decode_header);
```
shurli: .replace(/=\?([\w_-]+)(\*[\w_-]+)?\?([bqBQ])\?([\S]*?)\?=/g, _decode_header);

what about the \s in the regex?

rfc 2047 2: An 'encoded-word' is defined by the following ABNF grammar.  The
   notation of RFC 822 is used, with the exception that white space
   characters MUST NOT appear between components of an 'encoded-word'.

msimerson: Shucks, I don't know. This PR doesn't change the regexes at all, I just optimized away the variable assignments.

It appears that we _could_ remove it, since there shouldn't ever be any whitespace after the first regex, but since it won't ever match, its presence won't hurt.

shurli: The first regex strips whitespaces between encoded-words.

Whereelse the `\s` matches whitespaces within encoded-words,
where they must not appear since encoded-words should be atoms.
Theoretically encoded-words with inner whitespaces  aren't encoded-words at all and should not be decoded.

Of course this could also be discussed and possibly changed in another PR

msimerson: > (.*?) in 2016 but was changed to ([\s\S]*?), isn't it the same?

It's *almost* the same:

- \s matches `[ \t\r\n\v\f]`
- \S matches `[^ \t\r\n\v\f]`
- . matches any char **except** newlines. 

So .* will match everything except a newline, and [\s\S] will include newlines (I'm guessing in a folded header).

shurli: ok.
My opinion is that line folding within an encoded-word is not rfc conform and mulitiple encoded words is the way to do.

But as not every mailsender is rfc conform, its probably the best was to handle this lazy and accept also folding within encoded words.

for this reason, according to your last post we should change this in the first regex also.
from 
`.replace(/(=\?.+?\?=)\s+(?==\?.+?\?=)/g,"$1")`
to 
`.replace(/(=\?[\s\S]+?\?=)\s+(?==\?[\s\S]+?\?=)/g,"$1")`

PR #2710, opened 2019-10-06
Replace concatenated strings with template literals

lnedry:
Fixes # Replaced concatenated strings with template literals


lnedry:
I made an error so am closing.

PR #2722, opened 2019-10-13
es6 templates: replace concatenated strings

msimerson:
- replace many string concatenations with es6 templates

We've been chipping away at this this for a couple years. It's starting to feel like paper cuts, so I went looking and found an Eslint rule and applied it.

Checklist:
- [x] tests updated


msimerson:
github workflow test failures are transient DNS (SPF) timeouts.

Issue #2708, opened 2019-10-01
Add option to enforce TLSV1.2


The-Nutty:
### Expected behavior
There should be a way to enforce using TLSV1.2.
### Observed behavior
Currently for some reason haraka is currently negitiating with clients to use. 
The following is from a connection.tls object in the data_post hock.
```
    "cipher": {
        "name": "ECDHE-RSA-AES128-GCM-SHA256",
        "version": "TLSv1/SSLv3"
    },
```

Further more i am seeing the following header in emails
```
Received: from sending-server (mail.sending-server.domain [1.1.1.1])

	by my-server.domain (Haraka/2.8.24) with ESMTPS id 07416364-7082-4B6B-A773-581B2714D96B.1

	envelope-from <me@gmail.com>

	(version=TLSv1/SSLv3 cipher=ECDHE-RSA-AES128-GCM-SHA256 verify=FAIL);
```
### Steps to reproduce
Use haraka with the tls plugin
```
key=key.pem
cert=cert.pem
```
I have also tried adding the following to no luck 
```
secureProtocol=TLSv1_2_method
ciphers=ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES128-SHA256:ECDHE-RSA-AES256-SHA384:DHE-RSA-AES256-SHA384:ECDHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA256:HIGH:!aNULL:!eNULL:!EXPORT:!DES:!RC4:!MD5:!PSK:!SRP:!CAMELLIA
```

msimerson:
Fixed in #2648 and released in Haraka 2.8.25.

danieltigse:
quick question. So If I use `secureProtocol=TLSv1_2_method` am I enforcing the connection to use TLS V1.2?

Issue #2634, opened 2019-04-15
Remove TLSSocket peer certificate verification info from sslheader


andris9:
The `Received` header has a TLS section in it ([defined here](https://github.com/haraka/Haraka/blob/6913ae3479682cee60a848031dc898dfa7708f5f/connection.js#L1446-L1461)) that lists `cipher`, `version` and `verify` properties. I'd suggest making the following changes:

* Remove `version` property. It is a legacy field in Node TLS object and [always has the same value](https://nodejs.org/api/tls.html#tls_tlssocket_getcipher) (TLSv1/SSLv3). Some people freak out on seeing SSL in mail headers
* Remove `verify` property or alternatively only keep it if there is a valid `peerCertificate`. Currently all messages get `verify=FAIL` in the `Received` header since no clients that connect to a MX server have a valid client certificate.

Current:

```
Received: from...
  (version=TLSv1/SSLv3 cipher=ECDHE-RSA-AES128-GCM-SHA256 verify=FAIL)
```

Suggested:

```
Received: from...
  (cipher=ECDHE-RSA-AES128-GCM-SHA256)
```

msimerson:
> since no clients that connect to a MX server have a valid client certificate

Some clients do have valid certs. That ‘fail’ does mislead plenty of people (there are issues here created about it). Maybe remove that property unless it’s true?

PR #2723, opened 2019-10-13
restore TLS version, correctly

msimerson:
Related to: #2634, #2648, and #2708 

- add cipher version back to header
- this time, use the current connections version, not the minimum supported

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -3,6 +3,8 @@
 
 ### Changes
 
+* restore TLS version info, set correctly #2723
+
 ### New features
 
 ### Fixes
--- a/connection.js
+++ b/connection.js
@@ -1449,7 +1449,7 @@ class Connection {
         let sslheader;
 
         if (this.get('tls.cipher.version')) {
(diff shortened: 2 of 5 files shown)


msimerson:
`<disclaimer>`It's a small sample size but`</disclaimer>` every TLS connection that has arrived since I ported this to a production server (granted, on a pretty slow Sunday afternoon) has been `TLSv1.2`. 

I also wish our tls connection properties were named `tls.cipher` and `tls.version` (instead of `tls.cipher.name` and `tls.cipher.version`, but I don't want to break any existing use of them for the sake of making them a little tidier.

msimerson:
More data. Measured across 70,000 transactions (ie, connections that made it to at least the `MAIL FROM` phase of the connection:

- TLSv1.2 = 98.2%
- TLSv1.1 = 0.6%  (just two connections, both spam, from germany and vietnam)
- TLSv1 = 0.6%  (dozens of valid connections.)

PR #2729, opened 2019-11-04
use npx in package.json scripts

msimerson:
### Changes

- package.json: use npx in scripts (works everywhere- locally, GitHub Actions, Travis, AppVeyor)
- tls_socket: process all entries in `config/tls dir`. Uses async.map so previous versions would abort after the first error in a TLS cert was encountered. This surfaced in Windows test where initial err aborted processing of tls dir contents.
- ignore an SPF error (when DNS lookup fails) when testing on Windows

### Checklist

- [x] Changes updated
- [x] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -10,6 +10,9 @@
 
 ### Fixes
 
+* TLS: don't abort loading certs in config/tls dir when an error is encountered.
+  Process every cert file and then emit errors. #2729
+
 
 ## 2.8.25 - 2019-10-11
 
--- a/package.json
+++ b/package.json
@@ -89,8 +89,8 @@
   },
   "scripts": {
     "test": "node run_tests",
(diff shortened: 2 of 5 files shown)


baudehlo:
Is this even necessary? Node adds node_modules/.bin to the PATH anyway, so just `eslint` on its own should work fine.

msimerson:
Generally, yes, if one values consistency. Do you want the tests run with the locally installed version of $dep in ./node_modules/.bin or the system installed one in /usr/local/bin or /usr/bin? Which is first in `PATH`?  Is that the same on every platform?

msimerson:
> Generally, yes, if one values consistency.

I'm going to amend that slightly. I haven't tested this in a couple years. It used to that the only way to get consistent passing CI tests on both Travis & Appveyor was to fully qualify the paths to the locally installed deps _and_ precede that path with `node`. Nothing else consistently worked. Since then we have dropped a couple older versions of node.

Issue #2176, opened 2017-10-03
Damaged encoding when body is non-utf


analogic:
I've found that some delivered emails are completely unreadable - typicaly non-utf8 plaintext emails. After some time of investigating I've come to this code https://github.com/haraka/Haraka/blob/master/transaction.js#L95-L108

it basicaly does this
```javascript
            let new_line = line.toString('UTF-8');
            ...
            line = new Buffer(new_line, 'UTF-8');
```
The catch is that unicode or ascii will pass this transcoding ok, but with other encodings (I am in country where iso-8859-2 or windows-1250 is still used) it will damage output and mail became unreadable.

I would send pull request but I am still missing bigger picture and also bit confused why body parsing alters actual email if plugin doesn't do any changes (attachments.js and queued through qmail plugin)...

example of damaged email, input:

```
...
MIME-Version: 1.0
Content-Type: text/plain; charset=iso-8859-2; format=flowed
Content-Language: en-US
Content-Transfer-Encoding: 8bit

Příliš žluťoučký kůň úpěl ďábelské ódy.
```

output:

```
...
MIME-Version: 1.0
Content-Type: text/plain; charset=iso-8859-2; format=flowed
Content-Language: en-US
Content-Transfer-Encoding: 8bit

Pďż˝ďż˝liďż˝ ďż˝luďż˝ouďż˝kďż˝ kďż˝ďż˝ ďż˝pďż˝l ďż˝ďż˝belskďż˝ ďż˝dy.

```

smfreegard:
Yeah - this looks wrong.  @baudehlo - the punycode stuff you added changed the encoding from 'binary' to 'utf-8' in #1944, but I can't work out why that was necessary - I can understand the need for headers, but for the message body as well?

PR #2737, opened 2019-11-15
fix bannering on nested mime parts

### LINKED ISSUE #2736: bannering doesn't work with nested mime parts
### system info

Haraka | (master)
 --- | :--- 
Node | v12.13.0
OS | Linux chazblip 4.15.0-62-generic #69-Ubuntu SMP Wed Sep 4 20:55:53 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
openssl | OpenSSL 1.1.1  11 Sep 2018

### Expected behavior

Enabling bannering causes the banners to be applied to the correct mime parts, even when those mime parts are nested.

### Observed behavior

Banners are not applied to messages with a nested structure, like the following:

```
Content-Type: multipart/mixed; boundary="TOP_LEVEL"

--TOP_LEVEL
Content-Type: multipart/alternative; boundary="INNER_LEVEL"

--INNER_LEVEL
Content-Type: text/plain; charset=us-ascii

Hello, this is a text part
--INNER_LEVEL
Content-Type: text/html; charset=us-ascii

<p>This is an html part</p>
--INNER_LEVEL--
--TOP_LEVEL--
```

### Steps to reproduce

Enable bannering (e.g. `transaction.set_banner(...)` from a plugin) and receive a message with a nested mime structure.  The banner is not applied.

I have a fix for this that I'll write up shortly.

chazomaticus:
Fixes #2736

This updates the bannering code to work even on nested mime structures.  I've included a new test ("bannering with nested mime structure" in tests/transaction.js) that shows off the issue.

Unfortunately the fix is a bit complicated.  At first I had just updated transaction.js to properly capture the output of `body.parse_more()` where it was being dropped before, using the existing string manipulation to add/remove CR chars and do dot-stuffing.  I couldn't get that not to mess up text encoded in non-UTF-8 encodings (e.g. the tests/transaction.js test for #2176) unfortunately, so I ended up rewriting the CR/dot-stuffing handling to work on buffers directly instead of strings.  That requires some changes to mailbody.js as well to handle buffers instead of strings.  The whole thing is kind of large for what seems like a simple fix, but I believe the code works well (the only thing is it might run slightly hotter than it did using strings, but at our volume it's been pretty negligible).

There's also a battle-tested fix to attachment_stream.js that I added because when we applied this, we were seeing workers stop responding, looping infinitely doing nothing.  I couldn't come up with a separate test to show off that behavior, but the bug is pretty obvious when you see the change there.

Please let me know what you think, or if I can explain anything in more detail.

Checklist:
- [ ] docs updated
- [x] tests updated


chazomaticus:
Oh shoot there may have been other test changes I forgot to peel off our version, I'll update this shortly.

baudehlo:
I still have the tab open. Keep meaning to get to it...

mailbody.js
```diff
@@ -44,21 +50,27 @@ class Body extends events.EventEmitter {
...
+
         return this[`parse_${this.state}`](line);
     }
 
     parse_child (line) {
+        const line_string = line.toString();
+
         // check for MIME boundary
-        if (line.substr(0, (this.boundary.length + 2)) === (`--${this.boundary}`)) {
+        if (line_string.substr(0, (this.boundary.length + 2)) === (`--${this.boundary}`)) {
```
baudehlo: Would this not be faster leaving as a Buffer and using `buf.compare()`?

chazomaticus: I was defaulting to doing all these comparisons as strings, since we end up needing string versions anyway, when we emit the boundary.  (I think it makes sense to emit strings for the boundaries, it just fits my expectations better than using buffers there.)  This more closely matches the existing code, so it feels less risky too.

It seems reasonable to me that we might gain some speed out of comparing as buffers and only converting to strings if necessary.  I don't have a good intuition of the performance characteristics around this.  If you think it'd be significant, I could try coding it up -- let me know how you feel about it.

baudehlo: I'm sort of "meh" on it. Once you enable mail parsing in Haraka most performance gains are gone anyway.

Merging.

chazomaticus:
Thanks!  That was pretty much my feeling about the performance too.  Cheers!

PR #2738, opened 2019-11-15
add minVersion to tls.ini options

analogic:
- add configurable minVersion to tls socket options

Checklist:
- [x] docs updated
- [x] tests updated


docs/plugins/tls.md
```diff
@@ -97,7 +97,17 @@ A list of allowable ciphers to use. Example:
...
-See also: [Strong SSL Ciphers](http://cipherli.st) and the [SSLlabs Test Page](https://www.ssllabs.com/ssltest/index.html)
+See also: [Mozilla SSL configuration generator](https://ssl-config.mozilla.org/) and the [SSLlabs Test Page](https://www.ssllabs.com/ssltest/index.html)
+
+### minVersion
+
+Specifies minimum allowable TLS protocol version to use. Example:
+
+     minVersion=TLSv1.1 
+
+If not set Haraka will use by default node's tls.DEFAULT_MIN_VERSION constant.
```
msimerson:
```suggestion
If unset, node's default tls.DEFAULT_MIN_VERSION constant.
```

docs/plugins/tls.md
```diff
@@ -97,7 +97,17 @@ A list of allowable ciphers to use. Example:
...
+
+(**Node.js 11.4+ required**, for older instances you can use *secureProtocol* settings)
```
msimerson: With this being the case, we *could* have the plugin check the running node version and map this that setting. That seems like a nicer solution than documenting the issue and foisting it on the user.

analogic: Problem is that minVersion setting is not equal to obsolete secureProtocol, it works in slightly different way...

msimerson: fair enough

PR #2740, opened 2019-11-22
add ES2017 support, drop node 8

msimerson:
- node 8 maintenance ends soon, drop it from testing
- add eslint ecmaVersion: 8 (2017), to add async/await support
- ignore a couple frequent DNS failures in tests (on GitHub Actions & AppVeyor)

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -3,7 +3,13 @@ env:
   node: true
 
 parserOptions:
-  ecmaVersion: 8
+  #ecmaVersion: 2015 (6)
+  #ecmaVersion: 2016 (7) (node 6)
+  #ecmaVersion: 2017 (8) (node 8)
+  ecmaVersion: 2017
+  #ecmaVersion: 2018 (9) (node 10)
+  #ecmaVersion: 2019 (10) (node 12)
+  #ecmaVersion: 2020 (11) (node 14)
 
 plugins:
   - haraka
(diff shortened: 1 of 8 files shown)


celesteking:
Why not start dropping semicolons?

celesteking:
ECMAScript since its earliest days supported automatic semicolon inference/insertion, of course. But people coding ECMAScript 5 started it in an era where lots of tools (especially source code compressors) had problems when semicolons where left out from the source code. As a consequence, most ECMAScript 5 coders maintained the traditional coding style with semicolons as the syntactic sugar, although in most of the cases they are not necessary. But this era is gone today. Both ECMAScript 6 and all tools (including compressors) perfectly support automatic semicolon inference/insertion nowadays. As a consequence, ECMAScript 6 coders nowadays can get rid of nearly all semicolons and remove clutter from their source code. Ralf S. Engelschall is a strong promoter of reducing source code to its bare minimum. Hence, in his personal opinion ECMAScript 6 should be coded with as less syntactic sugar as possible and hence without semicolons.

msimerson:
I'm not opposed. In new JS projects, I use them *only* when necessary.

celesteking:
I propose the following:
* Not using semicolons in any new code
* Removing semicolons in old code that's being modified, changed or just lying nearby (e.g., inside the same function)

All this to remove the necessity of going through entire codebase and making a huge semicolons-only change commit.

PR #2768, opened 2020-03-01
Allow web interface to be bound to a unix socket

pjeby:

To better support embedded use of Haraka, this change adds support for specifying listener addresses as /path/to/some.sock[:mode], where `mode` is an optional octal permissions spec for chmod.  Removal of an existing socket is handled automatically.

This change also consolidates listening-endpoint address parsing and formatting to a single class, removing duplicated code and simplifying the setup of servers.

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -12,6 +12,7 @@
 * connection_close_message: added ability to override close connection message replacing `closing connection. Have a jolly good day.`
 * tls: add configurable minVersion to tls socket options
 * add JSON format for logging
+* support binding web interface to unix socket
 
 ### Fixes
 
--- a/config/http.ini
+++ b/config/http.ini
@@ -3,5 +3,9 @@
 ; default: [::]:80 (port 80 on all IPv4 and IPv6 addresses)
 ; listen=[::]:80
 
+; listen can also be a unix socket path, with an optional 3-digit permission mask
+; e.g. listen=/path/to/some.sock or listen=/path/to/some.sock:777
+; if no mask is specified, the default permissions are determined by the umask.
+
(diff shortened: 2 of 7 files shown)


pjeby:
I haven't come up with anything either.  I mean, you could call it a listener factory or endpoint builder or something like that, but I'm not sure it's any clearer.

pjeby:
Thanks!  Any thoughts on when the next release will be?

msimerson:
### Thoughts on releases

* Releases happen when something nudges us to cut a new release. A significant new feature or bug fix is a good example.
* We have a [release checklist](https://github.com/haraka/Haraka/wiki/Release-Process) that documents the necessary steps.
* I cut most of the releases. I typically do so when I have 3-4 days of "attention available" time. Then I check out from master and run Haraka in a couple production environments for a few days. If anything blows up, I fix/revert the issue(s) and then proceed to rolling the release. I won't have that much time available until at least next week, but also have other priorities more important to me than a Haraka release, like a 20 mile ski race this Saturday.
* This PR is important to you. If you want to speed the release process, you can initiate the process per the wiki page.

pjeby:
No pressure was intended, sorry if it came across that way.  I was mostly just curious if there was a set release schedule or process, and you've answered that.  ;-)

For my use case I'm actually monkeypatching Haraka to address the websocket thing, I created this PR mostly to avoid the need to do that long-term, but don't have any set timeframe required.

In fact, as long as there *isn't* a new release, my monkeypatch can't possibly break, so I'm sure it'll be just fine till then.  ;-)

PR #2775, opened 2020-03-20
Better error message for invalid EHLO hostname

gramakri:
We used to have this error message in fc4752eb360a63fc66ddc194c1f4eac052b4cf84 

- Better error message

Checklist:
- [X ] docs updated
- [X ] tests updated


plugins/helo.checks.js
```diff
@@ -170,7 +170,7 @@ exports.valid_hostname = function (next, connection, helo) {
     if (!/\./.test(helo)) {
         connection.results.add(plugin, {fail: 'valid_hostname(no_dot)'});
         if (plugin.cfg.reject.valid_hostname) {
-            return next(DENY, 'Host names have more than one DNS label');
+            return next(DENY, 'HELO host name must have a dot');
```
msimerson: This is a SMTP denial/rejection message, not an error message.

I also don't agree this is a better message.

````
   Local nicknames or
   unqualified names MUST NOT be used.  There are two exceptions to the
   rule requiring FQDNs:

   o  The domain name given in the EHLO command MUST be either a primary
      host name (a domain name that resolves to an address RR) or, if
      the host has no name, an address literal, as described in
      Section 4.1.3 and discussed further in the EHLO discussion of
      Section 4.1.4.
````

While a primary domain name / FQDN and address literals both do have dots, that's a proxy for the requirements and not a very useful rejection message. How about this instead:

````
return next(DENY, 'HELO host must be a FQDN or address literal (RFC 5321 §2.3.5');
````

gramakri:
`Host names have more than one DNS label`

Now that I read this message again, it makes a lot of sense. For some reason, the other day when I was debugging, this error message felt very cryptic.

@msimerson thanks for the review, I will close this one.

msimerson:
I agree, it is a little cryptic, especially if you aren't a SMTP guru. Expanding it similar to what I proposed above would be an improvement.

gramakri:
@msimerson I will fix the PR.

PR #2776, opened 2020-03-23
Fix support for DKIM signing when forwarding and aliasing is enabled

### LINKED ISSUE #2774: DKIM_sign tries to sign using the envelop from
### system info

system info

Haraka | Haraka.js — Version: 2.8.23
 --- | :---
Node | v10.15.0
OS | Linux cyril 4.9.185-xxxx-std-ipv6-64 #799710 SMP Wed Jul 10 08:31:28 UTC 2019 x86_64 GNU/Linux
openssl | OpenSSL 1.1.0l  10 Sep 2019

$ grep -v "^#" /var/lib/haraka/config/plugins
syslog
dnsbl
helo.checks
tls
rcpt_to.alias_forward2
data.headers
dkim_sign
queue/discard
limit

$ cat dkim_sign.ini
[main]
disabled = false
selector = mars2020
domain = mydomain.com
headers_to_sign = To, From, Subject, Content-Type, Message-ID, Date
dkim.private.key = /var/lib/haraka/config/dkim/mydomain.com/private

### Expected behavior
When using the dkim_sign plugin with an alias/forward plugin, the dkim_sign should use the current server's domain and not the envelope's from (or if missing the From header).

### Observed behavior

In bold, see the dkim_sign error
#### Extract of logs:
```
[INFO] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core]  hook=rcpt plugin=rcpt_to.alias_forward2 function=alias_forward params=<fwd.me@mydomain.com> retval=OK msg=""
[DEBUG] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] running rcpt_ok hooks
[NOTICE] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] recipient <fwd.me@mydomain.com> code=OK msg="" sender=me@gmail
[PROTOCOL] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] S: 250 recipient <fwd.me@mydomain.com> OK
[PROTOCOL] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: DATA state=1
[DEBUG] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] running data hooks
[PROTOCOL] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] S: 354 go ahead, make my day
**[DEBUG] [-] [dkim_sign] undefined
[NOTICE] [-] [dkim_sign] skipped: no private key for todomain.com**
[DEBUG] [-] [core]  hook=pre_send_trans_email plugin=dkim_sign function=hook_pre_send_trans_email params="" retval=CONT msg=""
[INFO] [-] [core] [outbound] Processing delivery for domain: todomain.com
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: Received: from MacBook-Pro-de-me.local (unknown [myip])\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: 	(Authenticated sender: me)\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: 	by smtp3-g21.todomain.com (Postfix) with ESMTPSA id B47B213F8AE\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: 	for <fwd.me@mydomain.com>; Thu, 19 Mar 2020 19:03:18 +0100 (CET)\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: To: fwd.me@mydomain.com\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: From: Me <me@todomain.com>\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: Subject: Test\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: Message-ID: <554b28ee-a5f9-633c-953c-d5593457b732@free.fr>\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: Date: Thu, 19 Mar 2020 19:03:17 +0100\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.13; rv:68.0)\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C:  Gecko/20100101 Thunderbird/68.6.0\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: MIME-Version: 1.0\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: Content-Type: text/plain; charset=utf-8; format=flowed\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: Content-Transfer-Encoding: 7bit\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: Content-Language: fr\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: \r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: Test\r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: \r
[DATA] [0A24668F-A8FE-4A00-AE4D-3C6DE6EB114C.1] [core] C: .\r
```

### Steps to reproduce
Enable a forwarding plugin like alias_forward or alias_forword and set up DKIM.

X-Ryl669:
This partially solve DMARC failing and forwarded message considered as spam.
Fixes #2774

The other requirement are out of haraka's repository. One must enable SRS plugin (or use my alias_forward plugin which does both).


plugins/dkim_sign.js
```diff
@@ -355,7 +355,8 @@ exports.get_sender_domain = function (connection) {
         }
     }
 
-    if (!txn.header) return domain;
+    // In case of forwarding, only use the Envelope
+    if (!txn.header || txn.notes.forward) return domain;
```
msimerson: Should this be a compound conditional? Does something else set `txn.notes.forward`?

msimerson:
Forgive me if any of these are dumb questions. You're the most recent person to think through this issue so I'm asking now.

There's several cases here and I'd like someone to be confident that this change does The Right Thing in all of these cases:

1. typical outbound (From header, then ENV FROM)
2. forwarding (?)
3. aliasing (?)
4. forwarding and aliasing (ENV FROM)

plugins/dkim_sign.js
X-Ryl669: The alias forwarding plugins sets them (at least mine does). See [here](https://github.com/X-Ryl669/haraka-alias-forward/blob/356ade4d75c575d2417cf541d4ac2766689177ae/plugins/rcpt_to.alias_forward.js#L40)

X-Ryl669:
> * typical outbound (From header, then ENV FROM)
> 
As previously

> * forwarding (?)

If you forward a mail without any modification, you don't need to sign it since any previous signature remains valid. You might need to use SRS if the source domain is using SPF. 
I'm not sure what is the difference here with plain relaying. For me, this situation is a bit strange since I don't think of any reason a SMTP server would connect to your domain to send an email to another domain (it should connect to the other domain directly no?)

 
> * aliasing (?)
Aliasing only means you are the recipient SMTP server (typically, you received `test@domain.com` and you alias to `bob@domain.com`) so you'll never enable DKIM sign here but DKIM verify (so this change does not apply).
 
> * forwarding and aliasing (ENV FROM)
This PR, no need to explain further.

plugins/dkim_sign.js
msimerson: I understand that, but why make that a compound conditional?  (those tend to increase in complexity over time.). Why not simply do this?

```js
if (txn.notes.forward) return domain; // forwarding, only use the Envelope
```

msimerson:
> For me, this situation is a bit strange since I don't think of any reason a SMTP server would connect to your domain to send an email to another domain (it should connect to the other domain directly, shouldn't it?)

This is a very frequent use of a SMTP server. The typical description for a MTA like Haraka/Postfix/Sendmail/Exim serving in this role is a [Message Submission Agent](https://en.wikipedia.org/wiki/Message_submission_agent).

You keep equating a mail server with a domain and that's too narrow. Many (most?) MTAs serve more than one domain. My own Haraka instance accepts and relays mail for many domains.

plugins/dkim_sign.js
X-Ryl669: Sure that's possible if you prefer.

msimerson: > Sure that's possible if you prefer.

Simpler is better, and that's simpler. Anyone can glance at that and know exactly when it'll execute.

X-Ryl669:
Ok. If you forward but don't set `notes.forward` to true, then nothing is going to change from previous behavior. Do you have `rcpt_to.alias_forward` in your plugin list ?

msimerson:
> Ok. If you forward but don't set notes.forward to true, then nothing is going to change from previous behavior. Do you have rcpt_to.alias_forward in your plugin list?

I don't. I have some aliases I want to move into Haraka but have yet to do it.

X-Ryl669:
Here's the adapted code. To answer the question about forwarding only, in case you are using a Haraka as a MSA, if I understand correctly, the client should provide a `mail_from` value, so `domain` will be set. If you don't enable aliasing, then no change to the previous behavior will be done. If you enable aliasing then only mail who are accepted by the aliasing plugin will have their transaction modified to include `forward = true`, as expected. 

If you intend to implement aliasing please check the 2 tutorials I've written [here](https://steemit.com/haraka/@xryl669/forwarding-email-to-your-domain-with-generic-email-naming-pattern-update-2020) as it is easier to copy & modify than trying to understand each plugin specificity (DKIM_sign is a real pain to set up right since it implies so many external steps).

msimerson:
I've looked at the adapted code, looks the same as before. Am I missing something?

X-Ryl669:
Ahem*... Cough... Well sorry, forgot to commit my changes before amending. Should be ok now.

PR #2782, opened 2020-04-08
Fix outbound.bounce_message To: header (and add Auto-Submitted)

sriccio:
While dealing with an infinite bounce loop, I've noticed that the To: header in the example template for outbound.bounce_message is {from}.

I guess this should be {to} (or am I wrong?)

I've added Auto-Submitted header also to indicate this is an automated reply.

Checklist:
- [ ] docs updated 
- [ ] tests updated


msimerson:
> I guess this should be {to} (or am I wrong?)

The bounce is a new message that we're creating, and we're sending it back to the sender. Right? Therefore, the {to} address should be {from}. I think.

sriccio:
Hi,

Well, I'm not really sure, that's why I ask. 

But in our case haraka was setting <> as the To: header field because the MAIL FROM (sender enveloppe) from was <> (a mailer daemon).

Setting it to {to} instead of {from} resolved the issue because then the To: header was set to the correct recipient instead of <>.

Not sure i'm really clear in explaining here :)

sriccio:
Hello,

To be sure I wasn't completly out of my mind, I did a new simple test:

Scenario:

I'm sending a mail from a mailbox on a mailserver that uses haraka as outgoing gateway to relay the mail, this to a invalid recipient (the target user doesn't exists on target mail system).

Here are the bounce messages I receive:

1) When using `To: {from}` in the `outbound.bounce_message` file:
mailer_daemon_with_from
Headers found in returned bounce message:
```
Return-Path: <>
[cut]
From: MAILER-DAEMON@mta-gw.swisscenter.net
To: <>
[cut]
```

2) When using `To: {to}` in the `outbound.bounce_message` file:
mailer_daemon_with_to

Headers found in returned bounce message:
```
Return-Path: <>
[cut]
From: MAILER-DAEMON@mta-gw.swisscenter.net
To: <originalsender@mailserver.ch>
[cut]
```

So it looks that actually `{from}` `{to}` variables actually resolves to the values for the outgoing bounce, not the original message, therefore using `To: {from}` resolves to `<>` which is the correct from for a bounce, and `To: {to}` resolves to recipient of the bounce (the original sender).

Does it makes sense ?

Thank you for reading :)
Kind regards

PR #2792, opened 2020-04-28
STARTTLS "TLS NO-GO" improvements

celesteking:
Right now, when there's a STARTTLS failure, we note the host in redis. It connects again and we skip the STARTTLS capability in addition to clearing the `no_tls` flag in redis. Which means next time it connects, STARTTLS is advertised again and the process repeats. 

There are 2 downsides to this:
* client on win7 (tlsv1.0) connecting to nodejs >11 (min tlsv1.2 unless passed a CLI flag) will get a failure, then no STARTTLS cap (and essentially an error on the client because of missing STARTTLS cap), then rinse and repeat...
* the `no_tls|1.2.3.4` flag in redis never expires and this has potential to clash with outbound's TLS NO-GO feature where we mark remote host as TLS NO-GO for `tls.ini/redis.disable_expiry` secs. 

My proposal is to align STARTTLS with Outbound TLS blackout code. Another config param is introduced, `disable_inbound_expiry` that will "blackout" STARTTLS for remote host for that many seconds.

Checklist:
- [x] docs updated
- [?] tests updated

### DIFF
--- a/config/tls.ini
+++ b/config/tls.ini
@@ -42,8 +42,11 @@
 ; host = 127.0.0.1
 ; "TLS NO-GO" db
 ; db = 3
-; Expiry time in seconds
+; TLS NO-GO Expiry time in seconds
 ; disable_expiry = 604800
+
+; TLS NO-GO Inbound expiry time in seconds
+; disable_inbound_expiry = 3600
 
 ; no_tls_hosts - disable TLS for servers with broken TLS. (applies to inbound only)
 [no_tls_hosts]
--- a/plugins/tls.js
+++ b/plugins/tls.js
@@ -8,12 +8,12 @@ const tls_socket = require('./tls_socket');
 exports.net_utils = require('haraka-net-utils');
 
 exports.register = function () {
-    const plugin = this;
+    tls_socket.load_tls_ini()
 
-    tls_socket.load_tls_ini();
+    if (tls_socket.cfg.redis.disable_for_failed_hosts) this.logdebug('Will disable STARTTLS for failing TLS hosts')
 
-    plugin.register_hook('capabilities',         'advertise_starttls');
-    plugin.register_hook('unrecognized_command', 'upgrade_connection');
+    this.register_hook('capabilities',         'advertise_starttls')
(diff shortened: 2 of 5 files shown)


plugins/tls.js
```diff
@@ -57,21 +57,21 @@ exports.advertise_starttls = function (next, connection) {
 
         if (!dbr) return enable_tls();
 
-        // last TLS attempt failed
-        redis.del(dbkey); // retry TLS next connection.
-
         connection.results.add(plugin, { msg: 'no_tls'});
-        return next();
+        return next(CONT, 'STARTTLS disabled because previous attempt to set it up failed')
```
msimerson, line 61:
```suggestion
        return next(CONT, 'STARTTLS disabled because previous attempt failed')
```

plugins/tls.js
```diff
@@ -57,21 +57,21 @@ exports.advertise_starttls = function (next, connection) {
...
     });
 }
 
-exports.set_notls = ip => {
+exports.set_notls = function (connection) {
 
-    if (!tls_socket.cfg.redis) return;
```
msimerson, line 70: Leave this line here, else Haraka crashes while attempting to access a config property that doesn't exist in older config files / versions of Haraka.

config/tls.ini
```diff
@@ -42,9 +42,12 @@
 ; host = 127.0.0.1
 ; "TLS NO-GO" db
 ; db = 3
-; Expiry time in seconds
+; TLS NO-GO Expiry time in seconds
 ; disable_expiry = 604800
 
+; TLS NO-GO inbound expirty time in seconds
```
msimerson, line 48:
```suggestion
; TLS NO-GO inbound expiry time in seconds
```

plugins/tls.js
celesteking: No, it won't, because that setting is specified in config boolean list https://github.com/haraka/Haraka/blob/master/tls_socket.js#L221

msimerson: 👍

celesteking:
Have added some docs. I've also modded `outbound/tls.js ` to store ISO date and not a locale/system specific string.

PR #2796, opened 2020-05-06
smtp_forward: remove redundant outbound hook

### LINKED ISSUE #2795: smtp_forward: remove redundant outbound hook
### Expected behavior

When using smtp_forward for outbound emails, the mail gets delivered via the SMTP mail delivery queue (just like normal emails). I expect `delivered` hook to be called after mail is delivered, but it's not.

### Observed behavior

Mail is delivered via the  `queue_outbound` hook. Because, it ties into this hook, if a mail gets delivered, the `delivered` hook is never called.

Instead, by relying on the `get_mx` hook, the email flow is now the same for email when using smtp_forward or not.

gramakri:
Fixes #2795

Fixes #

- Remove the redundant use of outbound hook

Checklist:
- [X ] docs updated
- [X ] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -21,7 +21,7 @@
 * TLS: don't abort loading certs in config/tls dir when an error is encountered.
   Process every cert file and then emit errors. #2729
 * fix connection pool not being unique when hosts and ports were equal between domains #2788
-
+* smtp_forward: this plugin does not use the queue_outbound hook anymore #2795
 
 ## 2.8.25 - 2019-10-11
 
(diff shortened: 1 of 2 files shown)


celesteking:
Makes sense to slap a 3.x tag onto this breaking change as some might expect particular behavior, even if it's logically wrong.

Changes.md
```diff
@@ -21,7 +21,7 @@
 * TLS: don't abort loading certs in config/tls dir when an error is encountered.
   Process every cert file and then emit errors. #2729
 * fix connection pool not being unique when hosts and ports were equal between domains #2788
-
+* smtp_forward: this plugin does not use the queue_outbound hook anymore
```
msimerson, line 24: can you specify in this PR when it stopped using that hook?

gramakri: Will do

gramakri: Done

PR #2797, opened 2020-05-06
use node v10's mkdir instead of mkdirp

solderjs:
Just a suggestion.

- use node v10's `fs.mkdir` rather than dependening on `mkdirp`

Checklist:
- **N/A** [ ] docs updated
- **N/A** [ ] tests updated

### DIFF
--- a/bin/haraka
+++ b/bin/haraka
@@ -10,7 +10,6 @@ const nopt = require('nopt');
 const path = require('path');
 const os   = require('os');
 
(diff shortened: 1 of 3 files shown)


msimerson:
@solderjs , still waiting for the second call to mkdirp to get updated. Then tests should pass and this can be merged.

solderjs:
Just saw the message. Updated accordingly.

PR #2802, opened 2020-05-11
fix outbound config reload

celesteking:
Some smartass decided to split outbound into pieces and forgot to code the config reload properly.

Checklist:
- [ ] docs updated
- [x] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -1,6 +1,8 @@
 
 ## NEXT - 20YY-MM-DD
 
+* fix outbound config reloading after outbound has been split off into pieces
+
 ### Changes
 
 * lint: add 'prefer-template'
--- a/outbound/client_pool.js
+++ b/outbound/client_pool.js
@@ -7,21 +7,21 @@ const sock         = require('../line_socket');
 const server       = require('../server');
 const logger       = require('../logger');
 
-const cfgmod       = require('./config');
+const obc          = require('./config');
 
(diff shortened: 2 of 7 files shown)


celesteking:
Strangely, no one noticed this before. And I spent several hours to understand why outbound delivery suspension didn't work.

msimerson:
In principle, I agree with this change. Changing `cfg` to something else makes sense, because everywhere else in Haraka we use `cfg` to represent the parsed configuration. What I don't like is the name `cfgmod`. It's  too generic. Is it a modified config? A config module? Is it  **the** `haraka-config` module?

What it is is the *parsed outbound config*. That's a mouthful. So how about naming the variable obcfg, or ob_cfg?

celesteking:
You can't agree or disagree in this case as this is essentially a bugreport ;)

msimerson:
> Shall I do the needful?

Yes please. And then I will too.

celesteking:
Update: actually, I object ob_cfg as we'll have to use ob_cfg.cfg which doesn't look nice.

msimerson:
I agree, redundancy isn't a good look. Proposal: `const ob_cfg = require('./config').cfg;`

celesteking:
... have you seen the patch? We can't use that as cfg object is being reassigned on config reload. That was the whole point/issue.

msimerson:
Okay, `const obc = require('./config');` and then reference it with `obc.cfg`?

celesteking:
here you go

PR #2834, opened 2020-08-20
Prevent node crash with zero byte file

analogic:
When there is 0 byte file in queue it causes terrible node.js crash with ERR_BUFFER_OUT_OF_BOUNDS message

### DIFF
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -93,6 +93,11 @@ class HMailItem extends events.EventEmitter {
                 this.temp_fail(errMsg);
                 return
             }
+            
+            if (bytes.length === 0) {
+                this.logerror(`Error reading queue file ${this.filename}: no bytes read`);
+                return
+            }
 
             const todo_len = bytes.readUInt32BE(0);
             this.logdebug(`todo header length: ${todo_len}`);


msimerson:
@analogic , this change looks like it would work. However, I think that we should instead do it in the calling function where we are explicitly checking the file size.

```js
diff --git a/outbound/hmail.js b/outbound/hmail.js
index b58436c2..8e982776 100644
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -77,11 +77,16 @@ class HMailItem extends events.EventEmitter {
                 // we are fucked... guess I need somewhere for this to go
                 self.logerror(`Error obtaining file size: ${err}`);
                 self.temp_fail("Error obtaining file size");
+                return;
             }
-            else {
-                self.file_size = stats.size;
-                self.read_todo();
+
+            if (stats.size === 0) {
+                self.logerror(`Error reading queue file ${self.filename}: zero bytes`);
+                return;
             }
+
+            self.file_size = stats.size;
+            self.read_todo();
         });
     }
  
```

You can also get rid of the `const self = this` in that function as it's no longer needed.

PR #2863, opened 2020-11-19
dns: add dns.ini config file to set the default dns server list

gramakri:
This feature is useful if you have dns servers running on a non-standard port. This is useful when running haraka in Docker. The base OS usually has systemd-resolved or equivalent which is a forwarding name server. In contrast, haraka ideally requires a recursive nameserver (for dnsbl to work, many OS default to google nameservers which zen spamhaus does not respond to). Unfortunately, one cannot set the default name server in the container's /etc/resolv.conf because that file does not support custom ports.

- This calls dns.setServers() on startup. Both master and child workers in a cluster will get the setting
- [node docs](https://nodejs.org/docs/v10.0.0/api/dns.html#dns_dns_setservers_servers)

Checklist:
- [X] docs updated
- [ ] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -1,6 +1,9 @@
 
 ## NEXT - 2020-MM-DD
 
+### New features
+
+* dns: add configurable dns servers using dns.ini config file
 
 ## 2.8.26 - 2020-11-18
 
--- a/config/dns.ini
+++ b/config/dns.ini
@@ -1,0 +1,4 @@
+; servers: a comma separated list of rfc5952 formatted addresses
+; servers=8.8.8.8
+; servers=4.4.4.4,[2001:4860:4860::8888],4.4.4.4:1053,[2001:4860:4860::8888]:1053
+
--- a/docs/CoreConfig.md
+++ b/docs/CoreConfig.md
@@ -144,3 +144,9 @@ The list of plugins to load
 * connection\_close\_message
   
   Defaults to `closing connection. Have a jolly good day.` can be overrridden with custom text
+
+* dns.ini
+
+  The `servers` key contains a comma separated list of [RFC 5952](https://tools.ietf.org/html/rfc5952#section-6)
+  formatted addresses.
+
--- a/server.js
+++ b/server.js
@@ -6,6 +6,7 @@ const fs          = require('fs');
 const os          = require('os');
 const path        = require('path');
 const tls         = require('tls');
+const dns         = require('dns');
 
 // let log = require('why-is-node-running');
 const tls_socket  = require('./tls_socket');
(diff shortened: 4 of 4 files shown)


msimerson (approved):
This looks fine to me. I have a couple general questions for @core. 

1. Should the load_dns_ini function should go in haraka-net-utils instead?
2. Instead of loading another config file, should this be in a `[dns_servers]` section of smtp.ini?

smfreegard:
> This looks fine to me. I have a couple general questions for @core.
> 
> 1. Should the load_dns_ini function should go in haraka-net-utils instead?

No - I don't think so.

> 2. Instead of loading another config file, should this be in a `[dns_servers]` section of smtp.ini?

Yes, should really should be in smtp.ini IMO, there's no need for a separate config file.   Because there's very little to configure here, `servers=` could probably just be `dns_servers=` and it can then live in `[main]`.

gramakri:
@smfreegard @msimerson OK, I will move it to a key named `dns_servers` in the main section.

msimerson:
>> 2. Instead of loading another config file, should this be in a `[dns_servers]` section of smtp.ini?
> Yes, should really should be in smtp.ini IMO, there's no need for a separate config file.   Because there's very little to configure here, `servers=` could probably just be `dns_servers=` and it can then live in `[main]`.

Might there be other dns settings in the future?  If there are, grouping them in a `dns` section would make sense. 

```ini
[dns]
servers=....
```

msimerson (changes requested):
Switch status to changes pending.

gramakri:
Thanks for review but we found another way to workaround the DNS issue we had and we don't need this patch anymore. I will close this since I suspect it will bitrot if we don't use it!

msimerson:
Glad you got the issue resolved. It'd be useful if you posted a brief summary of your workaround.

msimerson:
Should anyone want/need this in the future, a patch like this to Haraka would be welcome:

```diff
diff --git a/config/smtp.ini b/config/smtp.ini
index 0ebd21a4..e9f5888a 100644
--- a/config/smtp.ini
+++ b/config/smtp.ini
@@ -59,3 +62,7 @@ max_lines=1000
 
 ; replace max_received_count
 max_received=100
+
+[dns]
+; servers: a comma separated list of rfc5952 formatted addresses
+; servers=4.4.4.4,[2001:4860:4860::8888],4.4.4.4:1053,[2001:4860:4860::8888]:1053
diff --git a/server.js b/server.js
index a776130f..ffe8c993 100644
--- a/server.js
+++ b/server.js
@@ -102,6 +111,11 @@ Server.daemonize = function () {
     }
 }
 
+Server.configDNS = (servers) => {
+    try       { require('dns').setServers(servers.split(','));        }
+    catch (e) { logger.logerror(`Invalid DNS servers: ${e.message}`); }
+}
+
```

gramakri:
@msimerson The issue isn't Haraka related but very specific to our platform (cloudron).

But a brief summary is: on ubuntu 20, systemd-resolved is the default DNS server. That DNS server is non-recursive and thus queries to dnsbl like zen spamhaus may get blocked (on many VPS, it just forwards to Google DNS). We have Haraka deployed in container with dnsbl plugin. The dnsbl lookups fail because spamhaus blocks Google DNS. So, the fix was to use an additional unbound server. We couldn't run unbound on port 53 because systemd-resolved is there and thus I made this patch so that unbound runs on some other port and Haraka can then contact unbound for DNS. Sadly, this was causing other issues because /etc/resolv.conf cannot take a DNS port number and we have to then patch every software like this one :/ The fix was that we decided to disable systemd-resolved and replace it with unbound for all cloudron installations. This came with it's own set of headaches because of resolvconf but that is a story for another day!

msimerson:
Sounds like fun! Like pulling out ones fingernails. 😬 Instead of running a recursing resolver on another port, did you consider/try binding unbound to another IP?

PR #2866, opened 2020-12-02
Configurable smtputf8

gene-hightower:
Fixes #

- 
- 

Checklist:
- [x] docs updated
- [ ] tests updated

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -918,8 +918,11 @@ class Connection {
                     `${this.local.host} Hello ${this.get_remote('host')}${this.ehlo_hello_message}`,
                     "PIPELINING",
                     "8BITMIME",
-                    "SMTPUTF8",
                 ];
+
+                if (!!this.cfg.main.smtputf8) {
+                    response.push("SMTPUTF8");
+                }
 
                 response.push(`SIZE ${this.max_bytes}`);
 
--- a/connection.js
+++ b/connection.js
@@ -920,7 +920,7 @@ class Connection {
                     "8BITMIME",
                 ];
 
-                if (!!this.cfg.main.smtputf8) {
+                if (this.cfg.main.smtputf8) {
                     response.push("SMTPUTF8");
                 }
 
--- a/docs/CoreConfig.md
+++ b/docs/CoreConfig.md
@@ -41,6 +41,7 @@ The list of plugins to load
     specify -1 to disable spooling completely or 0 to force all messages to be spooled to disk.
   * graceful\_shutdown - (default: false) enable this to wait for sockets on shutdown instead of closing them quickly
   * force_shutdown_timeout - (default: 30) number of seconds to wait for a graceful shutdown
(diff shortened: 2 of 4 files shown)


connection.js
```diff
@@ -918,9 +918,12 @@ class Connection {
                     `${this.local.host} Hello ${this.get_remote('host')}${this.ehlo_hello_message}`,
                     "PIPELINING",
                     "8BITMIME",
-                    "SMTPUTF8",
                 ];
 
+                if (!!this.cfg.main.smtputf8) {
```
msimerson, line 923:
```suggestion
                if (this.cfg.main.smtputf8) {
```

msimerson (changes requested):
Other than the lint failure due to "Redundant double negation", the implementation looks good. Please fix that and also add an entry to the Changes.md file. 

I'm also curious why you'd want to disable SMTPUTF8.

connection.js
gene-hightower: I'm new to JavaScript, won't this throw is this.cfg.main.smtputf8 is undef?

gene-hightower:
> I'm also curious why you'd want to disable SMTPUTF8.

In this particular setup I'm working on Haraka is not the final hop in the chain, and subsequent hops may or may not support the SMTPUTF8 extension.

connection.js
msimerson: Nope. JS isn't strongly typed. In a case like this, undefined is the same as "not true."

msimerson (dismissed):
Looks good. If you'll also update the Changes.md file, that'd be great.

```diff
diff --git a/Changes.md b/Changes.md
index 6398821c..a7df7637 100644
--- a/Changes.md
+++ b/Changes.md
@@ -1,6 +1,14 @@
 
 ## NEXT - 2020-MM-DD
 
+### Changes
+
+### New features
+
+- add ability to disable SMTPUTF8 advertisement #2866
+
+### Fixes
+
 
 ## 2.8.26 - 2020-11-18
 
```

msimerson (approved):

docs/CoreConfig.md
```diff
@@ -41,6 +41,7 @@ The list of plugins to load
     specify -1 to disable spooling completely or 0 to force all messages to be spooled to disk.
   * graceful\_shutdown - (default: false) enable this to wait for sockets on shutdown instead of closing them quickly
   * force_shutdown_timeout - (default: 30) number of seconds to wait for a graceful shutdown
+  * smtputf8 - (default: true) advertise support for SMTPUTF8
```
spathon, line 44: Isn't it default false?

msimerson: The default is enabled. See [here](https://github.com/haraka/Haraka/blob/master/server.js#L37)

spathon: Ah thanks :)

PR #2867, opened 2020-12-03
fix: plugin names: removing 'haraka-plugin-' from plugin names breaks…

AuspeXeu:
… with plugins 'known-senders' and 'recipient-routes' enabled (at least, maybe more); look up plugin by actual name when running hooks

### DIFF
--- a/plugins.js
+++ b/plugins.js
@@ -441,9 +441,10 @@ plugins.run_hooks = (hook, object, params) => {
     object.hooks_to_run = [];
 
     if (plugins.registered_hooks[hook]) {
+        const registered_plugins = Array.from(Object.values(plugins.registered_plugins).values());
         for (let i=0; i<plugins.registered_hooks[hook].length; i++) {
             const item = plugins.registered_hooks[hook][i];
-            const plugin = plugins.registered_plugins[item.plugin];
+            const plugin = registered_plugins.find(({name}) => name === item.plugin);
             object.hooks_to_run.push([plugin, item.method]);
         }
     }


msimerson:
You should tell us what `by actual name` means. Also, this seems like a PR that definitely should have a test or three to better define the behavior.

AuspeXeu:
I don't think I have the depth of understanding for this code base to actually know what to test here but basically this PR fixes #2870 which seems to be some sort of regression on how plugin names are handled.

msimerson:
Okay, I'll poke at it when I get some free cycles. I think it's best to get the plugins loading whether referenced by their short name, and take that into account when parsing the list of plugins to load.

msimerson:
I've looked at this again and honestly, the underlying problem is that:

• we (the authors of Haraka that added support for NPM packaged plugins) always assumed that plugins would be listed in the `config/plugins` file _without_ the `haraka-plugin` prefix.
• you expect that plugins should be listed on config/plugins with the `haraka-plugin` prefix.

I think this will fix the discrepancy.

```diff
 ✗ git diff plugins.js
diff --git a/plugins.js b/plugins.js
index 9dea5142..aff49d2b 100644
--- a/plugins.js
+++ b/plugins.js
@@ -320,6 +320,7 @@ plugins.load_plugins = override => {
     }
 
     plugin_list.forEach(plugin => {
+        if (plugin.substr(0,14) === 'haraka-plugin-') plugin = plugin.substr(14)
         if (plugins.deprecated[plugin]) {
             logger.lognotice(`the plugin ${plugin} has been replaced by '${plugins.deprecated[plugin]}'. Please update config/plugins`)
             plugins.load_plugin(plugins.deprecated[plugin]);
```

AuspeXeu:
I guess I assumed that whenever I would write a plugin called `my-awesome-plugin` I would add that full name into the plugins config. Hence I would do the same for `haraka-plugin-` prefixed ones. I think your suggestion to strip it from the config is a good way forward.

PR #2869, opened 2020-12-03
If the option is set, and to anything but true, we don't advertise th…

### LINKED ISSUE #2868: The new smtputf8 flag has changed the default (when not set) to disable the extension
### system info

GNU/Linux node 14

### Expected behaviour

When the smtputf8 flag is not set, or set to false, the extension should be advertised.

### Observed behavior

If the flag was not set at all the default was to disable the extension (not to advertise)

### Steps to reproduce

Don't set the flag, then:
telnet <host> smtp
Connected to localhost.
Escape character is '^]'.
220 example.net ESMTP Haraka is ready
ehlo example.com
250-example.net Hello Unknown [127.0.0.1]Haraka is ready to receive your message.
250-PIPELINING
250-8BITMIME
250-SIZE 0
250 STARTTLS
quit
221 example.net closing connection. Have a jolly good day.
Connection closed by foreign host.

gene-hightower:
…e SMTPUTF8 extension.

Fixes #2868

- 
- 

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -920,7 +920,7 @@ class Connection {
                     "8BITMIME",
                 ];
 
-                if (this.cfg.main.smtputf8) {
+                if (!(this.cfg.main.smtputf8 && this.cfg.main.smtputf8 === true)) {
                     response.push("SMTPUTF8");
                 }
 
--- a/connection.js
+++ b/connection.js
@@ -90,6 +90,7 @@ class Connection {
 
         this.cfg = config.get('smtp.ini', {
             booleans: [
+                '+main.smtputf8',
                 '+headers.add_received',
                 '+headers.show_version',
                 '+headers.clean_auth_results',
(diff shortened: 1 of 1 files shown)


msimerson:
Hi @gene-hightower , lets try instead making sure it's always defined:

```diff
diff --git a/connection.js b/connection.js
index 7a2a27bf..49b205e5 100644
--- a/connection.js
+++ b/connection.js
@@ -90,6 +90,7 @@ class Connection {
 
         this.cfg = config.get('smtp.ini', {
             booleans: [
+                '+main.smtputf8',
                 '+headers.add_received',
                 '+headers.show_version',
                 '+headers.clean_auth_results',
```

msimerson:
Related to #2866

gene-hightower:
Great!  Much better idea: how do we get this merged?

msimerson:
apply that patch to this branch, revert the previous change, and push. Then this PR will have that and I'll merge it

gene-hightower:
> apply that patch to this branch, revert the previous change, and push. Then this PR will have that and I'll merge it

Done. Thanks again for working with me to make this happen.

msimerson (approved):
Looks good!  Did you test locally and verify this works as expected?

gene-hightower:
> Looks good! Did you test locally and verify this works as expected?

Yes, I tested with my application and it works now just like it should. (I tested with true, false, and no value set. I did not test with other random values)

PR #2874, opened 2020-12-14
allow disabling STARTTLS "TLS NO-GO" feature on specific ports

celesteking:
This is related to #2792, #770,  and allows disabling the feature on specific ports.
Reasoning: STARTTLS sometimes gets disabled because MUA is misconfigured or too slow to set up a TLS session. MUAs should not be kicked out into insecure field, especially when they're set to use port 587 (vs. 25). Most modern MUAs are sane and refuse to AUTH when TLS checkbox is ticked in their settings.

This will disable TLS NO-GO mechanism on port 587 by default. 25 is unaffected.

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,6 +5,7 @@
 
 - strip _haraka-plugin-_ prefixes off plugin names in config/plugins #2873
 - pass smtp.ini config from Server into connections & transactions #2872
+- constrain STARTTLS "TLS NO-GO" to port 25 (by default)
 
 ### New features
 
--- a/config/tls.ini
+++ b/config/tls.ini
@@ -48,6 +48,9 @@
 ; TLS NO-GO Inbound expiry time in seconds
 ; disable_inbound_expiry = 3600
 
+; TLS NO-GO Inbound engages on the following ports only
+; disable_inbound_on_ports[] = 25
+
 ; no_tls_hosts - disable TLS for servers with broken TLS. (applies to inbound only)
 [no_tls_hosts]
 ; 127.0.0.1
--- a/plugins/tls.js
+++ b/plugins/tls.js
@@ -42,11 +42,12 @@ exports.advertise_starttls = function (next, connection) {
         next();
     }
 
-    if (!tls_socket.cfg.redis || !server.notes.redis) {
+    const redis = server.notes.redis;
+    const nogo_ports = tls_socket.cfg.redis.disable_inbound_on_ports || [25];
+    if (!tls_socket.cfg.redis || !redis || !nogo_ports.includes(connection.local.port)) {
         return enable_tls();
     }
 
-    const redis = server.notes.redis;
     const dbkey = `no_tls|${connection.remote.ip}`;
 
     redis.get(dbkey, (err, dbr) => {


celesteking:
wtf , let me check

msimerson:
The NO-GO feature was implemented to permit email connections between MTAs to work in the face of TLS failures. I feel like it's an oversight that it was ever applied to port 587.  Clients connecting to port 587 nearly always AUTH and credentials should never be sent over the network in plain text. For obvious reasons, TLS NO-GO doesn't apply to [port 465](https://github.com/haraka/Haraka/blob/master/docs/plugins/relay.md#authentication), which we should all be moving towards anyway. I'm in favor of getting rid of the option so that Haraka automatically does The Right Thing.

celesteking:
Well, it wasn't me who coded that feature in initially :wink: Instead of hard-coding, we could just instead introduce another option `disable_inbound_only_on_ports`.  The NO-GO feature is useful, e.g. when MTAs communicate between countries with different mandated standards , e.g., russia (gost) - X (rsa), china (GM/T,SM?) - X (rsa), etc, and can't find a common language to talk in. It shouldn't be removed, I think.

msimerson:
Keep the feature, I agree it is useful. I'm suggesting that we apply it to port 25 only.

analogic:
Isn't proper way to use separate configs for submission and smtp?

msimerson:
> Isn't proper way to use separate configs for submission and smtp?

That's one way. But I'm not seeing the relevance. Haraka's default configuration should be a good configuration, and I don't see valid use cases for TLS NO-GO on port 587.

analogic:
Yep, I am not againts. We have recently same problem with our customer - pool of clients behind NAT and one of client with old soft/bad settings can effectively block others with NO-GO when TLS auth is required

celesteking:
Would suggest moving your clients over  465 or 587. There's no reason of keeping them try 25, which is usually filtered out by ISPs. And also, it's easy for the sysop to understand who's banging the wrong door.

msimerson:
@celesteking, see #2875 . If you change this PR to be like that, I'll happily approve and merge this.

msimerson:
Why have the config option?

We offer config options so that sysadmins can choose between more than one [good / reasonable / valid] options. And we require those config options to be documented so that someone making that choice can understand what behavior they are choosing.

Disabling TLS on port 587 is **never** a good choice. We should not be offering bad choices.  If someone wants to make the bad choice to enable TLS NO-GO on port 587, they can still edit a line of code to do so.

celesteking:
It disables the disablement feature on port that is not 25, by default. What's not clear? 
Config option is for cases when HK is running on port[s] other than 25 and admin wants those port[s] to be covered by tls no-go feature.

msimerson:
> when running as unpriv user without priv ports bind access

Ahah, a valid use case!  In reviewing my arguments against having a config option, my main point is that "port 587 should never have NO-GO applied."  My PR achieved that by excluding all ports except 25, which wasn't the best way to accomplish that.  What do you think about only disabling NO-GO on port 587?   (I just updated my PR to reflect this).

celesteking:
I don't see what's wrong with my approach - it's configurable, it has sane defaults, it doesn't hard-code anything. It's such a tiny change that it isn't even worth spending more time on.

msimerson:
> I don't see what's wrong with my approach - it's configurable

It's a config option that I don't think anyone will ever have a need for. Better to just have sane behavior.

PR #2877, opened 2020-12-16
Update issue templates

msimerson:
update ISSUE_TEMPLATE with newer discreet templates for Bug Report, Feature Request, and Custom


msimerson:
This will probably require more tweaking after merging.

PR #2890, opened 2021-01-06
drop node 10 support

msimerson:
- drop support for node.js 10
- use address-rfc2821 2.0.0

Checklist:


baudehlo (approved):
I assume we don't care that much about Node 11?

msimerson:
Correct, we generally follow the [LTS schedule](https://nodejs.org/en/about/releases/) and odd numbered releases drop support after just 6 months.

msimerson:
We won't actually drop node 10 support early in a release unless we happen to release again before 2021-04-30.

PR #2896, opened 2021-01-11
haraka-plugin-redis is a dev dependency

dcharbonnier:
outbound tls is using haraka-plugin-redis


msimerson (approved):
Yes indeed, its importance has evolved and not having it as a dependency had tripped up a few people.

PR #2900, opened 2021-01-13
clone the header_list

dcharbonnier:
It's too dangerous to return the header_list object without cloning it, this object should never be modified outside of this class
we could even freeze it to make sure people know what they are doing wrong

### DIFF
--- a/mailheader.js
+++ b/mailheader.js
@@ -98,7 +98,7 @@ class Header {
     }
 
     get_all (key) {
-        return [].concat(this.headers[key.toLowerCase()] || []);
+        return Object.freeze([...(this.headers[key.toLowerCase()] || [])]);
     }
 
     get_decoded (key) {
--- a/mailheader.js
+++ b/mailheader.js
@@ -98,7 +98,7 @@ class Header {
     }
 
     get_all (key) {
-        return this.headers[key.toLowerCase()] || [];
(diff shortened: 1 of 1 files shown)


mailheader.js
```diff
@@ -98,7 +98,7 @@ class Header {
     }
 
     get_all (key) {
-        return this.headers[key.toLowerCase()] || [];
+        return [].concat(this.headers[key.toLowerCase()] || []);
```
msimerson, line 101: Instead of Array.concat, how about using the es6 spread operator?

```suggestion
        return Object.freeze([...(this.headers[key.toLowerCase()] || [])]);
```

mailheader.js
```diff
@@ -163,7 +163,7 @@ class Header {
     }
 
     lines () {
-        return this.header_list;
+        return [].concat(this.header_list);
```
msimerson, line 166:
```suggestion
        return Object.freeze([...this.header_list]);
```

msimerson (changes requested):

mailheader.js
dcharbonnier: yes, but this code base is old I was not sure it would be accepted this way ;-) perfect

msimerson (approved):
LGTM

PR #2910, opened 2021-01-21
Implement forced TLS: outbound.force_tls_hosts[] in tls.ini

Juerd:
Implemented using an ini array instead of like the [no_tls_hosts] section, so
that it can be different for outbound and inbound, for future use.

Currently does not accept wildcards like "*.example.org".

Also fixes a small bug in the "delivered" notice, where a previously TLS
upgraded socket from the pool would be logged as tls=N

- New feature: forced TLS for selected hosts
- Configuration in tls.ini outbound.no_tls_hosts[], prepared for adding inbound.no_tls_hosts[] later (and both can then take from main.no_tls_hosts[] of course)
- Bug fix: tls=N in log should be tls=Y for re-used upgraded socket from pool
- Debug feature: indicate in debug log message that a re-used socket is a TLS socket
- Bug fix: invalid SMTP codes were not treated as fatal errors

Checklist:
- [X] docs updated
- [ ] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -11,7 +11,7 @@
 
 ### New features
 
-- 
+- tls: require secure and verified sockets for configured hosts/domains
 
 ### Fixes
 
--- a/docs/plugins/tls.md
+++ b/docs/plugins/tls.md
@@ -91,6 +91,20 @@ If needed, add this section to the `config/tls.ini` file and list any IP ranges 
 
 The [Node.js TLS](http://nodejs.org/api/tls.html) page has additional information about the following options.
 
+### force_tls_hosts
+
+For known good TLS hosts, it's possible to force that the outbound mailer will only connect via secure sockets. This makes Haraka use *forced TLS* instead of *opportunistic TLS*. For forced TLS, the STARTTLS upgrade must succeed with a valid certificate (overriding `rejectUnauthorized`). The list is matched both against the host (MX record or `nexthop` in `relay_dest_domains.ini`), and the domain name of the email address.
+
+Note: unlike `no_tls_hosts`, this feature is implemented as an array:
+
+```ini
+[outbound]
+force_tls_hosts[]=172.17.123.1
+force_tls_hosts[]=172.17.124.0/24
+force_tls_hosts[]=mx.example.org
+force_tls_hosts[]=example.com
+```
+
 ### ciphers
 
 A list of allowable ciphers to use. Example:
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -308,15 +308,11 @@ class HMailItem extends events.EventEmitter {
         let host = mx.exchange;
 
         self.force_tls = false;
-        const force_tls_hosts = {};  // ip_in_list doesn't take an array
-        for (const item of obtls.cfg.force_tls_hosts) {
-            force_tls_hosts[item] = 1;
-        }
-        if (net_utils.ip_in_list(force_tls_hosts, host)) {
+        if (net_utils.ip_in_list(obtls.cfg.force_tls_hosts, host)) {
             this.logdebug(`Forcing TLS for host ${host}`);
             self.force_tls = true;
         }
-        if (net_utils.ip_in_list(force_tls_hosts, self.todo.domain)) {
+        if (net_utils.ip_in_list(obtls.cfg.force_tls_hosts, self.todo.domain)) {
             this.logdebug(`Forcing TLS for domain ${self.todo.domain}`);
             self.force_tls = true;
         }
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -1535,7 +1535,7 @@ function sort_mx (mx_list) {
     return sorted;
 }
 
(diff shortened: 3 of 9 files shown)


Juerd:
There's an array, `cfg.no_tls_hosts` in `tls.cfg` `load_config()`, already defined in current Haraka master. That's inconsistent with the *object* from the `[no_tls_hosts]` section used by the inbound stuff, and does not seem to be populated because that only uses the sections `[main]` and `[outbound]`. However, I chose to implement the suggested `force_tls_hosts` as an array because that seemed to make more sense considering that in the future you may want to force incoming TLS from certain recipients in a similar fashion, but not necessarily always for the same hosts.

Juerd:
None of the failing tests seem related to this PR.

msimerson:
There's only one failing test (there are other errors, but they aren't test failures) and it is related to this PR.

````
index
✔ outbound - converts \n and \r\n line endings to \r\n
✔ outbound - log_methods added
✔ qfile - name() basic functions
✔ qfile - name() with overrides
✔ qfile - rnd_unique() is unique-ish
✔ qfile - parts() updates previous queue filenames
✔ qfile - parts() handles standard queue filenames
2021-01-21T17:33:28.761Z [INFO] [-] [core] loading tls.ini
✖ get_tls_options - gets TLS properties from tls.ini.outbound

AssertionError:
{
  key: <Buffer 4f 75 74 62 6f 75 6e 64 54 6c 73 4b 65 79 4c 6f 61 64 65 64>,
  cert: <Buffer 4f 75 74 62 6f 75 6e 64 54 6c 73 43 65 72 74 4c 6f 61 64 65 64>,
  dhparam: <Buffer 2d 2d 2d 2d 2d 42 45 47 49 4e 20 44 48 20 50 41 52 41 4d 45 54 45 52 53 2d 2d 2d 2d 2d 0a 4d 49 49 42 43 41 4b 43 41 51 45 41 35 75 30 42 67 39 67 43 ... 374 more bytes>,
  ciphers: 'ECDHE-RSA-AES256-GCM-SHA384',
  minVersion: 'TLSv1',
  rejectUnauthorized: false,
  requestCert: false,
  honorCipherOrder: false,
  redis: { disable_for_failed_hosts: false },
  no_tls_hosts: [],
  force_tls_hosts: [],
  servername: 'mail.example.com'
}
deepEqual
{
  servername: 'mail.example.com',
  key: <Buffer 4f 75 74 62 6f 75 6e 64 54 6c 73 4b 65 79 4c 6f 61 64 65 64>,
  cert: <Buffer 4f 75 74 62 6f 75 6e 64 54 6c 73 43 65 72 74 4c 6f 61 64 65 64>,
  dhparam: <Buffer 2d 2d 2d 2d 2d 42 45 47 49 4e 20 44 48 20 50 41 52 41 4d 45 54 45 52 53 2d 2d 2d 2d 2d 0a 4d 49 49 42 43 41 4b 43 41 51 45 41 35 75 30 42 67 39 67 43 ... 374 more bytes>,
  ciphers: 'ECDHE-RSA-AES256-GCM-SHA384',
  minVersion: 'TLSv1',
  rejectUnauthorized: false,
  requestCert: false,
  honorCipherOrder: false,
  redis: { disable_for_failed_hosts: false },
  no_tls_hosts: []
}
````

**Comment by Juerd** [comment]
> (there are other errors, but they aren't test failures) 

Didn't know there was a difference. Good to know!

> There's only one failing test (...) and it is related to this PR.

Ah, missed that. Will fix.

**Comment by Juerd** [comment]
So now we're back at a test failing that doesn't seem to be related to the PR. The Windows test fails on an earlytalker thing, so I'm guessing it could be a timing related race condition gone wrong.

**Review by msimerson — CHANGES_REQUESTED** [review]
This is a great PR, well done. I'd like to see one minor change:

```js
         const force_tls_hosts = {};  // ip_in_list doesn't take an array
         for (const item of obtls.cfg.force_tls_hosts) {
             force_tls_hosts[item] = 1;
         }
```

Instead of doing this for every connection, I'd rather see it moved into the config file loader so that it's only done when the config file is loaded (the config loader has caching built-in).

Also, that 1 is being used as a boolean, right?  So lets make that an actual boolean (true) instead.

Juerd:
> This is a great PR, well done.

Thank you!

> I'd like to see one minor change: (...)
> Instead of doing this for every connection, I'd rather see it moved into the config file loader so that it's only done when the config file is loaded (the config loader has caching built-in).

Good point, but:

> Also, that 1 is being used as a boolean, right? So lets make that an actual boolean (true) instead.

If I understand `ip_in_list` correctly, the value is ignored and could be `undefined` for all it cares.

https://github.com/haraka/haraka-net-utils/blob/master/index.js#L375 uses `in` in an expression, in which case it just checks whether a property of an object exists. In the next line, it uses `in` in a loop, which also supports arrays... So I guess it supports an array of IP addresses, but not an array of hostnames :)

It's probably better to fix this in `ip_in_list` instead. I'll provide a PR tomorrow, unless anyone objects before I do :)

Juerd:
Cleaner code implemented

PR #2919, opened 2021-02-04
Fixes for RFC-2045 Quoted-Printable and RFC-2047 encoded-words

### LINKED ISSUE #2918: Too much quoting in Quoted-Printable
**Describe the bug**
Both mailheader.js and mailbody.js use haraka-utils functions encode_qp and decode_qp. The problem here is that message bodies use "Quoted-Printable" encoding defined by RFC-2045, but message headers use the "Q" encoding defined by RFC-2047. These are not the same.
The encoding actually implemented by haraka-utils is closer to the "Q" encoding than true Quoted-Printable.
Also: spaces are *not* allowed in RFC-2047 encoded-words (see section 2 "Syntax of encoded-words")

**Expected behavior**
I expected that headers use RFC-2047 "Q" encoding and message bodies use RFC-2045 Quoted-Printable encoding.

**Observed behavior**
One specific observable behavior is that message bodies using quoted-printable end up encoding *every* space into =20, not just trailing spaces.

**Steps To Reproduce**
Update this test case with quoted-printable encoding to illistates the problem:

<pre><code>
diff --git a/tests/mailbody.js b/tests/mailbody.js
index e3a36191..4982de88 100644
--- a/tests/mailbody.js
+++ b/tests/mailbody.js
@@ -17,8 +17,10 @@ function _fill_body (body, quote) {
     body.parse_more(" URL*0=\"ftp://\"\n");
     body.parse_more(" URL*1=\"cs.utk.edu/pub/moore/bulk-mailer/bulk-mailer.tar\"\n");
     body.parse_more(" title*=us-ascii'en-us'This%20is%20%2A%2A%2Afun%2A%2A%2A\n");
+    body.parse_more("Content-Transfer-Encoding: quoted-printable\n");
     body.parse_more("\n");
-    body.parse_more("Some text for your testing pleasure.\n");
+    body.parse_more("Some text for your =\n");
+    body.parse_more("testing pleasure.   \n");
     body.parse_more("Yup that was some text all right.\n");
     body.parse_more("\n");
     let text = body.parse_more("--abcdef\n");
</code></pre>

**System Info:**

Haraka | Haraka.js — Version: 2.8.27
 --- | :--- 
Node | v14.15.4
OS | Linux digilicious.com 5.10.11-200.fc33.x86_64 #1 SMP Wed Jan 27 20:21:22 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
openssl | OpenSSL 1.1.2-dev  xx XXX xxxx (Library: OpenSSL 1.1.1i FIPS  8 Dec 2020)

gene-hightower:
Fixes #2918


Changes.md
```diff
@@ -19,7 +19,7 @@
 - bounce: correctly set bounce recipients #2899
 - Get local_ip from getsockname() instead of server properties #2914
 - Received header TLS section adheres more closely to RFC 8314 #2903
-
+- Improve processing of RFC-2045 Quoted-Printable and RFC-2047 encoded-words
```
msimerson, line 22:
```suggestion
- use RFC-2045 Quoted-Printable in mail bodies
- use RFC-2047 Q encoded-words in email headers
```

mailheader.js
```diff
@@ -201,7 +197,7 @@ function _decode_header (matched, encoding, lang, cte, data) {
 
     switch (cte) {
         case 'Q':
-            data = utils.decode_qp(data.replace(/_/g, ' '));
+            data = libqp.decode(data.replace(/_/g, ' '));
```
msimerson, line 200: since this is in a header, why are we using libqp.decode instead of libmime.decodeWords?

package.json
```diff
@@ -9,7 +9,7 @@
     "server",
     "email"
   ],
-  "version": "2.8.27",
+  "version": "2.8.28",
```
msimerson, line 12:
```suggestion
  "version": "2.8.27",
```

this PR won't likely drive a Haraka release.

mailheader.js
gene-hightower: The function libmime.decodeWords works on the whole encoded-word thing, like "=?iso-8859-1?q?this=20is=20some=20text?=".
This deep into the code we're only looking at the part between the last two '?' chars, like "this=20is=20some=20text", so the decoding will work (correctly) the same way as quoted-printable at this point.
I considered re-working more of the code in mailheader.js to use the functionality of libmime, but decided against doing so as I'm not aware (right now) of any actual bugs in the mailheader.js version; so just replaced the code from haraka-utils.

gene-hightower: Replacing the *encoding* of RFC-2047 'encoded-word's corrected the problem of including spaces. This was reflected in the test add_headers.add_utf8 where a header with a space was encoded as:
<pre><code>
'Bar: =?UTF-8?Q?b=C3=B8=C3=B8?= 1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890\n'
</code></pre>
Whereas RFC-2047 says (at the top of page 5) specifically:

<pre><code>
   IMPORTANT: 'encoded-word's are designed to be recognized as 'atom's
   by an RFC 822 parser.  As a consequence, unencoded white space
   characters (such as SPACE and HTAB) are FORBIDDEN within an
   'encoded-word'.  For example, the character sequence

      =?iso-8859-1?q?this is some text?=

   would be parsed as four 'atom's, rather than as a single 'atom' (by
   an RFC 822 parser) or 'encoded-word' (by a parser which understands
   'encoded-words').  The correct way to encode the string "this is some
   text" is to encode the SPACE characters as well, e.g.

      =?iso-8859-1?q?this=20is=20some=20text?=
</code></pre>

msimerson (approved):

baudehlo (approved):

PR #2924, opened 2021-02-25
Fix `Authentication-Results` header cleanup code.

celesteking:
- fix AR cleanup code so that it works as explained in the comment above
- beautify AR header output for improved human consumption

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -161,35 +161,35 @@ class Connection {
             local_port: self.local.port
         });
 
-        const has_host = self.remote.host ? `${self.remote.host} ` : '';
-        const rhost = `client ${has_host}[${self.remote.ip}]`
-
         if (!self.client.on) return;
+
+        const log_data = {ip: self.remote.ip}
+        if (self.remote.host) log_data.host = self.remote.host
 
         self.client.on('end', () => {
             if (self.state >= states.DISCONNECTING) return;
             self.remote.closed = true;
-            self.loginfo(`${rhost} half closed connection`);
+            self.loginfo('client half closed connection', log_data);
             self.fail();
         });
 
         self.client.on('close', has_error => {
             if (self.state >= states.DISCONNECTING) return;
             self.remote.closed = true;
-            self.loginfo(`${rhost} dropped connection`);
+            self.loginfo('client dropped connection', log_data);
             self.fail();
         });
 
         self.client.on('error', err => {
             if (self.state >= states.DISCONNECTING) return;
-            self.loginfo(`${rhost} connection error: ${err}`);
+            self.loginfo(`client connection error: ${err}`, log_data);
             self.fail();
         });
 
         self.client.on('timeout', () => {
             if (self.state >= states.DISCONNECTING) return;
             self.respond(421, 'timeout', () => {
-                self.fail(`${rhost} connection timed out`);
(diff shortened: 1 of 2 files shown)


connection.js
```diff
@@ -1495,7 +1495,7 @@ class Connection {
             header = header.concat(this.transaction.notes.authentication_results);
         }
         if (header.length === 1) return '';  // no results
-        return header.join('; ');
+        return header.join(";\n\t");
```
msimerson, line 1498: If you're going to fold this line, don't you need to use a FWS?

celesteking: Not sure what you're implying.

     authres-header = "Authentication-Results:" [CFWS] authserv-id
                  [ CFWS authres-version ]
                  ( no-result / 1*resinfo ) [CFWS] CRLF
     resinfo = [CFWS] ";" methodspec [ CFWS reasonspec ]
                   *( CFWS propspec )

WSP = `space (SP, ASCII value 32) and horizontal tab (HTAB,   ASCII value 9) characters`

I guess I inserted a "CRLF" before WSP, which is just "\t" in this case. And "\n" is gonna be translated later into \r\n by some header magic I don't care of. Anyway, I've done [what you've done](https://github.com/haraka/Haraka/blob/master/connection.js#L1463) right above this code: 
```js
received_header += `;\n\t${utils.date_to_str(new Date())}`
```

celesteking:
There are still a couple of issues with that AR code. As per draft that's referenced:
- > This header field SHOULD be prepended to the existing header
- > A processing agent adding this field SHOULD NOT add this field if one
   already exists, as presumably any earlier handling agents were closer
   to the origination of the message.  If it does, it MUST remove all
   existing instances of this field before adding a new one.

Instead, we are appending and we're removing AR instead of Original-AR and appending multiple Original-AR instead of having just 1. This needs some thinking. It also makes zero sense as ARC has superseded it.

connection.js
msimerson: I asked, "don't you need to use a FWS?"

The implication is that you didn't use FWS. You've added a `\n\t`, which does not include a FWS, because per the definition, a FWS uses a CRLF.

> And "\n" is gonna be translated later

Is there a reason to depend on that instead of using a CRLF here?

celesteking: CFWS           =   (1*([FWS] comment) [FWS]) / FWS
    FWS            =   ([*WSP CRLF] 1*WSP)
    WSP            =   " " / "\t"

For `header = ['x1', 'x2']`, the final result is going to be: `Authentication-Results: x1;\r\n\tx2`. Tell me what's not clear. Should I change the "\n" to "\r\n"? If so, why the code excerpt I've provided doesn't contain it?

msimerson: > Should I change the "\n" to "\r\n"? 

I would. Because it's more clear. Someone doesn't need to know about the later conversion in order to recognize that is a correct FWS.

> If so, why the code excerpt I've referenced earlier

That code was added many years ago by someone who was intimately familiar with that later conversion. I'm a fan of moving the cookie jar down to lower shelves.

celesteking: you asked, we delivered

msimerson (approved):

PR #2927, opened 2021-02-25
allow skipping SA via notes

celesteking:
-  basically, what subj says. Actually, it would've been wonderful if plugins were able to be skipped via some kind of standardized runtime configuration.

Checklist:
- [x] docs updated
- [ ] tests updated


msimerson:
> Actually, it would've been wonderful if plugins were able to be skipped via some kind of standardized runtime configuration.

So why not do that?

## Idea

```js
    // in connection.js
    connection.notes.skip-plugin = [];

    // somewhere in the connection, maybe in another plugin, maybe not
    connection.notes.skip-plugin.push('spamassassin');

    // inside the spamassassin plugin
    if (connection.notes.skip-plugin.includes('spamassassin')) return next();
```

celesteking:
oops... yeah, I like the idea because it's a very low hanging fruit to pick.

msimerson:
> What are the thoughts on moving forward with this? 

It seems that both @celesteking and I like the idea I proposed, just nobody has spent the time to implement it. Yet. If a PR showed up that implemented it, it would receive a warm welcome.

msimerson:
fixed by #2966

Issue #2933, opened 2021-03-09
TLS certificates loading error if private key is after certs in pem file


spajak:
**Describe the bug**

TLS pam certificates are not loaded from config/tls if private key and certs are concatenated in different order than expected.
This diff makes it possible to load certificate from pem files correctly no matter if the key is before the cert or the opposite.
Tested on Haraka 2.8.27 with 2 domains using testssl.sh.

**Proposed solution**

```diff
diff --git a/tls_socket.js b/tls_socket.js
index 445ac08b..b67661d2 100644
--- a/tls_socket.js
+++ b/tls_socket.js
@@ -211,18 +211,24 @@ exports.parse_x509_expire = (file, string) => {
 exports.parse_x509 = string => {
     const res = {};

-    const match = /^([^-]*)?([-]+BEGIN (?:\w+\s)?PRIVATE KEY[-]+[^-]+[-]+END (?:\w+\s)?PRIVATE KEY[-]+\n)([^]*)$/.exec(string);
-    if (!match) return res;
-
-    if (match[1] && match[1].length) {
-        log.logerror('leading garbage');
-        log.logerror(match[1]);
+    // Search for first private key.
+    const keyMatch = /^[-]+BEGIN (?:\w+\s)?PRIVATE KEY[-]+[^-]+[-]+END (?:\w+\s)?PRIVATE KEY[-]+$/m.exec(string);
+    if (keyMatch && keyMatch[0]) {
+        res.key = Buffer.from(keyMatch[0]);
+    }
+
+    // Search for all cetrificates, then concat them into one.
+    const certPattern = /^[-]+BEGIN CERTIFICATE[-]+[^-]+[-]+END CERTIFICATE[-]+$/mg;
+    let certMatch = null;
+    let certData = [];
+    while (certMatch = certPattern.exec(string)) {
+        if (certMatch && certMatch[0]) {
+            certData.push(certMatch[0]);
+        }
+    }
+    if (certData.length) {
+        res.cert = Buffer.from(certData.join('\n'));
     }
-    if (!match[2] || !match[2].length) return res;
-    res.key = Buffer.from(match[2]);
-
-    if (!match[3] || !match[3].length) return res;
-    res.cert = Buffer.from(match[3]);

     return res;
 }
```

baudehlo:
Can you turn this into a PR?

PR #2934, opened 2021-03-09
Make loading tls certificates from directory pem files more flexible

spajak:
Fixes #

#2933
TLS certificate from config/tls are not loaded correctly if private key is not in a fixed position in a PEM file.

- Make loading of TLS certificates from directory more flexible: private key can be before certs or after in a single pem file; all certs in file are merged into one.

PS config/tls directory is not read at all without this workaround in function `Server.get_smtp_server` line 397:
```js
    else {
        server.has_tls = false;
        tls_socket.getSocketOpts('*', opts => {
            Server.listeners.push(server);
            done(server);
        });
    }
```

Checklist:
- [x] docs updated
- [x] tests updated


msimerson (changes requested):
This PR breaks one of our tests and needs more work.

PR #2939, opened 2021-03-17
SA plugin: destroy socket on timeout

celesteking:
We shall destroy and forget about the socket on timeout otherwise it will wake up and send us data and we'll call next() on it in case the client hasn't disconnected yet.

Checklist:
- [ ] docs updated
- [ ] tests updated


plugins/spamassassin.js
```diff
@@ -309,8 +309,8 @@ exports.get_spamd_socket = function (next, conn, headers) {
         else {
             conn.logerror(plugin, 'timeout waiting for results');
         }
-        socket.end();
-        return next();
+        socket.destroy()
+        return next(CONT, 'timeout error');
```
msimerson, line 313:
```suggestion
       next(CONT, 'timeout');
```

msimerson: • A timeout isn't necessarily an error. 
• the return at EOF is superfluous, right?

msimerson (changes requested):

msimerson:
fixed by #2967

PR #2940, opened 2021-03-21
tls: add no_starttls_ports config variable

gramakri:
haraka will not advertise STARTTLS on these incoming/local ports.

- a new config variable no_starttls_ports where Haraka won't advertise STARTTLS

Checklist:
- [X] docs updated
- [X] tests updated


msimerson (approved):

gramakri:
use case/more context: 

We use Haraka as the mail server in Cloudron. We have a wide variety of frameworks/apps on the platform and each has it's own limitation when dealing with TLS certificates. We had been relying on `no_tls_hosts` so far to disable STARTTLS entirely for all the apps. But we are now having many apps which won't send username/password without STARTTLS even if the mail server IP is internal. These apps are usually Go apps and this is hardcoded in the Go mail library.

With this new variable, we can make haraka listen on multiple ports - say 2525, 2587 and then disable STARTTLS on 2525 alone. This way, it's not based on the source IP but on the destination port.

PR #2952, opened 2021-06-15
fix(localhost-loop): fix outbound loop when target mx is localhost

### LINKED ISSUE #2809: Localhost mail loop problem
I am trying to solve kind of configuration problem - it is not Haraka issue per se

There are multiple servers which returns MX as 0.0.0.0 or 127.0.0.1, or other "problematic" IPs. What happens is that Haraka will deliver to itself and relay via outbound again. It depends on added headers but after ±100 deliveries Haraka will reject mail as looped mail. 

Thing is that malicious user can easily induce dos with this kind of behaviour. Only solution I've come up is kind of hacky with abusing metadata in headers. Am I missing something?

DoobleD:
Fixes #2809

- keep target mx only if its IP is not a local IP

Checklist:
- [x] docs updated
- [ ] tests updated

### DIFF
--- a/docs/Outbound.md
+++ b/docs/Outbound.md
@@ -95,6 +95,12 @@ Set this to `0` to completely disable the pooling code.
 This value determines how many concurrent connections can be made to a single
 IP address (destination) in the pool. Default: 10 connections.
 
+* `local_mx_ok`
+
+Default: false. By default, outbound to a local IP is disabled, to avoid creating 
+outbound loops. Set this to true if you want to allow outbound to local IPs. 
+This is be useful if you want to deliver mail to localhost on another port.
+
 ### outbound.bounce\_message
 
 See "Bounce Messages" below for details.
--- a/docs/Outbound.md
+++ b/docs/Outbound.md
@@ -99,7 +99,7 @@ IP address (destination) in the pool. Default: 10 connections.
 
 Default: false. By default, outbound to a local IP is disabled, to avoid creating 
 outbound loops. Set this to true if you want to allow outbound to local IPs. 
-This is be useful if you want to deliver mail to localhost on another port.
+This could be useful if you want to deliver mail to localhost on another port.
 
 ### outbound.bounce\_message
 
--- a/outbound/config.js
+++ b/outbound/config.js
@@ -43,9 +43,6 @@ function load_config () {
     if (!cfg.received_header) {
         cfg.received_header = config.get('outbound.received_header') || 'Haraka outbound';
     }
-    if (!cfg.local_mx_ok) {
-        cfg.local_mx_ok = config.get('outbound.local_mx_ok') || false;
-    }
 }
 
 load_config();
--- a/outbound/config.js
+++ b/outbound/config.js
@@ -42,6 +42,9 @@ function load_config () {
     if (!cfg.received_header) {
         cfg.received_header = config.get('outbound.received_header') || 'Haraka outbound';
     }
+    if (!cfg.local_mx_ok) {
+        cfg.local_mx_ok = config.get('outbound.local_mx_ok') || false;
+    }
 }
 
 load_config();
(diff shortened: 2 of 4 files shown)


baudehlo (changes requested):
I would wrap this in some sort of config too, something like cfg.local_mx_ok (which will disable this check, i.e. outbound to local IPs is disabled by default, as per this patch), and add it to booleans in outbound/config.js, and don't forget the docs.

msimerson:
I'm not sure this is the right default. There are plenty of use cases where a MTA accepts mail and then delivers it to localhost on another port.

baudehlo:
Sure but I guess my opinion is that, that is the rare scenario for people who know what they are doing. If we add a flag to it we have to pick a default and default safe seems best.

DoobleD:
Thanks for the feedback. Here's an attempt at a config param as suggested.

outbound/config.js
```diff
@@ -42,6 +42,9 @@ function load_config () {
     if (!cfg.received_header) {
         cfg.received_header = config.get('outbound.received_header') || 'Haraka outbound';
     }
+    if (!cfg.local_mx_ok) {
```
baudehlo, line 45: I think this needs added to `booleans` at the top too.

baudehlo (changes requested):

outbound/config.js
DoobleD: Thanks, fixed. :)

docs/Outbound.md
```diff
@@ -95,6 +95,12 @@ Set this to `0` to completely disable the pooling code.
 This value determines how many concurrent connections can be made to a single
 IP address (destination) in the pool. Default: 10 connections.
 
+* `local_mx_ok`
+
+Default: false. By default, outbound to a local IP is disabled, to avoid creating 
+outbound loops. Set this to true if you want to allow outbound to local IPs. 
+This is be useful if you want to deliver mail to localhost on another port.
```
msimerson, line 102:
```suggestion
This could be useful if you want to deliver mail to localhost on another port.
```

msimerson (dismissed):
LGTM

msimerson (approved):

outbound/mx_lookup.js
```diff
@@ -33,8 +35,10 @@ exports.lookup_mx = function lookup_mx (domain, cb) {
         }
         else if (addresses && addresses.length) {
             for (let i=0,l=addresses.length; i < l; i++) {
-                const mx = wrap_mx(addresses[i]);
-                mxs.push(mx);
+                if (obc.cfg.local_mx_ok || !net_utils.is_local_ip(addresses[i].exchange)) {
```
analogic, line 38: Sorry for being late to the party but addresses[i].exchange gives me hostname not IP

DoobleD: Thanks for the feedback @analogic. 

I just tested it and you're right, I get the hostname too, whereas `net_utils.is_local_ip` seems to expect an ip. 

The change solves #2809 anyway because (from what I understand) the issue was addressing cases where the MX record was returning a local IP (not a hostname), such as with domain yeah.com:

```
$ dig MX yeah.com
...
yeah.com.		534	IN	MX	1000 0.0.0.0.
...
```

In that case `addresses = [ { exchange: '0.0.0.0', priority: 1000 } ]` and the fix does work.

The fix would fail for an MX record resolving to a local ip. Is this a legitimate use case? If yes, perhaps we could use `net_utils.get_ips_by_host` to resolve `addresses[i].exchange` to an ip? I'm wondering if this can badly impact performance though. I'm not knowledgeable enough on Haraka to evaluate this.

msimerson: use `net.isIP` to define if .exchange is an IP or not. When its not an IP, using `net_utils.get_ips_by_host` to resolve should be fine.

DoobleD: I've made two PRs as a fix attempts:
- https://github.com/haraka/Haraka/pull/3002
- https://github.com/haraka/haraka-net-utils/pull/63

PR #2958, opened 2021-08-07
add a default route for force_routing

patchlog:
Fixes #

add a default domain in relay_dest_domains to be used if the destination domain doesn't match any of the other domains

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/plugins/relay.js
+++ b/plugins/relay.js
@@ -191,8 +191,6 @@ exports.force_routing = function (next, hmail, domain) {
 
     if (!route) {
         route = plugin.dest.domains["any"];
-        plugin.logdebug(plugin,"checking wildcard");
-        plugin.logdebug(plugin,plugin.dest.domains);
         if(!route){
             plugin.logdebug(plugin, `using normal MX lookup for: ${domain}`);
             return next();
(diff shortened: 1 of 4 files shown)


plugins/relay.js
```diff
@@ -187,11 +187,14 @@ exports.force_routing = function (next, hmail, domain) {
     if (!plugin.cfg.relay.force_routing) { return next(); }
     if (!plugin.dest) { return next(); }
     if (!plugin.dest.domains) { return next(); }
-    const route = plugin.dest.domains[domain];
+    let route = plugin.dest.domains[domain];
 
     if (!route) {
-        plugin.logdebug(plugin, `using normal MX lookup for: ${domain}`);
-        return next();
+        route = plugin.dest.domains["any"];
```
msimerson, line 193:
```suggestion
        route = plugin.dest.domains.any;
```

plugins/relay.js
```diff
@@ -187,11 +187,14 @@ exports.force_routing = function (next, hmail, domain) {
...
+        if(!route){
```
msimerson, line 194:
```suggestion
        if (!route) {
```

msimerson (changes requested):
There are some lint updates that need to be made

msimerson:
tip: you can likely run `npm run lintfix` to automatically fix all the lint issues.

msimerson (approved):
LGTM

PR #2965, opened 2021-09-02
Switch to caret version ranges in NPM package manifests

### LINKED ISSUE #2960: Lock dependencies or get rid of `*` version ranges for reproducibility
Is it possible to maintain `.lock` file for `package.json` in this repository, or at least to omit [`*` and `>=` version ranges in `package.json`](https://github.com/haraka/Haraka/blob/2.8.27/package.json) if not?

It's quite bothers that builds are not reproducible and running the same command `npm install -g Haraka@2.8.27` in different points of time can result in breaking changes as some plugin may be broken due to breaking changes or regressions. This also introduces problems for downstreams (like [`instrumentisto/haraka` Docker image](https://github.com/instrumentisto/haraka-docker-image)), where minor patch bump can have subtle [unexpected consequences](https://github.com/instrumentisto/haraka-docker-image/blob/v2.8.26-node15-r1/CHANGELOG.md#2826-node15-r1--2020-12-22) for workloads using version ranges.

To simplify burden of tracking/updating/upgrading dependencies we could leverage the [dependabot](https://dependabot.com/).

tyranron:
Fixes #2960

Checklist:
- [x] ~~docs updated~~
- [x] ~~tests updated~~


msimerson (approved):

msimerson:
I think your failing test is a result of losing the pin on the nodemailer version. There's probably a config change or similar that's needed for newer versions of nodemailer to work.

tyranron:
@msimerson thanks for pointing out!

msimerson (approved):

PR #2967, opened 2021-09-10
Allow spamassassin to return DENYSOFT on errors/timeouts

superman20:
- Allow spamassassin to return DENYSOFT on connection errors, connection timeouts, and scan timeouts
- New configuration options added for each scenario, defaulting to false (honoring existing behavior)

Checklist:
- [X] docs updated
- [ ] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -11,7 +11,7 @@
 - http: use CDN for bootstrap/jquery, drop bower #2891
 - drop support for node 10  #2890
 - outbound: disable outbound to localhost by default #2952
-- spamassassin: allow returning DENYSOFT on errors
+- spamassassin: allow returning DENYSOFT on errors #2967
 
 ### New features
 
--- a/Changes.md
+++ b/Changes.md
@@ -11,6 +11,7 @@
 - http: use CDN for bootstrap/jquery, drop bower #2891
 - drop support for node 10  #2890
 - outbound: disable outbound to localhost by default #2952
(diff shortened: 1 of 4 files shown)


superman20:
Any interest in this one, yet?   It looks like it would also incidentally address #2939.

msimerson:
Sorry, I should have replied earlier. The thing I don't like is that the config option is `reject` but the action is a `deferral`. Other plugins (spf, avg) use the keyword `defer` for this purpose.

superman20:
I didn't love the naming convention I proposed either, but I was paralleling the ClamAV plugin.  I can change to the `defer` nomenclature if you'd prefer.

msimerson:
>  I was paralleling the ClamAV plugin. I can change to the defer nomenclature if you'd prefer.

In the clamav plugin, reject.virus returns a DENY (5NN), and sometimes reject.error returns a DENY and other times a DENYSOFT (4NN) error. So, the usage in clamav plugin is inconsistent. We want to avoid POLA violations so if the knob results in a deferral, then it should be named as such.

msimerson:
The other issue is the log messages like this, "Spam scan error". That's not super helpful when a DSN gets forwarded to the email admin and they're running spamassassin, rspamd, and/or another filtering engine. Better to put something explicit and unique to spamassassin in there, such as: spamd, sa, or spamassassin.

msimerson (approved):

PR #2969, opened 2021-09-16
User configurable temporary fail delay times

superman20:
- Add a new config option in outbound to allow configuration of delay times
- The configuration would look something like (see the docs for specifics):
   `temp_fail_intervals` = `1m, 5m, 9m, 15m*3, 30m*10, 1h*6, 2h*6` 

Checklist:
- [x] docs updated
- [ ] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -18,6 +18,7 @@
 - tls: add `no_starttls_ports` - an array of incoming ports where STARTTLS is not advertised
 - outbound: add local_mx_ok config #2952
 - skip plugins at runtime by pushing name into transaction.skip_plugins #2966
+- outbound: add ability to specify delay times for temporary fails in `temp_fail_intervals` #2969
 
 ### Fixes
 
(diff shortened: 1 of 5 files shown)


msimerson:
a. I like this change
b. I'd love to see some test coverage of `set_temp_fail_intervals`
c. after giving others time to look it over, I'll merge it even without b.

superman20:
I'm not opposed to testing, but not sure how it works here.  I'll have to learn the details first.  Is there any helpful information on getting started with that?  Also, I wrote the parsing code to fail over back to the original 2^N method at any sign of trouble...although I'm not saying someone couldn't input something that would totally break it.  I might need to add a limit on the array size to keep someone from consuming too much memory with a ridiculous multiplier.

msimerson:
>  Is there any helpful information on getting started with that?

Looking at the existing tests is the best documentation we have.

> I wrote the parsing code to fail over back

👍

> I might need to add a limit on the array size to keep someone from consuming too much memory with a ridiculous multiplier.

That's something test coverage could help answer.

superman20:
I added some tests.  Please let me know if that meets your expectations.

msimerson (approved):
Looks great!

PR #3002, opened 2021-11-28
prevent local delivery loop when target exchange resolves to a local hostname

### LINKED ISSUE #2809: Localhost mail loop problem

DoobleD:
Fixes #2809 also for host names (#2952 solved it for IPs only). Requires merge of [haraka-net-utils#63](https://github.com/haraka/haraka-net-utils/pull/63) first.

- new function `net_utils.is_local_host` is used instead of `net_utils.is_local_ip` to check if the exchange is localhost.

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/outbound/mx_lookup.js
+++ b/outbound/mx_lookup.js
@@ -35,7 +35,7 @@ exports.lookup_mx = function lookup_mx (domain, cb) {
         }
         else if (addresses && addresses.length) {
             for (let i=0,l=addresses.length; i < l; i++) {
-                if (obc.cfg.local_mx_ok || !net_utils.is_local_ip(addresses[i].exchange)) {
+                if (obc.cfg.local_mx_ok || !net_utils.is_local_host(addresses[i].exchange)) {
                     const mx = wrap_mx(addresses[i]);
                     mxs.push(mx);
                 }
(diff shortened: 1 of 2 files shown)


msimerson (approved):

DoobleD:
The call to `is_local_host` seems to badly impact performance for some reason. Applying the change to production made CPU jump from ~15% to 99%.

msimerson:
Yes, there's a bug in it somewhere that is calling done when done is undefined. I haven't had a chance to investigate further.

DoobleD:
> Yes, there's a bug in it somewhere that is calling done when done is undefined. I haven't had a chance to investigate further.

Thanks for the feedback @msimerson. Indeed the call to `is_local_host` in `mx_lookup.js` simply doesn't use the callback...I'll open fix PRs soon.

DoobleD:
Here are the fix PRs:

- https://github.com/haraka/Haraka/pull/3010/
- https://github.com/haraka/haraka-net-utils/pull/65

Not sure if that works for you as it uses promises.

To be sure I tested this version on a production server and it seems to be working fine so far.

PR #3009, opened 2021-12-24
2732#connection properties type-safe guarding

### LINKED ISSUE #2732: Should connection.transaction be checked at start of each hook?
On some of our servers, we have noticed sporadic crashes when accessing the `connection.transaction` in our plugins. It appears that this variable gets set to `null` when the transaction went away in the middle of an async hook.

I see code like this https://github.com/haraka/Haraka/blob/15a025f27fe2016160b376e72e51d6052ab98645/plugins/avg.js#L39 in some places where the start of the hook itself checks if the transaction is valid.

I am wondering what is the correct fix to this is:
Style 1: Check connection.transaction at start of each hook. This means there are lot of bugs like https://github.com/haraka/Haraka/blob/15a025f27fe2016160b376e72e51d6052ab98645/plugins/clamd.js#L160 , https://github.com/haraka/Haraka/blob/15a025f27fe2016160b376e72e51d6052ab98645/plugins/delay_deny.js#L21 (just some random plugins i checked)

Style 2: Expect connection.transaction to be _valid_ at start of each hook but it can become null later if the hook is async. In this case, is OK to hold on to the transaction object to avoid having to check for nullness after every async function? What i mean is:

```
let txn = connection.transaction;
asyncFunc1(function () {
    txn.notes.xxx = 'yyy'; // this is OK even if the transaction is gone
}
```

MuraraAllan:
Related to #2732 

-  connection properties type-safe guarding

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/plugins/tarpit.js
+++ b/plugins/tarpit.js
@@ -26,11 +26,14 @@ exports.tarpit = function (next, connection) {
 
     let delay = connection.notes.tarpit;
 
-    if (!delay && connection.transaction) {
+    if (delay == null && connection.transaction != null) {
         delay = connection.transaction.notes.tarpit;
     }
 
-    if (!delay) return next();
+    if (delay == null) {
+        connection.logwarn("tarpit Delay not found, skipping >>>")
+        return next();
+    }
 
     connection.loginfo(plugin, `tarpitting response for ${delay}s`);
     setTimeout(() => next(),  delay * 1000);
--- a/plugins/tarpit.js
+++ b/plugins/tarpit.js
@@ -22,12 +22,14 @@ exports.register = function () {
 }
 
 exports.tarpit = function (next, connection) {
-    const plugin = this;
+    const { transaction } = connection;
+    if (!transaction) return next();
 
+    const plugin = this;
     let delay = connection.notes.tarpit;
 
-    if (delay == null && connection.transaction != null) {
-        delay = connection.transaction.notes.tarpit;
+    if (delay == null) {
+        delay = transaction.notes.tarpit;
     }
 
     if (delay == null) {
--- a/plugins/tarpit.js
+++ b/plugins/tarpit.js
@@ -23,17 +23,18 @@ exports.register = function () {
 
 exports.tarpit = function (next, connection) {
     const { transaction } = connection;
-    if (!transaction) return next();
+    if (!transaction) {
+        return next();
+    }
 
     const plugin = this;
     let delay = connection.notes.tarpit;
 
-    if (delay == null) {
+    if (!delay) {
         delay = transaction.notes.tarpit;
     }
 
-    if (delay == null) {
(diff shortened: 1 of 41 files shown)


MuraraAllan:
@msimerson LFR

I added some logs also to clarify we entered the "null guard at x function"

I just catch and fixed some other throw points in case of connection null.
I also saw some code here and there null guarding the same way for other variables, maybe a "remove unsafe null guard and add log" task would come handy, what do you think?


I believe I forgot something on CI 😨 Can you help me with that?

plugins/tarpit.js
```diff
@@ -26,11 +26,14 @@ exports.tarpit = function (next, connection) {
 
     let delay = connection.notes.tarpit;
 
-    if (!delay && connection.transaction) {
+    if (delay == null && connection.transaction != null) {
```
msimerson, line 29: Here you are trading the imprecision of truthy and falsy with the imprecision of ==. Using == is an anti-pattern because so few people grok the difference between == and ===. Or so I thought.

msimerson:
I don't think we need the additional logging. When these conditions are hit, it's not an error or unexpected, and log entries for ended transactions tend to be confusing.

Also, which is easier to read and _looks_ better in the code?

### falsy

```js
    let results = connection.transaction?.results || connection.results;
```

### w/o falsy

```js
    let results = connection.results;
    if (connection.transaction != null) {
        results = connection.transaction.results;
    }
```

plugins/tarpit.js
MuraraAllan: @msimerson I'm 100 % with you about the trade-off. 

I disagree although about people knowing the difference, I have been working on typescript on last years, so maybe I'm biased here.

I will stick to whichever looks most comfortable for Haraka's code base patterns.

MuraraAllan:
> 
100 % I will update removing the log, and also fix for windows-16.x


I vote for the w/o falsy, cause Reading in "chunks" grabs easier to me.

We say the results are based on connection.results and when the transaction is not null, the results are actually the transaction results.

while with the assignment we do 2 actions in the same variable assignment, like "check if it is null, assign its value if not" or "return the connection .results"

I have been out for a bunch of time, so feel free to decide whichever looks more Haraka's pattern and I will stick with it \o

MuraraAllan:
@msimerson Sorry long time no see

Hey, I'm not sure if this error on  CD windows 16x happens because of GCC, build-essential.
Can you help me with that?


Fallback to falsy check
Removed logs
Type-guarded the rest of the system

MuraraAllan:
@baudehlo Hey there \o looks like sqlite3 does not work yet on 16.x. I fallback the 16.x on win test so we wait a bit for that

MuraraAllan (commented):
Looks like sqlite3 doesn't work on 16.x, I fallback ci-test-win to only 14.x, ok?

PR #3035, opened 2022-04-04
Fix attachment.js error handling with complex archive

analogic:
Fixes [#](https://github.com/haraka/haraka-plugin-attachment/issues/29) (didn't know that external attachment plugin differs with native attachment.js)

- whole unarchiveing process rewritten into multiple functions to be more intuitive
- fixed error handling flow
- more of async/await 
- latest debian's bsdtar produce different message for encrypted archives - don't know if other platforms has same bsdtar or not so I've added some tests 
- not sure if tests will work since it depends on bsdtar, please guide me if something there needs to be done

Checklist:
- [ ] docs updated
- [x] tests updated


msimerson:
running `npm run lintfix` will address all of the lint issues. I'm still looking at the PR, but I'm liking the es6 updates.

msimerson:
Looks like you can install bsdtar by adding this to the .github/workflows/ci:

```diff
➜  haraka git:(master) ✗ git diff
diff --git a/.github/workflows/ci-test.yml b/.github/workflows/ci-test.yml
index 6eaff2ef..ef4d0d29 100644
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -36,6 +36,9 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
 
+    - name: Install bsdtar
+      run: sudo apt-get install -y bsdtar
+
     - name: Install
       run: npm install
 
```

msimerson:
For windows, I don't know how to get `bsdtar` installed. If you can't figure it out, a workaround is to add a check in the test script and if you can't access a bsdtar, then skip the attachment plugin tests.

msimerson:
Hmmm, libarchive does have [CI tests for windows](https://github.com/libarchive/libarchive/blob/master/.github/workflows/ci.yml), so its a supported platform. Now to figure out how to get it installed...

analogic:
The only thing I've found is that it might be already included https://docs.microsoft.com/en-us/virtualization/community/team-blog/2017/20171219-tar-and-curl-come-to-windows

find_bsdtar_path() will need to be updated to find the path inside windows. Sadly I've no access to windows for couple years to check this

msimerson:
Well, I'm okay with you just putting in a prehook in the attachment plugin tests, and if you can't find bsdtar, then skip the tests.

superman20:
I use Windows and I can tell you the following:

- bsdtar is included by default only in Windows Server 2019 or higher and Windows 10 v1809 (October 2018 update) or higher
- bsdtar is called just "tar" on Windows
- It is located in the primary System32 folder (i.e., C:\Windows\System32).  This location would typically be in the PATH by default.
- It is known to work on older versions of Windows by copying the exe's, but there is no official way to get it on older versions, to my knowledge.

msimerson (approved):
LGTM

msimerson (changes requested):
Oops, the tests aren't running at all, because plugin.bsdtar_path isn't exported. I suggest changing [line 75](https://github.com/haraka/Haraka/pull/3035/files?w=0#diff-3aaa610f414351c4cfe36a831bcc4265a88c876d808a2e5a97ebf2afea61bc64R75) from this:

```js
            bsdtar_path = `${dir}/bsdtar`;
```

to this:

```js
            plugin.path = bsdtar_path = `${dir}/bsdtar`;
```

msimerson:
> However, I'd be more interested in figuring out why you had trouble on Server 2019 (which should in theory work without issue). I peeked (quickly) at some of your initial changes for Windows and nothing looked off. Do you recall what problems you were having?

Tests are working with windows-2019 and node 16. There were build/install failures on [Windows server 2019 and node 14](https://github.com/haraka/Haraka/runs/5765110130?check_suite_focus=true), but it might be that I already fixed it by adding a newer version of node-gyp to our dependencies.

analogic:
I've tried two blind tests but execution with absolute and relative path was failing, see:
https://github.com/haraka/Haraka/runs/5814356719?check_suite_focus=true#step:5:700
https://github.com/haraka/Haraka/runs/5814251741?check_suite_focus=true#step:5:690

(I'll do requested change tomorrow)

msimerson:
[Fresh tests](https://github.com/msimerson/Haraka/runs/5821441050?check_suite_focus=true), still failing on node 16 (14 works, contrary to what I said earlier). Even with pegged node-gyp.

superman20:
I'll investigate the Windows/Node 16 issues soon.  I've been running Windows Server 2019/Node 14 for my server.  Now that 16 is the LTS version, we definitely need to make it work.

msimerson:
windows tests conversation moved over to [here](https://github.com/haraka/Haraka/pull/3036) and deleted from this PR

msimerson (approved):

analogic:
Sorry, been out of office. Thanks Matt!

PR #3036, opened 2022-04-04
enable windows node 16 testing

msimerson:
Checklist:
- [ ] docs updated
- [x] tests updated


superman20:
My first attempt at building this with Node 16.14.2 was successful on my primary development machine.  It is Windows 10 v21H2 with all development tools installed separately (not from the windows-build-tools or similar).  This makes me wonder if the problem is environment setup related.  From the failed test here, I did notice that a file it couldn't find was in a different directory on my system.  The test is trying to find file:

`C:\Users\runneradmin\AppData\Local\node-gyp\Cache\16.14.0\common.gypi`
but on my system, it was located at
`C:\Users\runneradmin\AppData\Local\node-gyp\Cache\16.14.2\include\node\common.gypi`

I'm not sure why that is.  I'll keep investigating.  The problem is I'm not at all familiar with how the test environments are established and there may be something there that is the issue.

superman20:
OK...making some progress.  If I install Haraka with `npm install -g haraka/Haraka#master` then everything is fine.  If, however, I create my own directory and run `npm install haraka/Haraka#master` from that directory then I get exactly the errors here.  I don't yet know how to fix that.  There seems to be some directory and path confusion without the -g option.

superman20:
The problem seems to be described [here](https://github.com/nodejs/node-gyp/issues/2584).  It seems identical to my test system where I can reproduce the behavior.

superman20:
Let me know what you think about the workaround.  It doesn't seem to like the syntax in Node 14, but 16 and 17 seem OK.  I did not evaluate the actual testing errors that are now failing in Node 17.  I'm also not sure why the workaround in Node 16 is picking up node-gyp 8.4.1 instead of 9.0.0.

msimerson:
> Let me know what you think about the workaround. It doesn't seem to like the syntax in Node 14, but 16 and 17 seem OK.

We could create separate build tests for node 14 and 16, but that's icky. I think I'd rather just wait a month and see if the issue doesn't get sorted out for us.

> I did not evaluate the actual testing errors that are now failing in Node 17. I'm also not sure why the workaround in Node 16 is picking up node-gyp 8.4.1 instead of 9.0.0.

I noticed there's TLS failures on node 17.  The 8.4.1 version might be related to the npm bundled versions of node-gyp vs the installed one. Or maybe you need to specify @9.0.0 to force the newest to be installed. Or, we just try again in the future and maybe the issues will get sorted out for us.

superman20:
I'm fine to wait it out.  The `npm install -g` option still works properly in all environments and is what I use anyway.

msimerson:
I made one subtle change, based on a fresh look at the build errors on windows + node 16. I changed the node versions from '14.x' and '16.x' to just bare numbers. Either that change fixed it, or something else fixed it. Tests are now passing with both versions.

superman20:
I noticed in the latest commit that the windows node-gyp errors are back.  We may need to re-apply the workaround to address these for a while until node-gyp fixes the problem more officially.  I wonder if the Node 14 version would work with the workaround since you removed the ".X" from the version numbers??  I may open a new pull request soon to test the workaround again with the cleaner version numbers.

msimerson:
Yeah, sometimes.... In the PR I just merged, the node 16 passed in the PR build but then failed when merged. And the node 16 on windows 2022 even passed once! 🤷‍♂️

superman20:
The comments on the node-gyp site indicate that it is an intermittent problem that starts to show up more often starting in Node 16...that seems very consistent with what we are seeing here.

PR #3047, opened 2022-04-18
enable windows 2022 testing

msimerson:
Checklist:
- [ ] docs updated
- [x] tests updated


superman20:
The failure in (windows-latest, 14) seems to be an improperly setup development environment (as it's complaining that it can't find Visual Studio) and not the node-gyp problem like the other failures.  Any ideas why the dev environment is incomplete?

msimerson:
> Any ideas why the dev environment is incomplete?

Nope. Outside my ken.

PR #3052, opened 2022-04-21
Windows node-gyp workaround testing

superman20:
Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/.github/workflows/ci-test-win.yml
+++ b/.github/workflows/ci-test-win.yml
@@ -16,8 +16,8 @@ jobs:
 
     strategy:
       matrix:
-        os: [ windows-latest ]
-        node-version: [ 14 ]
+        os: [ windows-2019, windows-latest ]
+        node-version: [ 14, 16 ]
       fail-fast: false
 
     steps:
@@ -33,8 +33,8 @@ jobs:
 
     - run: npm install -g npm
 
-    - name: Install node-gyp (temp workaround)
-      run: npx node-gyp install
+    #- name: Install node-gyp (temp workaround)
+    #  run: npx node-gyp install
 
     - run: npm install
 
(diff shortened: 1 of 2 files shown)


superman20:
FYI...I still plan to try some other ideas to resolve these problems as I have time.  So don't merge into master unless some intermittent fix is desired or until I've exhausted all my ideas.  I'll let you know when I give up.  Node 14 on Windows 2022 might be a sticky problem.

superman20:
This PR now represents what I feel like is the minimum needed workarounds to make all of the combinations of Node and Windows versions pass.   There are 2 workarounds needed:

1. An npm uprade.  Node 14's default node-gyp install does not recognize Visual Studio 2022 (which comes with Windows 2022).  An npm upgrade gets a version that recognizes VS 2022.  It should be harmless for the other node/windows combinations that don't technically need it, so I didn't isolate it.
2. A node-gyp tickle.  There appears to be an install timing issue for node-gyp on Windows that randomly creates some file confusion preventing a complete install.  It appears very frequently starting in Node 16.  This is discussed in #3036.

Please merge if you find these workarounds acceptable.  I will periodically test them and remove them once they no longer seem needed.

superman20:
I've also fixed the 2 Windows test fails on Node 18.  There were 2 tests creating the listening port on `localhost` but the test was connecting to `127.0.0.1`.  It appears that in Node 18 on Windows, localhost is IPv6 only, so it was only listening on `::1`.  To be consistent with the other tests, the fix was to listen on `127.0.0.1`.

Let me know if you'd rather split this fix from the node-gyp workaround fixes.

msimerson:
Thanks for your work on this issue!

PR #3059, opened 2022-05-02
Fix journal boundary corruption issue

### LINKED ISSUE #3057: Haraka pollutes the boundary marker when there is a big list of recipients on the incoming email
**Describe the bug**
When there is a big list of recipients in To/Cc/Bcc, the immediate mime part boundary after the recipients gets prefixed by arbitary/junk content which makes the boundary marker invalid (i.e boundary marker is not on the 1st column)

Impact of this is the immediate part cannot be parsed by standard parsers like javamail etc.,

**Expected behavior**
boundary markers should be preserved as-is and should start with 1st column
```
--_a85b44ef-2e52-427c-a8da-53d4e312e41b_
Content-Type: text/plain; charset="us-ascii"
Content-Transfer-Encoding: 7bit

```
**Observed behavior**
```
	<MicrosoftExchange329e71ec88ae4615bbc36ab6ce411--_a85b44ef-2e52-427c-a8da-53d4e312e41b_
Content-Type: text/plain; charset="us-ascii"
Content-Transfer-Encoding: 7bit

```

**Steps To Reproduce**
Send an email with a big list of recipients such that the entire recipient header size > 64k

**System Info:**

Please report your OS, Node version, and Haraka version by running this shell script on your Haraka server and replacing this section with the output.

Windows 10, Ubuntu - any version, same behavior
node v14.19.1

**Additional context**
I use the following config (smtp.ini) to allow large set of headers
max_lines=50000

sathishvenkataraman:
Fixes #3057

- suppression of chunk emit from left over buffer(contains header) which are done already.  
- There may be a better way to do this, suggest please.

Checklist:
- [ ] docs updated
- [ ] tests updated


msimerson:
Thanks for this PR! Now that you have delved into the issue and have a fix, might you be able to create a test case for it in ./tests/messagestream?

sathishvenkataraman:
> Thanks for this PR! Now that you have delved into the issue and have a fix, might you be able to create a test case for it in ./tests/messagestream?

Thanks @msimerson - Working on it..

msimerson:
> Working on it..

Need help?

sathishvenkataraman:
> > Working on it..
> 
> Need help?

Sure, that would help for sure.

msimerson:
replaced by #3068

PR #3060, opened 2022-05-13
bugfix queue/dilivery with latest generic-pool

kzub:
Fixes #

With the lates generic-pool (version 3.8.2 at package.json) doesn't work simple queue delivery
It is repeatedly creating new pool entry until JavaScript heap is over and process is terminated by low memory reason

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/outbound/client_pool.js
+++ b/outbound/client_pool.js
@@ -55,7 +55,7 @@ function get_pool (port, host, local_addr, is_unix_socket, max) {
 
         create () {
             return new Promise(function (resolve, reject) {
-                _create_socket(this.name, port, host, local_addr, is_unix_socket, (err, socket) => {
+                _create_socket(name, port, host, local_addr, is_unix_socket, (err, socket) => {
                     if (err) return reject(err)
                     resolve(socket)
                 })
@@ -111,7 +111,7 @@ exports.get_client = (port, host, local_addr, is_unix_socket, callback) => {
     }
 
     const pool = get_pool(port, host, local_addr, is_unix_socket, obc.cfg.pool_concurrency_max);
-    if (pool.waitingClientsCount() >= obc.cfg.pool_concurrency_max) {
+    if (pool.borrowed >= obc.cfg.pool_concurrency_max) {
         return callback("Too many waiting clients for pool", null);
     }
     pool.acquire().then(socket => {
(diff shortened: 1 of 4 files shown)


outbound/client_pool.js
```diff
@@ -111,7 +111,7 @@ exports.get_client = (port, host, local_addr, is_unix_socket, callback) => {
     }
 
     const pool = get_pool(port, host, local_addr, is_unix_socket, obc.cfg.pool_concurrency_max);
-    if (pool.waitingClientsCount() >= obc.cfg.pool_concurrency_max) {
+    if (pool.borrowed >= obc.cfg.pool_concurrency_max) {
```
msimerson, line 114: Are you sure this shouldn't be `pool.pending`?  See [info](https://github.com/coopernurse/node-pool) and [node-pool#279](https://github.com/coopernurse/node-pool/issues/279).

kzub: hmm, you are right in some way. This place is wrong. I dove into the code and there is some contradiction. `pool_concurrency_max` by it name states  that is  maximum parallel workloads. This feature provided by genereic-pool lib itself. When you are creating the pool
```
const opts = {
        max: max || 10,
        idleTimeoutMillis: obc.cfg.pool_timeout * 1000,
 }
const pool = generic_pool.createPool(factory, opts);
```
so later there is no need to check pending connections amount
```
if (pool.borrowed >= obc.cfg.pool_concurrency_max) {
        logger.loginfo(`[outbound] get_client 5`);
        return callback("Too many waiting clients for pool", null);
}
```
`pool.acquire()` will return the Promise that will do the job and will be fullfiled only when there will be free socket to serve.

i think this block should be removed at all or used with another config variable (pool_max_queue_length)
```
if (pool.pending >= obc.cfg.pool_max_queue_length) {
        logger.loginfo(`[outbound] get_client 5`);
        return callback("Too many waiting clients for pool", null);
}
```

and then, as you wrote above, will use `pool.pending` variable
agree?

msimerson: agreed

kzub: just did an update

msimerson (approved):

PR #3066, opened 2022-06-02
fix: try to play nice with smtp.163.com

feisuzhu:


msimerson:
```sh
echo 'VXNlcm5hbWU6' | base64 -d
Username:%
echo dXNlcm5hbWU6 | base64 -d
username:%  
```

The difference is that smtp.163.com is sending `334 username:` instead of `334 Username:`. (lowercase vs upper case letter U). I'm inclined to accept this patch, in accordance with Postel's Law, "Be liberal in what you accept, and conservative in what you send."

msimerson (approved):

PR #3095, opened 2022-09-22
hmail: output the bind_helo when attempting delivery

gramakri:
it's easier to read the logs with the hostname instead of just the IP address.

### DIFF
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -376,7 +376,7 @@ class HMailItem extends events.EventEmitter {
             host = mx.path;
         }
 
-        this.loginfo(`Attempting to deliver to: ${host}:${port}${mx.using_lmtp ? " using LMTP" : ""} (${delivery_queue.length()}) (${temp_fail_queue.length()})`);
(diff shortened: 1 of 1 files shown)


gramakri:
Message looks like this now:

```
[INFO] [10E52516-DA4F-4411-97F1-C263C70ED2CB.1.1] [outbound] Attempting to deliver to my.cloudron.xx: 45.91.168.163:25 (0) (1)
```

msimerson (approved):
LGTM

msimerson:
@gramakri, in this case, `bind_helo` is the localhost hostname that is being used as the HELO/EHLO name. Example:

````
Nov 21 01:17:46 haraka haraka[63455]: [INFO] [16144612-C8A1-483C-AC30-176B0AC71A35.1.1] [outbound] Attempting to deliver to haraka: 172.16.15.8:25 (0) (0)
````

In this case, the remote hostname is `vpopmail` and has the IP address `172.16.15.8`. The message should perhaps instead say, `Attempting to deliver from: haraka to: 172.16.15.8:25 (0) (0)`?

PR #3096, opened 2022-09-22
doc: default nodes value is 1 and not 0

gramakri:
Checklist:
- [X] docs updated

Matches the code in https://github.com/haraka/Haraka/blob/master/server.js#L48


msimerson (approved):

msimerson (approved):
LGTM

PR #3097, opened 2022-09-22
queue: add the filename in log output

gramakri:
Makes it easy to correlate logs with the actual queue file.


msimerson (approved):
I'm fine with this.

Issue #3100, opened 2022-09-26
outbound: constant reconnect to unreachable servers


gramakri:
**Describe the bug**
Haraka tries non-stop to connect to unreachable servers or servers that refuse a connection.

**Expected behavior**
Haraka should not try to connect non-stop.

**Additional context**

This issue is same as https://github.com/haraka/Haraka/issues/2694 and possibly https://github.com/haraka/Haraka/issues/2507 .

I have debugged this problem and found that the core problem is this:

* Outbound uses generic-pool module to manage socket connections per IP. When a server is unreachable (i.e MX points to an IP address which as no mail server) or the server refuses connection (maybe some IP based gray listing), then the socket creation fails at https://github.com/haraka/Haraka/blob/master/outbound/client_pool.js#L50 . This results in [`create` factory method](https://github.com/haraka/Haraka/blob/master/outbound/client_pool.js#L48) returning an error to generic-pool module.

* The issue is upstream generic-pool module has no code to make `acquire` return an error. So, https://github.com/haraka/Haraka/blob/master/outbound/client_pool.js#L110 never returns! The relevant issues upstream are https://github.com/coopernurse/node-pool/issues/175 and https://github.com/coopernurse/node-pool/issues/197 . The upstream code for `acquire` is (in essence) a while true loop calling `create` non-stop until it succeeds. There is no back off logic or anything. This causes Haraka to non-stop connect to unreachable servers.

Issue is easy to reproduce: just enable debug logs and send mail to a random IP.

gramakri:
A way to mitigate the issue is to set `acquireTimeoutMillis`. This will make `acquire` fail after `acquireTimeoutMillis` milli seconds (the default is to try forever). Note that this will busy loop for `acquireTimeoutMillis` at the minimum :-(

Not sure, what else we can do here since upstream module lacks abort feature.

msimerson:
First, great job debugging the problem. I think the bits missing from outbound is an `.on('factoryCreateError', ....`  handler. It's referenced in both of the upstream issues you cited, we have in smtp_client, be I didn't see it in a quick glance of the outbound code.

gramakri:
@msimerson the `factoryCreateError` workaround posted in the upstream issues, just deque the last item. This won't work if we have multiple items being created in parallel, no? There is a race and that code is not making sure that the work item being dequed, is the one that actually errored.

msimerson:
Yes, anything between 10 and 60 seconds seems eminently reasonable to me.

[...]

PR #3104, opened 2022-11-09
outbound: set acquireTimeoutMillis

gramakri:
...to prevent constant reconnect to unreachable servers

The generic-pool module is built on the assumption that acquire always succeeds. It is implemented as a busy loop of calling create() non-stop and there is no way to make acquire() return an error.

As a mitigation error, we make acquire() fail after 10 seconds. Note that it will still busy loop for 10 seconds. We have to fix upstream module or replace generic-pool to really fix the problem.

see #3100

Fixes #

Checklist:
- [ ] docs updated
- [ ] tests updated


msimerson (approved):
LGTM

PR #3107, opened 2022-11-17
A records matching 127.255.255.xxx are query errors, not block listings.

### LINKED ISSUE #3106: Spamhaus zen 127.255.255.255 return blocking connections
**Describe the bug**
I see evidence that when zen.spamhaus.org returns 127.255.255.xxx for an IP query, the dnsbl plugin will reject that connection.
<https://www.spamhaus.org/news/article/807/using-our-public-mirrors-check-your-return-codes-now>

This was fixed in the domain block list query, but what about the IP address query?

**Expected behavior**
A records in the 127.255.255.xxx range are *query errors* and should not be taken as a block listing.

**Observed behavior**

**Steps To Reproduce**

**System Info:**

Please report your OS, Node version, and Haraka version by running this shell script on your Haraka server and replacing this section with the output.

I noticed this from *other* systems that report running various Haraka versions.

**Additional context**

gene-hightower:
Fixes #3106 

Check for 127.255.255.x error values.

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/plugins/dns_list_base.js
+++ b/plugins/dns_list_base.js
@@ -50,6 +50,12 @@ exports.lookup = function (lookup, zone, cb) {
             return cb(err, null);  // Return a null A record
         }
 
+        // <https://www.spamhaus.org/news/article/807/using-our-public-mirrors-check-your-return-codes-now>
+        if (a && a.includes('127.255.255.')) {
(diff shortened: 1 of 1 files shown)


msimerson (approved):
LGTM

Issue #3022, opened 2022-02-12
smtp_bridge allows to use an unauthorized connection pool


kayrus:
**Describe the bug**

Prerequisites: backend postfix server with multiple users + haraka configured as a reverse proxy with smtp_bridge plugin

According to smtp_bridge docs:
> This plugins simply post the data from the original connection to the remote SMTP server using the original AUTH details.

However there is a combination of bugs:
* incorrect auth handling in `try_auth_proxy` func (similar issue as in #3020)
* sharing the authed connection pool with all SMTP clients (`get_client` func) independently on authenticated username (for instance smtp_forward uses proper `get_client_plugin` func with unique connection per user)

**Expected behavior**

* `try_auth_proxy` should correctly handle auth
* a corresponding connection pool must be available only to a particular authenticated user

**Observed behavior**

When smtp_bridge plugin is used, there is a high probability that unauthorized user with any credentials can send emails on behalf of other authed haraka users.

**Steps To Reproduce**

1) Run haraka with smtp_bridge plugin configured
2) Set `pool_concurrency_max=1` to increase the bug hit probability
3) send an email with valid username/password
4) send an email with any (valid/invalid/anything) credentials, the email will be sent on behalf of recent user

**System Info:**

Haraka | Haraka.js — Version: 2.8.28
 --- | :--- 
Node | v10.24.1
OS | Linux kay-hp 5.4.0-99-generic #112-Ubuntu SMP Thu Feb 3 13:50:55 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
openssl | OpenSSL 1.1.1f  31 Mar 2020

**Additional context**

1) smtp_bridge must use the same pool sharing concept as smtp_forward (per username)
2) there should be less duplicate logic, like auth logic in `smtp_client.js` and in `plugins/auth/auth_proxy.js`
3) this may not be all the problems

short fix at least disallows the successful attempt to send emails by unauthenticated users:
```diff
diff --git a/plugins/auth/auth_proxy.js b/plugins/auth/auth_proxy.js
--- a/plugins/auth/auth_proxy.js
+++ b/plugins/auth/auth_proxy.js
@@ -131,6 +131,7 @@ exports.try_auth_proxy = function (connection, hosts, user, passwd, cb) {
                     cert = self.config.get(self.tls_cfg.main.cert || 'tls_cert.pem', 'binary');
                     if (key && cert) {
                         this.on('secure', () => {
+                            if (secure) return;
                             secure = true;
[...]

Issue #2945, opened 2021-04-22
smtp_proxy sending data before receiving 354 go ahead


lwimble:
**Describe the bug**
We have Haraka set up using smtp_proxy which proxies all connections to Amazon SES.  This works great, except that on (seemingly random) occasion, Haraka seems to get "out of sync" in the process of brokering the connection between the client and SES.  Haraka seems to solicit data from the client by providing it with the 354 code to go head, but has not yet seen the 354 from SES.

**Expected behavior**
After sending the DATA command to SES, smtp_proxy should wait for a 354 response code before soliciting data from the client  and passing it to the ongoing server.

**Observed behavior**
smtp_proxy appears to be soliciting data from the client in form of a 354 prior to receiving the 354 from the ongoing server.

Here is the send log from the client:
`
2021-04-21 11:23:43     SERVER -> CLIENT: 220 mailproxy.mydomain.com ESMTP Haraka/2.8.27 ready
2021-04-21 11:23:43     CLIENT -> SERVER: EHLO eagle.mydomain.com
2021-04-21 11:23:43     SERVER -> CLIENT: 250-mailproxy.mydomain.com Hello {hostname redacted} [{ip address redacted}]Haraka is at your service.
2021-04-21 11:23:43     CLIENT -> SERVER: AUTH CRAM-MD5
2021-04-21 11:23:43     SERVER -> CLIENT: 334 {redacted}
2021-04-21 11:23:43     CLIENT -> SERVER: {redacted}
2021-04-21 11:23:43     SERVER -> CLIENT: 235 2.7.0 Authentication successful
2021-04-21 11:23:43     CLIENT -> SERVER: MAIL FROM:<noreply@mydomain.com>
2021-04-21 11:23:43     SERVER -> CLIENT: 250 sender <noreply@mydomain.com> OK
2021-04-21 11:23:43     CLIENT -> SERVER: RCPT TO:<myname@mydomain.com>
2021-04-21 11:23:43     SERVER -> CLIENT: 250 recipient <myname@mydomain.com> OK
2021-04-21 11:23:43     CLIENT -> SERVER: DATA
**2021-04-21 11:23:43     SERVER -> CLIENT: 354 go ahead, make my day**
2021-04-21 11:23:43     CLIENT -> SERVER: Date: Wed, 21 Apr 2021 04:23:43 -0700
2021-04-21 11:23:43     CLIENT -> SERVER: To: myname@mydomain.com
[...]

PR #3115, opened 2022-12-02
Jettison generic pool from outbound

### LINKED ISSUE #2717: [Outbound] Ongoing connection times out at connect_timeout instead of pool_timeout
If recipient smtp server takes time to respond on cmd (such as RCPT TO), socket timeout triggered at connect_timeout time instead of pool_timeout. 

In _create_socket fn we are setting socket timeout to connect_timeout (say 30s) but on connect we are not setting socket timeout again to pool_timeout (say 300s). Though we are setting idleTimeoutMillis.

We might need to reset socket timeout to pool_timeout on connect as below

```
    socket.once('connect', () => {
        socket.removeAllListeners('error'); // these get added after callback
        socket.removeAllListeners('timeout');
        socket.setTimeout(cfg.pool_timeout * 1000);  // ??? NEED THIS
        callback(null, socket);
    });
```

msimerson:
Related to #3100
Fixes #2717

Checklist:
- [x] docs updated
- [x] tests updated

### Background

The use of generic-pool has long been problematic. After a long period of being stuck on v2, I recently updated to v3, which has a new promises API. That update then started causing problems for some users because generic-pool started emitting new timeout errors, which weren't mentioned in the updating docs. Hours of tracing and debugging. I also discovered that a function we used was also removed entirely and again, no mention in the docs, so I also fixed that. So now our pooling code is back to working as well as it ever has, which is to say, it mostly works, but it still has a number of edge cases, a few of which are documented in #2717, #2945, #3022, and #3100.

Part of me thinks, "disable pooling code by default and document it with a THAR BE DRAGONS."  The rest of me thinks, it's code that introduces a lot of complexity while chasing after a little bit of performance, get rid of it.


gramakri:
Sorry, first the internet was down and then I got the flu last few weeks.

@msimerson I tested this now and all our CI tests pass. I will report any issues as I find them.

gramakri:
No issues so far.

msimerson:
Same here. I was able to do additional testing and deployments before the holiday week and all is looking good.

PR #3129, opened 2022-12-21
P42: inline variables into their references

unquietwiki:
Fixes #3120 #3126 

@msimerson Okay... had to nuke the previous git tree from orbit, re-clone, and re-do the inline process. But the tests are good, and so is the lint!

PR #3130, opened 2022-12-21
P42: if-else conditional streamlining

unquietwiki:
https://p42.ai/documentation/code-assist/convert-if-else-to-conditional-expression

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -307,7 +307,6 @@ class Connection {
         return this._relaying;
     }
     process_line (line) {
-        const self = this;
 
         if (this.state >= states.DISCONNECTING) {
             if (logger.would_log(logger.LOGPROTOCOL)) {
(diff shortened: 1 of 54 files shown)


unquietwiki:
Note: this requires #3129 to be used.

unquietwiki:
Trying to re-factor in the last set of changes from upstream + additional if-else work I was pursuing.

msimerson:
FYI: you can keep doing that locally and then [force] push it to this branch, and leave the PR open.

unquietwiki:
@msimerson Yeah, that was what I was doing before, but after those upstream changes, I had already diverged into a follow-up branch. What I'm building now is a series of related commits for the if-else logic, that depend on each other to work. I'm being very careful about the order of changes, so they'll resolve correctly.

PR #3120, opened 2022-12-19
Optimize and reduce potential errors

unquietwiki:
Changes proposed in this pull request:
- Safe optimizations via [P42](https://p42.ai/)
- More use of option-chaining (assisted by P42)
- Error-checking via [SonarLint](https://www.sonarsource.com/products/sonarlint/)

Checklist:
- [n/a] docs updated
- [X] tests updated
- [X] [Changes](https://github.com/haraka/Haraka/blob/master/Changes.md) updated

PR #3133, opened 2022-12-21
P42 if-else refactorings

unquietwiki:
Fixes #3120 #3130 

Despite the git confusion (as we ran into in other PRs), these changes complement each other.

- https://p42.ai/documentation/code-assist/move-last-statement-out-of-if-else
- https://p42.ai/documentation/code-assist/remove-empty-if-block
- https://p42.ai/documentation/code-assist/merge-nested-if
- https://p42.ai/documentation/code-assist/convert-if-else-to-conditional-expression

Problematic refactorings: not present in this PR...
- https://p42.ai/documentation/code-assist/introduce-early-return
- https://p42.ai/documentation/code-assist/merge-into-preceding-if-statement

### DIFF
--- a/plugins/queue/quarantine.js
+++ b/plugins/queue/quarantine.js
@@ -4,12 +4,11 @@ const fs   = require('fs');
 const path = require('path');
 
 exports.register = function () {
-    const plugin = this;
 
-    plugin.load_quarantine_ini();
+    this.load_quarantine_ini();
 
-    plugin.register_hook('queue',          'quarantine');
-    plugin.register_hook('queue_outbound', 'quarantine');
+    this.register_hook('queue',          'quarantine');
+    this.register_hook('queue_outbound', 'quarantine');
 }
 
 exports.hook_init_master = function (next, server) {
@@ -19,9 +18,8 @@ exports.hook_init_master = function (next, server) {
 }
 
 exports.load_quarantine_ini = function () {
-    const plugin = this;
-    plugin.cfg = plugin.config.get('quarantine.ini', () => {
-        plugin.load_quarantine_ini();
+    this.cfg = this.config.get('quarantine.ini', () => {
+        this.load_quarantine_ini();
     })
 }
 
@@ -34,15 +32,11 @@ const zeroPad = exports.zeroPad = (n, digits) => {
 }
 
 exports.clean_tmp_directory = function (next) {
-    // At start-up; delete any files in the temporary directory
-    // NOTE: This is deliberately syncronous to ensure that this
-    //       is completed prior to any messages being received.
-    const plugin = this;
-    const tmp_dir = path.join(plugin.get_base_dir(), 'tmp');
+    const tmp_dir = path.join(this.get_base_dir(), 'tmp');
 
     if (fs.existsSync(tmp_dir)) {
         const dirent = fs.readdirSync(tmp_dir);
-        plugin.loginfo(`Removing temporary files from: ${tmp_dir}`);
+        this.loginfo(`Removing temporary files from: ${tmp_dir}`);
         for (const element of dirent) {
             fs.unlinkSync(path.join(tmp_dir, element));
         }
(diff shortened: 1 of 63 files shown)


unquietwiki:
@msimerson, I slipped https://p42.ai/documentation/code-assist/invert-condition in there too. Calling it a night!

msimerson (changes requested):
Now we're venturing into aesthetics. I disagree that changes like this are an improvement. Vertical alignment of similar code makes it much easier for humans to read and pick out the differences. Making it harder to read should only be done when there's something gained. What is the gain for changes like this?

image

unquietwiki:
@msimerson Good morning! I guess for this, when I've worked on C# code, ternary use comes up quite often: I have a friend who hates ternary if that's also your thought. One thing I like to do to make it more readable is to separate the longer & more complex ones into a vertical setup of tests (especially things like || statements); that's something I can go back and look at the changes to do. I can also hit up the P42.AI guy; I still have his email from a while back, and ask about spitting out cleaner transforms.

msimerson:
I don't dislike ternaries, I just don't think they're always the correct tool. IMO, the most important priority is making it easy for human readers to understand what the code is doing. Legibility matters tremendously for that. Secondary issues are making it easier to extend and maintain code without also reducing legibility and adding complication. Simplicity is often better.

unquietwiki:
@msimerson salient points. Well, I'll take a look as I said, and I want your thoughts on how we can flag / look at those errors that came up from two of the re-factoring tools; I think that may be what broke my original PR.

baudehlo:
I'm with MattS on this one. That particular refactor resulted in worse
readability. Ternary operators should be used for simple expressions only
IMHO (watch someone find one in Haraka I wrote that violates this now).

unquietwiki:
@baudehlo Maybe the needle-thread here is to avoid https://p42.ai/documentation/code-assist/convert-if-else-to-conditional-expression without doing some polishing for readability.

unquietwiki:
@msimerson @baudehlo I tried to undo the conditional formatting; it looks like there was still some ternary already in use, that it re-factored with the if-else merges. I reached out to Lars @ P42 to look at this PR and get his thoughts, since there are multiple concerns here with the output. Thank you for your patience on this.

msimerson:
> Ternary operators should be used for simple expressions only

Agreed. I was going to say something like, "ternaries are good for short expressions." But I frequently use ternaries when the comparisons are similar logically and similar _looking_. For example:

```js
const res = some.long.thingy.param === 'Y'      ? 'yes'
          : some.long.thingy.param === 'y'      ? 'yes'
          : some.long.thingy.other === 'OH YES' ? 'yes'
          : 'no'
```

The keys to making that easily readable are the vertical alignment of like things, so the differences stand out. It's very easy to read it quickly and understand exactly what's going to happen.

rfc1869.js
```diff
@@ -25,12 +25,7 @@ const chew_regexp = /\s+([A-Za-z0-9][A-Za-z0-9-]*(?:=[^= \x00-\x1f]+)?)$/;
 exports.parse = (type, line, strict) => {
     let params = [];
     line = (new String(line)).replace(/\s*$/, '');
-    if (type === 'mail') {
-        line = line.replace(strict ? /from:/i : /from:\s*/i, '');
-    }
-    else {
-        line = line.replace(strict ? /to:/i : /to:\s*/i, '');
-    }
+    line = type === 'mail' ? line.replace(strict ? /from:/i : /from:\s*/i, '') : line.replace(strict ? /to:/i : /to:\s*/i, '');
```
msimerson, line 28: This one isn't easier to read, please revert.

unquietwiki:
@msimerson @baudehlo I'll nuke this PR and wait for Lars to chime in on this stuff, while you get the additional refactorings in. If I defer that to you all going forward, one thing I can tackle is whatever SonarLint is flagging as problematic; that shouldn't collide as much either.

plugins/queue/quarantine.js
```diff
@@ -80,12 +80,7 @@ exports.quarantine = function (next, connection) {
     const yyyymmdd = d.getFullYear() + zeroPad(d.getMonth()+1, 2)
         + this.zeroPad(d.getDate(), 2);
 
-    let subdir = yyyymmdd;
-    // Allow either boolean or a sub-directory to be specified
-
-    if (typeof quarantine !== 'boolean' && quarantine !== 1) {
-        subdir = path.join(quarantine, yyyymmdd);
-    }
+    const subdir = typeof quarantine !== 'boolean' && quarantine !== 1 ? path.join(quarantine, yyyymmdd) : yyyymmdd;
```
msimerson, line 83: revert this one too. It's long and has too many operators: assignment, comparison, logical, and alternating.

unquietwiki:
@msimerson @baudehlo I heard back from Lars: he agrees with what you said here, but he's been using [Prettier](https://prettier.io/) to enforce that. I asked him to poke further at the two rules that actually break code here.

PR #3135, opened 2022-12-21
SonarLint pass against detected bugs and code smell issues

unquietwiki:
Fixes #3120 

[SonarLint fixes](https://rules.sonarsource.com/javascript) outside of "tests" (didn't wanna mess with those for now).

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -1184,19 +1184,14 @@ class Connection {
 
         // Validate source/destination IP
         /*eslint no-fallthrough: 0 */
-        switch (proto) {
-            case 'TCP4':
-                if (ipaddr.IPv4.isValid(src_ip) && ipaddr.IPv4.isValid(dst_ip)) {
-                    break;
-                }
-            case 'TCP6':
-                if (ipaddr.IPv6.isValid(src_ip) && ipaddr.IPv6.isValid(dst_ip)) {
-                    break;
-                }
-            // case 'UNKNOWN':
-            default:
-                this.respond(421, 'Invalid PROXY format');
-                return this.disconnect();
+        // TODO: verify replacing the switch-case here worked
+        if (proto == 'TCP4' && !ipaddr.IPv4.isValid(src_ip) && !ipaddr.IPv4.isValid(dst_ip)) {
+            this.respond(421, 'Invalid IPv4 PROXY format');
+            return this.disconnect();
+        }
+        if (proto == 'TCP6' && !ipaddr.IPv6.isValid(src_ip) && !ipaddr.IPv6.isValid(dst_ip)) {
+            this.respond(421, 'Invalid IPv6 PROXY format');
+            return this.disconnect();
         }
 
         // Apply changes
--- a/connection.js
+++ b/connection.js
@@ -307,7 +307,6 @@ class Connection {
         return this._relaying;
     }
     process_line (line) {
-        const self = this;
 
         if (this.state >= states.DISCONNECTING) {
             if (logger.would_log(logger.LOGPROTOCOL)) {
@@ -365,15 +364,14 @@ class Connection {
                 }
                 catch (err) {
                     if (err.stack) {
-                        const c = this;
-                        c.logerror(`${method} failed: ${err}`);
-                        err.stack.split("\n").forEach(c.logerror);
+                        this.logerror(`${method} failed: ${err}`);
+                        err.stack.split("\n").forEach(this.logerror);
                     }
                     else {
                         this.logerror(`${method} failed: ${err}`);
                     }
                     this.respond(421, "Internal Server Error", () => {
-                        self.disconnect();
+                        this.disconnect();
                     });
                 }
             }
@@ -416,7 +414,6 @@ class Connection {
         this._process_data();
     }
     _process_data () {
-        const self = this;
         // We *must* detect disconnected connections here as the state
         // only transitions to states.CMD in the respond function below.
         // Otherwise if multiple commands are pipelined and then the
@@ -454,7 +451,7 @@ class Connection {
                     this.logdebug('[early_talker]', { state: this.state, esmtp: this.esmtp, line: this_line });
                 }
                 this.early_talker = true;
-                setImmediate(() => { self._process_data() });
+                setImmediate(() => { this._process_data() });
                 break;
             }
             else if ((this.state === states.PAUSE || this.state === states.PAUSE_SMTP) && this.esmtp) {
@@ -494,7 +491,7 @@ class Connection {
                         this.logdebug('[early_talker]', { state: this.state, esmtp: this.esmtp, line: this_line });
                     }
                     this.early_talker = true;
-                    setImmediate(() => { self._process_data() });
+                    setImmediate(() => { this._process_data() });
                 }
                 break;
             }
@@ -511,7 +508,7 @@ class Connection {
                 this.client.pause();
                 this.current_data = null;
                 return this.respond(521, "Command line too long", () => {
-                    self.disconnect();
+                    this.disconnect();
                 });
             }
             else {
@@ -605,11 +602,10 @@ class Connection {
     }
     disconnect () {
         if (this.state >= states.DISCONNECTING) return;
-        const self = this;
-        self.state = states.DISCONNECTING;
-        self.current_data = null; // don't process any more data we have already received
+        this.state = states.DISCONNECTING;
+        this.current_data = null; // don't process any more data we have already received
         this.reset_transaction(() => {
-            plugins.run_hooks('disconnect', self);
+            plugins.run_hooks('disconnect', this);
         });
     }
     disconnect_respond () {
@@ -639,9 +635,8 @@ class Connection {
         this.client.end();
     }
     get_capabilities () {
-        const capabilities = [];
 
-        return capabilities;
+        return [];
     }
     tran_uuid () {
         this.tran_count++;
@@ -672,17 +667,16 @@ class Connection {
         this.resume();
     }
     init_transaction (cb) {
-        const self = this;
         this.reset_transaction(() => {
-            self.transaction = trans.createTransaction(self.tran_uuid(), self.cfg);
+            this.transaction = trans.createTransaction(this.tran_uuid(), this.cfg);
             // Catch any errors from the message_stream
-            self.transaction.message_stream.on('error', (err) => {
-                self.logcrit(`message_stream error: ${err.message}`);
-                self.respond('421', 'Internal Server Error', () => {
-                    self.disconnect();
+            this.transaction.message_stream.on('error', (err) => {
+                this.logcrit(`message_stream error: ${err.message}`);
+                this.respond('421', 'Internal Server Error', () => {
+                    this.disconnect();
                 });
             });
-            self.transaction.results = new ResultStore(self);
+            this.transaction.results = new ResultStore(this);
             if (cb) cb();
         });
     }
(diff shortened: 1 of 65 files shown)


msimerson:
1. This PR doesn't 'fix' anything, so remove that from that description.
2. You aren't done. There's a bunch of comments sprinkled throughout this PR

unquietwiki:
@msimerson I'll go back over it after I get some errands done. I'll also update the description since "fix" is arbitrary: it does find bugs and "code smell violations", but I guess if they don't hurt, there's not an immediate issue?

msimerson:
>  since "fix" is arbitrary

Not on GitHub. Fix has special meaning, and it links PRs and issues together and has special actions related to those linked issues and PRs.

unquietwiki:
> 

Didn't think of it that way, till you said so just now. In past workflows, I'm used to there being JIRA workflows that could track projects together; I'm not normally an Agile/Waterfall guy, so my use of those would be in project meta groups, tasks, and subtasks.

unquietwiki:
@msimerson Hey, I was working on the SonarLint TODOs, and kept running into testing failures. Turns out the DNS record for the anti-spam got deleted; there was a planned cutover a year+ ago to a new one.

https://www.abuseat.org/cutover.html

unquietwiki:
@msimerson Okay... I completed most of the TODOs I generated, and the testing is working. I started looking at the testing code, but gave up after running into some issues. The testing code's gonna need its own separate review aside from that DNSBL issue I had to resolve; otherwise, I think things are fine aside from any P42 stuff you still want to do, as well as anything Lars turns up and/or explore use of Prettier to deal with any ternary messes.

connection.js
```diff
@@ -1184,19 +1184,14 @@ class Connection {
...
-            case 'TCP6':
-                if (ipaddr.IPv6.isValid(src_ip) && ipaddr.IPv6.isValid(dst_ip)) {
-                    break;
-                }
-            // case 'UNKNOWN':
-            default:
-                this.respond(421, 'Invalid PROXY format');
-                return this.disconnect();
+        // TODO: verify replacing the switch-case here worked
+        if (proto == 'TCP4' && !ipaddr.IPv4.isValid(src_ip) && !ipaddr.IPv4.isValid(dst_ip)) {
```
msimerson, line 1188: I'm pretty sure the logic here is wrong. Here I am again harping on legibility. This code is now harder to grok than the switch/case. Please revert.

unquietwiki: @msimerson Yeah, this was my own doing after trying to resolve a switch-case confusion error by SonarLint.  I went back over it, and updated the logic for clarity. In the old function, the switch-case fallthrough test was being disabled, and then it relied on a fallthrough to deal with unknown connection events. I added comments & an unknown connection entry: I think this will meet your readability needs, and avoid that ESLint hack.

msimerson: It doesn't meet my legibility concerns. If it did, it would be obvious to the casual reader that you changed the logic here and inadvertently broke it.

plugins/spamassassin.js
```diff
@@ -224,7 +225,7 @@ exports.score_too_high = function (conn, spamd_response) {
         return "spam score exceeded threshold";
     }
 
-    return false;
+    return null;
```
msimerson, line 228: Please explain why switching this from boolean to null is an improvement.

connection.js
unquietwiki: Reverted, with comments to improve readability.

plugins/spamassassin.js
unquietwiki: @msimerson the error on these involved returning the same type. true/false is boolean, but the other returns were objects / structures.

connection.js
```diff
@@ -1182,23 +1182,23 @@ class Connection {
...
                 if (ipaddr.IPv6.isValid(src_ip) && ipaddr.IPv6.isValid(dst_ip)) {
-                    break;
+                    break; // Valid IPv6 address: break loop
                 }
-            // case 'UNKNOWN':
+            case 'UNKNOWN':
             default:
                 this.respond(421, 'Invalid PROXY format');
-                return this.disconnect();
+                return this.disconnect(); // Invalid address: return error function
```
msimerson, line 1200: This is the head of Duplicate Redundant Department calling. The words 'return',  'break', and 'valid' in these comments seems a tad redundant, eh? What additional information and/or clarity is being added?  I'd argue that the comments are actually misleading. What "loop" is being broken? Is an invalid address the only condition that would hit the default case?

server.js
```diff
@@ -27,6 +27,11 @@ const { logger }      = Server;
 // Need these here so we can run hooks
 logger.add_log_methods(Server, 'server');
 
+// TODO: There was an expressed desire to make these values configurable.
+//       Also, SonarLint said these could be top-level; maybe for re-use?
+const disconnect_timeout = 30;
+const exit_timeout = 30;
```
msimerson, line 33: Just because a variable _can_ be top level doesn't mean it _should_. One off variables tend to be best located near the code using them. Maybe you like scrolling up and down pages of code to read a function, but most of us don't. If we want to make them configurable, we'll put them into an .ini file and access them via a cfg object, not global variables. Please revert.

rfc1869.js
```diff
@@ -24,7 +24,7 @@ const chew_regexp = /\s+([A-Za-z0-9][A-Za-z0-9-]*(?:=[^= \x00-\x1f]+)?)$/;
 
 exports.parse = (type, line, strict) => {
     let params = [];
-    line = (new String(line)).replace(/\s*$/, '');
+    line = (String(line)).replace(/\s*$/, '');
```
msimerson, line 27: Why this change? I referred to the docs for String and near as I can tell, this is still the suggested syntax.

connection.js
unquietwiki: Fixing now...

logger.js
```diff
@@ -141,18 +141,19 @@ logger.log = (level, data, logobj) => {
 
     // buffer until plugins are loaded
     const emptyPluginList = !plugins || Array.isArray(plugins.plugin_list) && !plugins.plugin_list.length;
+
     if (emptyPluginList) {
         logger.deferred_logs.push(item);
-        return true;
```
msimerson, line 146: One of the options P42 often suggests is early returns. Here we have one, and SonarLint (or you) have decided against the early return. I tend to favor early returns as I think it makes code more linear and easier to read. Here, the change strikes me as completely arbitrary. Either revert, or make the case that this change makes the code _better_ in some way.

rfc1869.js
unquietwiki: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/String#return_value

I guess the Sonar folks were thinking of this particular issue of new String objects being usually unnecessary; vs using a String function for desired effects.

logger.js
unquietwiki: As I recall from this, the issue here was that all return paths were true, so this clarifies that fact.

server.js
unquietwiki: I fixed it; due apologies. Clarified that it shouldn't be moved if someone else finds it.

connection.js
```diff
@@ -1182,23 +1182,23 @@ class Connection {
...
         switch (proto) {
             case 'TCP4':
                 if (ipaddr.IPv4.isValid(src_ip) && ipaddr.IPv4.isValid(dst_ip)) {
-                    break;
+                    break; // Valid IPv4 address
                 }
             case 'TCP6':
                 if (ipaddr.IPv6.isValid(src_ip) && ipaddr.IPv6.isValid(dst_ip)) {
-                    break;
+                    break; // Valid IPv6 address
```
msimerson, line 1195: Seriously. `Valid` is a subset of valid ONE LINE UP. `IPv6` is a completely redundant token. How are these comments an improvement?

outbound/hmail.js
```diff
@@ -43,6 +43,8 @@ const obc = require('./config');
 
 function dummy_func () {}
 
+const CRLF = '\r\n';
```
msimerson, line 46: same here, unhoist this.

connection.js
unquietwiki: Because I don't have to do mental math on what's going on there. The whole reason I tried to change this is because in my take on readability, it looked like the traffic was getting sidelined in some weird way. It was affirming positives, and then acting on a negative; vs acting on all possible negatives.

outbound/hmail.js
unquietwiki: @msimerson I don't know if Sonar was being smart enough, but I could see potential use for that in the rest of the program as a global constant. CRLF will always be CRLF; unless there's some weird thing with CR/LF Mac-Unix-Windows fight going on.

outbound/hmail.js
```diff
@@ -205,7 +207,7 @@ class HMailItem extends events.EventEmitter {
                     // assume string
                     const matches = /^(.*?)(:(\d+))?$/.exec(mx);
                     if (!matches) {
-                        throw ("get_mx returned something that doesn't match hostname or hostname:port");
+                        throw (new Error("get_mx returned something that doesn't match hostname or hostname:port"));
```
msimerson, line 210: I agree that anything throwing an error should `throw new Error('....')`. But...we haven't always done that. Have you verified that any code that will catch this error is expecting an error object and not just a string?

unquietwiki: It passes the test suite, and I made this change aside from the other issues we were having with P42 if-else changes that I have Lars looking at.

connection.js
msimerson: But your mental math shortcut, and thus your comment, isn't RIGHT. This code validates that both the source and destination IPs are valid IPv6. IMO, because of the use of good library and function names (IPv6.isValid()), the code is self documenting. Any comments thus added should add additional clarity, and certainly not wrong impressions (address, vs addresses).

outbound/hmail.js
msimerson: Our test suite isn't complete, and is not along good enough to justify a change like this. Either do the legwork or revert this change.

plugins/spamassassin.js
msimerson: But this change fails that test as well. Switching to a null doesn't return the same type.

connection.js
unquietwiki: Ultimately, I am trying to resolve https://rules.sonarsource.com/javascript/RSPEC-128 : it is considered a blocker, and a CWE security hazard https://cwe.mitre.org/data/definitions/484 ; it's a "may be right, but won't always be right" kind of situation.

outbound/hmail.js
unquietwiki: - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw

If anything, we're not even utilizing the full power of Error here; there are newer ES20xx commands that can do error handling like you would in C#, Nim, or other languages. And I don't see much "catch" of throws, either. This still seems legit to me.

outbound/hmail.js
msimerson: If that EVER happens, the programmer using it elsewhere will hoist the variable. Or perhaps they'll do one better and move it over to `haraka-constants`.

plugins/spamassassin.js
unquietwiki: function returns

It'll even convert that last part into `return max && (score >= max) ? "spam score exceeded threshold" : null;` if so desired.

unquietwiki: Maybe return a '' string instead?

msimerson: Now we're talking. An empty string is still falsy and that would make the type consistent.

outbound/hmail.js
msimerson: You're preaching at the choir. The question was never whether or not this was legit or a good idea. The question is whether or not this BREAKS OUR CODE because the caller(s) of this function are expecting a different type of throw. If they are, more changes are required.

outbound/hmail.js
unquietwiki: Moved back for now, as requested.

plugins/spamassassin.js
unquietwiki: Done! Made sure the test was also working with it.

outbound/hmail.js
unquietwiki: @msimerson sorry I didn't see this sooner; I wasn't getting fully updated with notifications. I have everything else fixed so far; I'll dig into this in a bit. Thank you for your extreme patience.

msimerson:
For now, I've given this PR all the time I'm willing to. I am generally a fan of linters and automated code improvement tools, and reducing "detected bugs and code smells" sure sounds good, but SonarLint doesn't live up to that standard. Too many of the changes here aren't improvements and worse, several of the changes introduce bugs. This PR needs a fair bit more work and extremely careful review before it can be merged.

unquietwiki:
@msimerson That's fair, and it looks like you cherry-picked the stuff you wanted? I'm still worried about that potential CVE you & I were haggling over; otherwise, I'm fine with where things are at.

If there's any other tooling I'd suggest at this point, that might be better than SonarLint, would be [Synk](https://snyk.io/product/open-source-security-management/). I use it on my repos.

msimerson:
> I'm still worried about that potential CVE you & I were haggling over

Easy, create a PR that does just that.

unquietwiki:
@msimerson I'll take a stab on that again & try to make something we can agree on. Again, thank you for your patience on this: I plan to use this as part of a server project I'm putting together.

unquietwiki:
@msimerson I submitted #3139 with one more take on the connection.js issue. I have an explicit 5-part test on triggering the disconnect.

Issue #3023, opened 2022-02-12
smtp_proxy doesn't handle emails when backend is unavailable


kayrus:
**Describe the bug**

When backend is unavailable, haraka crashes.

**Expected behavior**

Haraka must successfully acknowledge the email and retry to deliver it later.

**Observed behavior**

Haraka crashes: https://gist.github.com/kayrus/728a7b121a72902cff0b644e869d57d7

**Steps To Reproduce**

Configure haraka to use smtp_proxy with a hostname, which is not available due to temp network issues.
Try to send an email.

**System Info:**

Haraka | Haraka.js — Version: 2.8.28
 --- | :--- 
Node | v10.24.1
OS | Linux kay-hp 5.4.0-99-generic #112-Ubuntu SMP Thu Feb 3 13:50:55 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
openssl | OpenSSL 1.1.1f  31 Mar 2020

msimerson:
I just perused through the code in smtp_proxy, and then smtp_client, and host_pool, and I don't see any DNS resolution functions in there. Which to my mind means that hostnames in the config will not work. At all. One must use IP addresses. The documentation and configs do not make this clear.

kayrus:
@msimerson

> I don't see any DNS resolution functions in there

That's weird. If you configure Haraka with a hostname, that can be resolved, it works. IMHO documenting that there should be only an IP is not a solution. Backend servers can listen TLS sockets, therefore Haraka should be able to resolve a hostname to verify the cert to establish a TLS connection.

msimerson:
Hi @kayrus , any chance you can test the changes in #3286 and see if that helps?

kayrus:
@msimerson unfortunately no, I haven't used haraka since 2y. We tried to onboard haraka as a proxy/queue solution in our project, but the amount of showstoppers I found led to our own custom solution which we're actively using right now.

PR #3139, opened 2022-12-24
Address potential CWE in connection.js

unquietwiki:
Attempts to resolve CWE issue discovered in #3135 work.

- https://cwe.mitre.org/data/definitions/484
- https://rules.sonarsource.com/javascript/RSPEC-128

**Edit:** I had a typo in this process; I meant CWE, not CVE.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,6 @@ queued for outbound delivery.
 ### Getting Help
 
 * [Join the mailing list][8] (implemented as a Haraka plugin)
-* [Join us on Slack][14]
 * [GitHub Issues][15]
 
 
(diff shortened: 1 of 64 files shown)


msimerson:
I read the links. I disagree. Strongly.  Simply because a switch/case can fall through inadvertently does not mean that any uses of switch case that purposefully use that behavior are somehow wrong or vulnerable. Fine, give them some extra scrutiny if you wish, but this is change for changes sake and I find the replacement code to be considerably less legible.

unquietwiki:
Thank you. This is my last submission for the time being; I look forward to 3.0.

unquietwiki:
Hey, I was looking at the open issues, and found #3023: could this be related at all?

unquietwiki:
I had a call with the dev of https://rubberduck.tech/ this morning. I tried running the old and new code through it: it said they were comparable in function, so that's somewhat helpful; he also thought both solutions were readable. Another friend of mine dislikes the original method here, but thought he could do better than what I tried: he was thinking of combining some of the tests, but I had tried that before without passing the readability test here.

PR #3142, opened 2023-01-08
Fixed error messages for help options

lnedry:
Fixed error messages for help options

- changed "**qlist** option requires config path" to "**qstat** option requires config path"
- changed "**qlist** option requires config path" to "**qempty** option requires config path"
- changed "**order** option requires config path" to "**test** option requires config path"


msimerson (approved):
Thanks!

PR #3179, opened 2023-04-28
fix: do not panic at zero-byte queue files #3175

### LINKED ISSUE #3175: [outbound] Error reading queue file - zero bytes
Every now and then I help with a zero-byte queue file problem, which happens under some extreme conditions - server crash, or just running out of space. Users without knowledge of the internals are unlikely to be able to solve this themselves - a better solution might be to just give a warning and skip the file. The second option is to automatically delete zero-byte files, but this is probably more problematic and dangerous.

**Describe the bug**
```
2023-04-25 17:06:10.428437500  [ERROR] [-] [outbound] Error reading queue file 1681927792741_1681927792741_0_4290_Wnuy9n_180_mail.domain.biz: zero bytes
2023-04-25 17:06:10.430866500  [CRIT] [-] [core] Error [ERR_UNHANDLED_ERROR]: Unhandled error. ('Error reading queue file 1681927792741_1681927792741_0_4290_Wnuy9n_180_mail.domain.biz: zero bytes')
2023-04-25 17:06:10.430887500  [CRIT] [-] [core]     at new NodeError (node:internal/errors:399:5)
2023-04-25 17:06:10.430889500  [CRIT] [-] [core]     at HMailItem.emit (node:events:501:17)
2023-04-25 17:06:10.430890500  [CRIT] [-] [core]     at /usr/lib/node_modules/Haraka/outbound/hmail.js:82:22
2023-04-25 17:06:10.430892500  [CRIT] [-] [core]     at FSReqCallback.oncomplete (node:fs:192:5)
2023-04-25 17:06:10.431935500  [NOTICE] [-] [core] Shutting down
```
Haraka | Haraka.js — Version: 3.0.1
 --- | :--- 
Node | v19.6.0
OS | Linux mail.poste.io 4.19.0-22-amd64 #1 SMP Debian 4.19.260-1 (2022-09-29) x86_64 GNU/Linux
openssl | OpenSSL 1.1.1n  15 Mar 2022

analogic:
Fixes #3175

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -1,6 +1,9 @@
 
 ### Unreleased
 
+#### Fixed
+
+- fix: do not panic at zero-byte queue files #3175
 
 ### [3.0.1] - 2023-01-19
 
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -79,7 +79,7 @@ class HMailItem extends events.EventEmitter {
             }
             if (stats.size === 0) {
                 this.logerror(`Error reading queue file ${this.filename}: zero bytes`);
-                this.emit('error', `Error reading queue file ${this.filename}: zero bytes`);
+                this.temp_fail(`Error reading queue file ${this.filename}: zero bytes`);
                 return
             }
 


msimerson (commented):
We can't succeed at delivering a zero byte file in the future so I'm not sure a temp fail is the right solution. Are there any cases where deleting the useless file would be wrong?

analogic:
The only scenario I can think of is to accidentally start two haraka processes fighting over each other. Since this is not impossible, it should be safe. However, history is full of cases where well-intentioned deletions have caused massive damage in circumstances the programmer did not anticipate. So I am a bit defensive here...

msimerson:
I think we should either do something catastrophic (like crash on startup) or delete the undeliverable file.

Since this was an explicit design choice, I'd like to get another @core opinion here.

analogic:
Closing, this PR is going to crash later on anyway. (I will solve the problem by fixing queue before the Haraka is started)

PR #3191, opened 2023-05-28
Update helo.checks.js

lnedry:
Resolved issue with literal_mismatch.

### DIFF
--- a/plugins/helo.checks.js
+++ b/plugins/helo.checks.js
@@ -30,7 +30,7 @@ exports.register = function () {
     this.register_hook('helo', 'init');
     this.register_hook('ehlo', 'init');
 
-    for (const c in checks) {
+    for (const c of checks) {
         if (!this.cfg.check[c]) continue; // disabled in config
         this.register_hook('helo', c);
         this.register_hook('ehlo', c);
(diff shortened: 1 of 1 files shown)


msimerson:
Can you define the issue and how this PR solves it?  Thanks.

plugins/helo.checks.js
```diff
@@ -41,6 +40,17 @@ exports.register = function () {
     this.register_hook('helo', 'emit_log');
     this.register_hook('ehlo', 'emit_log');
 
+    // IP literal that doesn't match remote IP
+    this.register_hook('helo', 'literal_mismatch');
+    this.register_hook('ehlo', 'literal_mismatch');
+
+    if (this.cfg.check.literal_mismatch === undefined) {
```
msimerson, line 47: this seems like a good use of the nullish coalescing operator (??).
```suggestion
    this.cfg.check.literal_mismatch === this.cfg.check.literal_mismatch ?? false;
```

lnedry: I've found more issues but I don't know how to cancel this pull request.
I create a new PR once I get these issues resolved.

msimerson: You can just push additional commit to your branch lnedry:patch-4 and the PR will be updated automatically.

lnedry: The first issue that I fixed was the default value for this.cfg.check.literal_mismatch. It was set to true when it should have had a value of 2.

lnedry: I found more issues. None of the hooks were getting registered.  There was a for...in loop which loops over the checks array. To get the correct values from the array, it should be a for...of loop. 

Fixing this issue gets the hooks registered, but now each hook get called twice. So now the first time a hook gets called, I check for a previously added result.

PR #3199, opened 2023-06-13
smtp_forward: tighten up queue.wants routing

msimerson:
- smtp_forward: tighten up queue.wants routing
- doc(smtp_forward): improve markdown formatting

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -3,9 +3,11 @@
 
 #### Fixed
 
-- q(forward): tighten up queue.wants handling #3199
+- feat(q_forward): add LMTP routing handling #3190
+- chore(q_forward): tighten up queue.wants handling #3190
+- fix(q_forward): correct path to next_hop
+- doc(q_forward): improve markdown formatting #3190
 - helo.checks: serveral fixes, #3191
-- q(forward): correct path to next_hop
 - parsing error leaking to SMTP communication #3176
 - rename redis command setex to setEx #3174
 
--- a/Changes.md
+++ b/Changes.md
@@ -3,6 +3,7 @@
 
 #### Fixed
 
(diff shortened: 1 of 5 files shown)


gramakri:
To add some more info, the way we configure the forwad plugin is:

```
enable_outbound=false
domain_selector=mail_from
[cloudron.addon]
enable_outbound=true
host=172.18.0.1
port=2587
enable_tls=true
auth_type=plain
auth_user=relay_user
auth_pass=relay_pass
```

With the above config, emails are not forwarded to the relay.

msimerson:
>  if enable_outbound is false. Is this intentional ?

Well yes. Sorry. The `get_mx` hook doesn't forward (in the sense of the smtp_forward plugin) at all. Instead, messages are accepted into Haraka's queue and then delivered with outbound. The `get_mx` hook _can_ then provide a more specific route to outbound. But it's really not forwarding, it's MX routing, right? And so get_mx should be controlled by outbound_enabled.

That's what I was thinking. But I missed something important: that disables the ability to have [outbound] forward MX routing disabled in the main config and enabled per-domain. And vise versa. #3204 does the needful, but it needs to have a test added so prevent future regressions. smtp_forward is used in quite a few ways (I also [ab]use it to deliver via LMTP using outbound).

gramakri:
> But I missed something important: that disables the ability to have [outbound] forward MX routing disabled in the main config and enabled per-domain

Yes, indeed, this is exactly how we were using it. I see you already reverted that part of the code, thanks! I will test and get back.

gramakri:
Can confirm your revert fixed the issue (for us).

PR #3207, opened 2023-07-02
Add missing X-Haraka-Virus header

lnedry:
Added  X-Haraka-Virus header so scoring can happen.


msimerson (approved):

msimerson:
What are you scoring the result(s) with?

PR #3237, opened 2023-10-19
Fixed incorrect use of .join() on bounce.

### LINKED ISSUE #3236: Calling .join() without parameter causes bad RegExp replacement on bounce
**Describe the bug**

I was testing bouncing and found that when multiple recipients were defined from the same domain, mysterious **&undefined;** appeared in bounce message.

**Expected behavior**

Email addresses joined with `,` like:
`Error: Some recipients failed: <test-1@example.com>, <test-2@example.com>`

**Observed behavior**

Comma replaced with "undefined":
`Error: Some recipients failed: <test-1@example.com>&undefined; <test-2@example.com>`

**Steps To Reproduce**

Configure server, enable bouncing, use default templates and send email from your email address (needed to receive the bounce) to multiple non-existing email addresses from the same existing domain.

**System Info:**

Haraka | Haraka.js — Version: 3.0.2
 --- | :--- 
Node | v20.5.1
OS | Linux test-mailer 6.1.0-10-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.38-1 (2023-07-14) x86_64 GNU/Linux
openssl | OpenSSL 3.0.9 30 May 2023 (Library: OpenSSL 3.0.9 30 May 2023)

**Additional context**

I traced this behavior down to **outbound/hmail.js** where .join() on line 1065 is used incorrectly:
https://github.com/haraka/Haraka/blob/eb5466f216bcc74bea297ac21e11863c69452ea7/outbound/hmail.js#L1065C1-L1065C1

When .join() is called without a parameter, default of `','` is used, thus creating the escape pattern of `/[&,<,>,",',\r,\n]/g`.
Correct way should be to supply .join() with empty string like `.join('')` and therefore the desired pattern `/[&<>"'\r\n]/g`.

stanek0j:
Fixes #3236

- Added empty string as a parameter of .join().

Checklist:

### DIFF
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -1062,7 +1062,7 @@ class HMailItem extends events.EventEmitter {
             "\r": '#10',
             "\n": '#13'
         };
-        const escape_pattern = new RegExp(`[${Object.keys(escaped_chars).join()}]`, 'g');
+        const escape_pattern = new RegExp(`[${Object.keys(escaped_chars).join('')}]`, 'g');
 
         bounce_msg_html_.forEach(line => {
             line = line.replace(/\{(\w+)\}/g, (i, word) => {


msimerson (approved):
LGTM

PR #3239, opened 2023-10-24
allow configuring optional queue arguments

benjamonnguyen:
Fixes #

- Allow configuring optional queue arguments in queue/rabbitmq_amqplib plugin

Checklist:
- [x] docs updated
- [x] tests updated

### DIFF
--- a/plugins/queue/rabbitmq_amqplib.js
+++ b/plugins/queue/rabbitmq_amqplib.js
@@ -26,6 +26,7 @@ exports.rabbitmq_queue = function (next, connection) {
 
 exports.init_amqp_connection = function () {
     const cfg = this.config.get("rabbitmq.ini").rabbitmq;
+    const queueArgs = this.config.get("rabbitmq.ini").queue_args;
 
     const protocol = cfg.protocol || "amqp";
     const host = cfg.host || "127.0.0.1";
@@ -58,7 +59,7 @@ exports.init_amqp_connection = function () {
                     return conn.close();
                 }
                 ch.assertQueue(queueName,
-                    {durable, autoDelete},
+                    {durable, autoDelete, arguments: queueArgs},
                     (err4, ok2) => {
                         if (err4) {
                             this.logerror(`Error asserting rabbitmq queue: ${err4}`);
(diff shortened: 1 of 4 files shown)


plugins/queue/rabbitmq_amqplib.js
```diff
@@ -26,6 +26,7 @@ exports.rabbitmq_queue = function (next, connection) {
 
 exports.init_amqp_connection = function () {
     const cfg = this.config.get("rabbitmq.ini").rabbitmq;
+    const queueArgs = this.config.get("rabbitmq.ini").queue_args;
```
msimerson, line 29: This could be shortened to:

```suggestion
    const queueArgs = cfg.queue_args;
```

or omit it entirely and reference it from cfg below.

plugins/queue/rabbitmq_amqplib.js
```diff
@@ -58,7 +59,7 @@ exports.init_amqp_connection = function () {
                     return conn.close();
                 }
                 ch.assertQueue(queueName,
-                    {durable, autoDelete},
+                    {durable, autoDelete, arguments: queueArgs},
```
msimerson, line 62:
```suggestion
                    {durable, autoDelete, arguments: cfg.queueArgs},
```

msimerson (changes requested):
Shorten up the code per the inline comments. 

Also, not a bad idea to add a commented code block in rabbitmq_amqplib.ini with the available settings. That's often the first place an admin will go looking for knobs to twiddle.

plugins/queue/rabbitmq_amqplib.js
```diff
@@ -59,7 +58,7 @@ exports.init_amqp_connection = function () {
                     return conn.close();
                 }
                 ch.assertQueue(queueName,
-                    {durable, autoDelete, arguments: queueArgs},
+                    {durable, autoDelete, arguments: this.config.get("rabbitmq.ini").queue_args},
```
msimerson, line 61:
```suggestion
                    {durable, autoDelete, arguments: cfg.queue_args},
```

benjamonnguyen:
@msimerson `const cfg` is set to the [rabbitmq] section, but I've made the statement inline as suggested and added more documentation!

msimerson (approved):
LGTM

PR #3245, opened 2023-11-03
add queue/kafka plugin

benjamonnguyen:
Fixes #

- add queue/kafka plugin

Checklist:
- [x] docs updated
- [ ] tests updated


msimerson:
Thank you for this submission. As a general rule, we no longer bundle plug-ins with Haraka and instead package them as independent NPM modules. I have been slowly unbundling plugins. This is especially true for plugins that add dependencies.

I'd suggest releasing this as `haraka-plugin-queue-kafka`. There is a plugin template that helps bootstrap the process.

Issue #3212, opened 2023-07-26
Haraka help displays registered plugins in the wrong run order


lnedry:
**Describe the bug**

Haraka help displays registered plugins in the wrong run order.

**Expected behavior**

Haraka help describes the --order option as: "Show all registered plugins and their run order"
I expect to see a list of plugins, sorted by the order in which the hooks are executed. And under each plugin, a list of their methods (this part seems to be sorted correctly).

**Observed behavior**

/etc/haraka# haraka --configs . --order | grep Hook
Hook: init_master --------------------------------------------------------------
Hook: init_child ---------------------------------------------------------------
Hook: connect_init -------------------------------------------------------------
Hook: disconnect ---------------------------------------------------------------
Hook: rcpt ---------------------------------------------------------------------
Hook: data ---------------------------------------------------------------------
Hook: log ----------------------------------------------------------------------
Hook: connect ------------------------------------------------------------------
Hook: mail ---------------------------------------------------------------------
Hook: helo ---------------------------------------------------------------------
Hook: ehlo ---------------------------------------------------------------------
Hook: data_post ----------------------------------------------------------------
Hook: lookup_rdns --------------------------------------------------------------
Hook: deny ---------------------------------------------------------------------
Hook: capabilities -------------------------------------------------------------
Hook: unrecognized_command -----------------------------------------------------
Hook: rcpt_ok ------------------------------------------------------------------
Hook: queue_ok -----------------------------------------------------------------
Hook: queue --------------------------------------------------------------------
Hook: get_mx -------------------------------------------------------------------

**System Info:**

Haraka | Haraka.js — Version: 3.0.1
 --- | :--- 
Node | v18.16.0
OS | Linux mx4 5.10.0-21-amd64 #1 SMP Debian 5.10.162-1 (2023-01-21) x86_64 GNU/Linux
openssl | OpenSSL 1.1.1n  15 Mar 2022

PR #3272, opened 2024-02-04
Fix Haraka help

lnedry:
Fixed issue #3212 - the help option for plugin execution order was sometimes displaying hooks in the wrong order.

Haraka help now displays plugin hooks in the correct order of execution.

There may be a better way to get a list of the hooks and their order of execution but it was not obvious to me. So I grabbed them from the Plugin.md file.

### DIFF
--- a/bin/haraka
+++ b/bin/haraka
@@ -460,17 +460,19 @@ else if (parsed.order) {
     plugins = require(path.join(base, "plugins"));
     plugins.load_plugins();
     console.log('');
-    const hooks = Object.keys(plugins.registered_hooks);
+    const hooks = ["init_master", "init_child", "init_http", "init_wss", "connect_init", "lookup_rdns", "connect", "capabilities", "unrecognized_command", "disconnect", "helo", "ehlo", "quit", "vrfy", "noop", "rset", "mail", "rcpt", "rcpt_ok", "data", "data_post", "max_data_exceeded", "queue", "queue_outbound", "queue_ok", "reset_transaction", "deny", "get_mx", "deferred", "bounce", "delivered", "send_email", "pre_send_trans_email"];
     for (let h = 0; h < hooks.length; h++) {
         const hook = hooks[h];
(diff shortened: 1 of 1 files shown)


msimerson:
Putting them inline is a reasonable approach, but requires modifying both the docs and the help file in the future, which is likely to get missed. Something more automated seems prudent to me:

```sh
awk '/## Available/,/## rcpt/{if(/^[^*]/) next; print $2 }' docs/Plugins.md | tr -d '\'
init_master
init_child
init_http
init_wss
connect_init
lookup_rdns
connect
capabilities
unrecognized_command
disconnect
helo
ehlo
quit
vrfy
noop
rset
mail
rcpt
rcpt_ok
data
data_post
max_data_exceeded
queue
queue_outbound
queue_ok
reset_transaction
deny
get_mx
deferred
bounce
delivered
send_email
pre_send_trans_email
```

## Annotated

```sh
awk '/^## Available/,/^## rcpt/{if(/^[^*]/) next; print $2 }' docs/Plugins.md | tr -d '\'
```

- for the lines between the markers that start with `## Available` and `## rcpt` ...
- if the lines don't start with `*`, skip them
- print the 2nd word (the first is `*`)
- use `tr` to delete the \ escape characters

lnedry:
When I said that there may be a better way to get a list of the hooks, I was referring to programmatically getting a list from within Haraka.

What are the chances that a new hook will be added in the future? If that is a possibility, I could add some code to print a warning when someone uses a hook that is not in this help list. Or throw an error.

msimerson:
> When I said that there may be a better way to get a list of the hooks, I was referring to programmatically getting a list from within Haraka.

I didn't see a straight forward way to do that. Hence, grabbing them from the docs.

> What are the chances that a new hook will be added in the future?

Probably somewhere around 100%, eventually. We *do* require new features to be documented, so if another hook was added, I'd expect it to be documented in docs/Plugins.md, same as the rest of them.

PR #3279, opened 2024-02-26
fix(outbound): allow LHLO over insecure socket if TLS is forced (#3278)

### LINKED ISSUE #3278: Enforcing TLS for LMTP not possible
**Describe the bug**

When enforcing TLS, the only commands available before a TLS connection is established are `EHLO` and `STARTTLS`. `LHLO` is not allowed. Therefore, a STARTTLS connection for LMTP can not be established if TLS is enforced for the LMTP host. [This line](https://github.com/haraka/Haraka/blob/2075b94bc4e100c54302346082b01a0e1f37c295/outbound/hmail.js#L458) is responsible for this behavior.

**Expected behavior**

TLS should also be enforceable for LMTP.

**Observed behavior**

When enforcing TLS for the configured LMTP host, the connection will always be aborted, even if TLS is possible.

**Steps To Reproduce**

- configure `queue/lmtp` plugin
- add the LMTP host to `force_tls_hosts[]` in the `[outbound]` section of `tls.ini`

**System Info:**

Haraka | Haraka.js — Version: 3.0.1
 --- | :---
Node | v21.6.2
OS | Linux example.fauringer.de 6.7.5-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 17 Feb 2024 14:02:33 +0000 x86_64 GNU/Linux
openssl | OpenSSL 1.1.1w  11 Sep 2023

(I know that this is not the newest Haraka version but the issue is also present in the latest release.)

**Additional context**

I am currently working on a fix and will open a PR soon.

felixauringer:
Fixes #3278 

- add `LHLO` to the allowed commands if TLS is enforced

Checklist:
- [x] docs updated (not affected)
- [x] tests updated (not affected)

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -1,6 +1,10 @@
 
 ### Unreleased
 
+#### Fixed
+
+- fix(outbound): allow LHLO over insecure socket if TLS is forced #3278
+
 ### [3.0.3] - 2024-02-07
 
 #### Added
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -455,7 +455,7 @@ class HMailItem extends events.EventEmitter {
                 }
                 return;
             }
-            if (self.force_tls && (cmd != 'EHLO' && cmd != 'STARTTLS') && !socket.isSecure()) {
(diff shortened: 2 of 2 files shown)


msimerson (approved):
LGTM

PR #3294, opened 2024-04-01
call cb without MX found

analogic:
I'm running automated tests between two instances in docker - so far I've been running a custom version of mx_lookup which works as intended, trying to run the original version ends up with mails in limbo. So here is a simple fix

Checklist:
- [ ] docs updated
- [ ] tests updated


msimerson (approved):

msimerson:
thanks @analogic !

PR #3300, opened 2024-04-02
Exlude non-SMTP services from local_mx_ok

analogic:
The idea of cfg.local_mx_ok is ok, but with LMTP local delivery enabled this option makes no sense.

I've come up with a solution to "filter" only destinations with port 25, as other ports are AFAIK only used in a custom defined way and therefore wanted. The original purpose of this option should be preserved as these loops only happen with SMTP. Another way could be to just check using_lmtp, but maybe there are other protocols?

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -3,6 +3,7 @@
 
 #### Changed
 
+- outbound: exlude non-SMTP services from cfg.local_mx_ok
 - outbound/mx_lookup: make it async/await
 - outbound: emit log message when ignoring local MX
 
--- a/Changes.md
+++ b/Changes.md
@@ -3,7 +3,7 @@
 
 #### Changed
 
-- outbound: exlude non-SMTP services from cfg.local_mx_ok #3300
+- outbound: include only default routes to cfg.local_mx_ok check #3300
 - auth_base: enable disabling constrain_sender at runtime #3298
 - connection: support IPv6 when setting remote.is_private #3295
 - outbound/mx_lookup: make it async/await
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -298,7 +298,7 @@ class HMailItem extends events.EventEmitter {
         const mx = this.mxlist.shift();
         const host = mx.exchange;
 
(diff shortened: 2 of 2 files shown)


msimerson:
This is an improvement, but this uses port 25 as a proxy for determining "is this message capable of being delivered locally."  Right?  Because mail routing is so flexible, even this is bound to have edge cases.

Maybe just check for `queue.wants`? If it's set, then some plugin has already determined that there is a _known_ forward path (which is **likely** local). 

I devised `queue.wants` and `queue.next_hop` to permit any type of complex routing. The idea is fairly easy to grok (I think): `queue.wants` is the name of the plugin/entity that knows how to properly deliver that email message. The `next_hop` is a URI used by that plugin to direct the delivery.

The smtp_forward and qmail-deliverable plugins each do some form of recipient validation. When they do, they may specify the `queue.wants` and the `queue.next_hop` options to control routing. Examples:

| plugin | queue.wants | queue.next_hop | example |
| ------ | ---------| ------- | --------- |
| smtp_forward | smtp_forward | as configured in smtp_forward.ini | smtp://172.16.15.8:25/
| qmail_deliverable | outbound | lmtp (for mailboxes) | lmtp://172.16.15.15:24/
| qmail_deliverable | outbound | smtp to qmail (for aliases, forwards, mailing lists) | smtp://172.16.15.8:25/

At queue time, all the queuing plugins (and outbound) check for the presence of `queue.wants`. If it is set and the value isn't for itself, that queue plugin bails out so that the correct queue plugin can process the message.

analogic:
You're right Matt, I'm a bit blindfolded by my usecase - obviously this patch won't work for smtp forward and qmail deliverable.

I think a queue.wants check would be good enough, and certainly more usable than the whole local on/off thing. But I've been investigating how to implement this, and have come up with an idea to make it simpler and achieve the original goal. What if we only apply the local block to last resort routes in hmail.js? Adding the "default: true" flag and checking those routes should be enough...

analogic:
closing => #3307

PR #3303, opened 2024-04-04
dependencies: bump versions

msimerson:
Fixes #3299

- bump version of all deps to latest
- remove eslint from devDeps (installs if/when needed via npx)
- outbound: fix error handling in mx_lookup

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -227,7 +227,9 @@ class HMailItem extends events.EventEmitter {
         }
 
         // if none of the above return codes, drop through to this...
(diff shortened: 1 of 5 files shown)


outbound/hmail.js
```diff
@@ -227,8 +227,9 @@ class HMailItem extends events.EventEmitter {
         }
 
         // if none of the above return codes, drop through to this...
-        mx_lookup.lookup_mx(this.todo.domain, this.found_mx)
```
msimerson, line 230: unfortunately, when called this way, the scope for `this` is no longer hmail. 😔

PR #3307, opened 2024-04-05
check for local_mx only when default route is used

analogic:
github somehow broke the #3300 PR - still showing me "processing updates" and refusing to accept changes, so this is a second attempt to update cfg.local_mx_ok functionality

It only checks local mx if it falls through to the default route - so any plugin's mx is excluded.

### DIFF
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -228,8 +228,10 @@ class HMailItem extends events.EventEmitter {
 
         // if none of the above return codes, drop through to this...
         mx_lookup.lookup_mx(this.todo.domain, (err, mxs) => {
-            this.found_mx(err, mxs);
-        });
+            if (mxs) mxs.forEach(m => { m.default = true; });
+
+            this.found_mx (err, mxs)
+        })
     }
 
     found_mx (err, mxs) {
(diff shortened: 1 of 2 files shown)


msimerson:
👍 I like this approach better than #3300. I'd like to consider it for another day or two, and give others a chance to weigh in.

msimerson:
The bit of code you had to change in found_mx I find to be clumsy (manually copying over every property). Instead, how about improving that code with `Object.assign`:

```diff
diff --git a/outbound/hmail.js b/outbound/hmail.js
index 68f57f72..1a817fd7 100644
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -272,15 +272,11 @@ class HMailItem extends events.EventEmitter {
                 if (mxlist[mx].path) {
                     this.mxlist.push(mxlist[mx]);
                 }
-                else if (obc.cfg.ipv6_enabled) {
-                    this.mxlist.push(
-                        { exchange: mxlist[mx].exchange, priority: mxlist[mx].priority, port: mxlist[mx].port, using_lmtp: mxlist[mx].using_lmtp, family: 'AAAA' },
-                        { exchange: mxlist[mx].exchange, priority: mxlist[mx].priority, port: mxlist[mx].port, using_lmtp: mxlist[mx].using_lmtp, family: 'A' }
-                    );
-                }
                 else {
-                    mxlist[mx].family = 'A';
-                    this.mxlist.push(mxlist[mx]);
+                    this.mxlist.push(Object.assign({}, mxlist[mx], { family: 'A' });
+                    if (obc.cfg.ipv6_enabled) {
+                        this.mxlist.push(Object.assign({}, mxlist[mx], { family: 'AAAA' });
+                    }
                 }
             }
             this.try_deliver();
```

or even better still, use the es6 spread operator:

```js
this.mxlist.push({ ...mxlist[mx], ...{ family: 'A' });
if (obc.cfg.ipv6_enabled) {
  this.mxlist.push({ ...mxlist[mx], ...{ family: 'AAAA' });
}
```

outbound/hmail.js
```diff
@@ -300,7 +302,7 @@ class HMailItem extends events.EventEmitter {
         const mx = this.mxlist.shift();
         const host = mx.exchange;
 
-        if (!obc.cfg.local_mx_ok) {
+        if (mx.default && !obc.cfg.local_mx_ok) {
```
msimerson, line 305: I'm trying hard to like the name `default` here. It's not very self documenting and requires one to go figure out how `default` got set. In this case, default means "looked up via DNS". But that's a bit long for a name. Perhaps something like `from_dns` or `via_dns`, so the `if` reads something like "if the mx was looked up in DNS and ! local_mx_ok)". That'll also avoid namespace collisions when someone goes searching to find how that property got set.

msimerson (changes requested):

outbound/hmail.js
analogic: I don't like default either, naming is hard...

msimerson (approved):

PR #3332, opened 2024-04-24
reduce severity of data after disconnect

### LINKED ISSUE #2938: log level of 'data after disconnect' message
I'd like to tone down the log level of: ` [WARN] [xx] [core] data after disconnect from xx`. This is a frequent occurrence over here because we tend to kick the bad hosts out in a forceful manner:
```
[core] S: 554 Your host [xx] is banned
[core] disconnect ip=xx rdns=NXDOMAIN helo="" relay=N ...etc
[core] data after disconnect from xx
```
I couldn't find the code that actually closes the socket, so I'm wondering whether receiving data after it's been kicked is actually an expected event and shouldn't be logged with WARN level. 
I would do the same myself if a door was abruptly shut right in front of my face.

msimerson:
fixes #2938

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -377,7 +377,7 @@ class Connection {
     }
     process_data (data) {
         if (this.state >= states.DISCONNECTING) {
-            this.logwarn(`data after disconnect from ${this.remote.ip}`);
+            this.loginfo(`data after disconnect from ${this.remote.ip}`);
             return;
         }
 


connection.js
```diff
@@ -377,7 +377,7 @@ class Connection {
     }
     process_data (data) {
         if (this.state >= states.DISCONNECTING) {
-            this.logwarn(`data after disconnect from ${this.remote.ip}`);
+            this.loginfo(`data after disconnect from ${this.remote.ip}`);
```
derekhardison-bae, line 380: LGTM

PR #3357, opened 2024-05-11
Improve JSON override doc

louis-lau:
Linked to https://github.com/haraka/haraka-config/pull/77

The doc seems to suggest you can't override log.ini in smtp.json as it would be loaded after it. Doesn't seem to be true in reality :)


msimerson (approved):

msimerson:
Thanks!

PR #3363, opened 2024-05-15
o/hmail: fix transaction notes setting bind_helo

### LINKED ISSUE #3362: Custom tx.notes.outbound_helo not working
It took me a while to find this bug. Outbound IP works fine, but HELO/EHLO does not

There is this code:
https://github.com/haraka/Haraka/blob/f4cc2a61dc3b55b919451a010d5b2fee063e9728/outbound/hmail.js#L296-L303

But as bind_helo is already set here
https://github.com/haraka/haraka-net-utils/blob/577087b08c770919c2477378162f9971ab454295/lib/HarakaMx.js#L28-L32

there is no way to set up a custom EHLO.

msimerson:
fixes #3362

### DIFF
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -293,13 +293,8 @@ class HMailItem extends events.EventEmitter {
         }
 
         // Allow transaction notes to set outbound IP helo
-        if (!mx.bind_helo){
-            if (this.todo.notes.outbound_helo) {
-                mx.bind_helo = this.todo.notes.outbound_helo;
-            }
-            else {
-                mx.bind_helo = net_utils.get_primary_host_name();
-            }
+        if (this.todo.notes.outbound_helo) {
+            mx.bind_helo = this.todo.notes.outbound_helo;
         }
 
         const host = mx.path ? mx.path : mx.exchange;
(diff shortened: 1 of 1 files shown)


outbound/hmail.js
```diff
@@ -287,19 +287,14 @@ class HMailItem extends events.EventEmitter {
...
         // Allow transaction notes to set outbound IP helo
-        if (!mx.bind_helo){
-            if (this.todo.notes.outbound_helo) {
-                mx.bind_helo = this.todo.notes.outbound_helo;
-            }
-            else {
-                mx.bind_helo = net_utils.get_primary_host_name();
-            }
+        if (this.todo.notes.outbound_helo) {
+            mx.bind_helo = this.todo.notes.outbound_helo;
```
msimerson, line 297: setting it unconditionally here makes the docs wrong (they say `get_mx` has precedence).

msimerson: > The difficult question is how and why others use outbound ip, if at all

Exactly. That's why I want to put that error log in there. If someone is using it, they may/should notice the error messages and then raise an objection. Then maybe we'll also gain understanding of the use case and hopefully the reasoning behind the "don't set this manually" comment in the docs.

analogic: makes sense!

analogic (approved):

PR #3376, opened 2024-06-04
Fix DNS error crash

### LINKED ISSUE #3375: queryMx critical error
```
2024-06-03 18:34:52.767587500  [CRIT] [-] [core] Error: queryMx ESERVFAIL kamenoprumysl.cz
2024-06-03 18:34:52.767597500  [CRIT] [-] [core]     at QueryReqWrap.onresolve [as oncomplete] (node:internal/dns/promises:275:17)
2024-06-03 18:34:52.768066500  [NOTICE] [-] [core] Shutting down
```

Email in queue causes this error. Latest version. Will investigate later...

analogic:
First idea how to fix #3375

- slight rewrite of code around get_mx and making get_mx_respond async

Checklist:
- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -97,6 +97,7 @@
 - fix(outbound): don't send SNI servername when connecting to an IP
 - fix(outbound): chown queue dir after creation #3291
 - fix(server): async endpoint.bind() and await in server.js #3366
+- fix(outbound): get_mx DNS error handling #3376
 
 ### [3.0.3] - 2024-02-07
 
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -221,7 +221,7 @@ class HMailItem extends events.EventEmitter {
 
         // none of the above return codes, drop through to DNS
         try {
-            let exchanges = await net_utils.get_mx(this.todo.domain);
+            const exchanges = await net_utils.get_mx(this.todo.domain);
         
             if (exchanges.length) {
                 this.found_mx(this.sort_mx(exchanges))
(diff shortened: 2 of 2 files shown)


msimerson:
Looks fine. Other than changing to async/await, it should do exactly the same thing as the current code, right? Does this change actually make any difference?

analogic:
The ".catch(this.get_mx_err)" line is the problem because the callback doesn't have "this" context and therefore fails on the first line with this.lognotice(). Am I right? Even if I am not, this fixes the problem!

msimerson:
> Am I right?

Nope. `this` does indeed have context there. The problem was this:

```diff
index b9c02d80..cf7d3b17 100644
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -231,7 +231,7 @@ class HMailItem extends events.EventEmitter {
                 this.bounce(`Nowhere to deliver mail to for domain: ${this.todo.domain}`);
             }
         })
-        .catch(this.get_mx_err)
+        .catch(this.get_mx_error)
     }
 
     get_mx_error (err) {
```

Which, when fixed, yields this error:

```
[CRIT] [-] [core] TypeError: Cannot read properties of undefined (reading 'lognotice')
[CRIT] [-] [core]     at get_mx_error (/Users/matt/git/haraka/Haraka/outbound/hmail.js:238:14)
```

In order for `get_mx_err` to also have `this` context, it must be called this way:

```diff
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -231,7 +231,9 @@ class HMailItem extends events.EventEmitter {
                 this.bounce(`Nowhere to deliver mail to for domain: ${this.todo.domain}`);
             }
         })
-        .catch(this.get_mx_err)
+        .catch((err) => {
+            this.get_mx_error(err)
+        })
     }
 ```

msimerson (approved):

PR #3408, opened 2024-09-27
dep(haraka-email-message): 1.2.3 -> 1.2.4

DoobleD:
Fixes https://github.com/haraka/email-message/issues/12.

- Update haraka-email-message to 1.2.4, which correcly recognizes attachment filenames containing semicolons.

With this update, emails containing attachments named e.g. "aaa@bbb.com; ccc@ddd.com.zip" won't be rejected anymore due to the combination of the filename being incorrectly read as "aaa@bbb.com" and the [attachment](https://github.com/haraka/haraka-plugin-attachment) Haraka plugin rejecting the com extension.

Checklist:

- [ ] docs updated
- [ ] tests updated

To the maintainer: I'm not entirely sure I'm following the right contribution process here (https://github.com/haraka/email-message/pull/13 not being merged yet), feel free to discard this PR for another one if appropriate.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -3,7 +3,7 @@
 
 #### Changed
 
-- dep(haraka-email-message): 1.2.3 -> 1.2.4
+- dep(haraka-email-message): bump 1.2.3 -> 1.2.4
 - fix: revert a refactoring error in queue/lmtp, #3407
 - removed dependency on ldap plugins #3399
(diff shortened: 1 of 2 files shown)


msimerson:
Contribution process is just fine, the only nit is that since the other PR isn't merged, that version of email-message isn't yet published on NPM, so the tests here fail.

msimerson (approved):

PR #3436, opened 2025-01-24
Fixed time out issue

lnedry:
Fixes #

- removed redundant time out
- fixed issue with promises
- added several new tests

Checklist:

- [ ] docs updated
- [X] tests updated

### DIFF
--- a/config/mail_from.is_resolvable.ini
+++ b/config/mail_from.is_resolvable.ini
@@ -1,4 +1,3 @@
-timeout=29
 allow_mx_ip=false
 re_bogus_ip=^(?:0\.0\.0\.0|255\.255\.255\.255|127\.)
 
--- a/package.json
+++ b/package.json
@@ -85,7 +85,8 @@
     "haraka-test-fixtures": "^1.3.8",
     "mocha": "^11.1.0",
     "mock-require": "^3.0.3",
-    "nodemailer": "^6.9.16"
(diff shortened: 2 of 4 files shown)


msimerson:
After this change, what happens when a DNS query takes 31+ seconds to resolve? 

My memory thinks that it used to be something unpleasant, which was why that 29s timeout was added.

lnedry:
I'll work on it over the next few days.

lnedry:
I'm confident that the timeout issues have been resolved. I've also added some timeout tests.

msimerson:
Now that this PR is complete ... 😬, can I make another ask?  (I'd just do it, except that then the commits would be tagged with my name, and I'd prefer you get the credit).

I just created [https://github.com/haraka/haraka-plugin-mail_from.is_resolvable](https://github.com/haraka/haraka-plugin-mail_from.is_resolvable). You should be able to clone the repo, hop inside, and something like this should Just Work[TM]:

```sh
cd haraka-plugin-mail_from.is_resolvable
npm install --save-dev sinon
cp ~git/haraka/Haraka/plugin/mail_from.is_resolvable.js index.js
cp ~git/haraka/Haraka/test/plugins/mail_from.is_resolvable.js test/index.js
npm test
```

If the tests need something more feature full than node's built-in test runner, you can switch it out for mocha (test with `npx mocha test`).

PR #3450, opened 2025-03-17
Set connection.tls and connection.notes.tls properties

dalibor:
Is there a reason we're returning if it encounters undefined value?
It results in `peerCertificate` not being set because `verifyError` is undefined.

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -211,7 +211,6 @@ class Connection {
         this.set('hello', 'host', undefined);
         this.set('tls',   'enabled', true);
         for (const t of ['cipher','verified','verifyError','peerCertificate']) {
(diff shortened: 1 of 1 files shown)


connection.js
```diff
@@ -211,7 +211,6 @@ class Connection {
         this.set('hello', 'host', undefined);
         this.set('tls',   'enabled', true);
         for (const t of ['cipher','verified','verifyError','peerCertificate']) {
-            if (obj[t] === undefined) return;
```
msimerson, line 214: This should be `if (obj[t] === undefined) continue;`

We shouldn't be setting properties to undefined. Pretty sure this was a refactoring error and that for loop used to be in a forEach.

PR #3452, opened 2025-03-24
Fix util.isFunction deprecation warning

dalibor:
When stopping the server, I'm getting the following deprecation warning
```
^C(node:69460) [DEP0049] DeprecationWarning: The `util.isFunction` API is deprecated.  Please use `typeof arg === "function"` instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

https://nodejs.org/api/deprecations.html#DEP0049

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -32,6 +32,11 @@ function stringify (obj) {
     return str.trim();
 }
 
+function isFunction (functionToCheck) {
+    const getType = {};
+    return functionToCheck && getType.toString.call(functionToCheck) === '[object Function]';
+};
+
 const logger = exports;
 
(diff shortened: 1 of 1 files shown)


msimerson:
I'm curious...the deprecation warning is suggesting that we use `typeof arg === "function"` instead. I have vague memories from long ago of that not always working. Hence the polyfill and then the use of util.isFunction. Considering that util.isFunction is now deprecated, perhaps typeof is now smart enough to do The Right Thing in every supported version of  node.js?

msimerson (approved):
LGTM. 🤞🏻

PR #3456, opened 2025-04-01
Fix crash when outbound connection fails

### LINKED ISSUE #3455: Crash in outbound when connection fails
### Describe the bug

There is a crash in the outbound logic when socket connection fails

### Expected behavior

Should not crash

### System Info

This is Cloudron (ubuntu 24) and node v22.13.1

### Additional context

```
[CRIT] [-] [core] TypeError: Cannot read properties of null (reading 'removeAllListeners')
[CRIT] [-] [core]     at HMailItem.try_deliver_host_on_socket (/app/code/haraka/outbound/hmail.js:330:20)
[CRIT] [-] [core]     at /app/code/haraka/outbound/hmail.js:320:18
[CRIT] [-] [core]     at pluggableStream.<anonymous> (/app/code/haraka/outbound/client_pool.js:35:9)
[CRIT] [-] [core]     at Object.onceWrapper (node:events:633:26)
[CRIT] [-] [core]     at pluggableStream.emit (node:events:518:28)
[CRIT] [-] [core]     at Socket.<anonymous> (/app/code/haraka/tls_socket.js:79:18)
[CRIT] [-] [core]     at Object.onceWrapper (node:events:633:26)
[CRIT] [-] [core]     at Socket.emit (node:events:518:28)
[CRIT] [-] [core]     at emitErrorNT (node:internal/streams/destroy:170:8)
[CRIT] [-] [core]     at emitErrorCloseNT (node:internal/streams/destroy:129:3)
```

gramakri:
In node 22 atleast, network code has started returning AggregateError. This means that when socket connection fails, err.message is an empty string. The on('error') handler as a result invoked callback('', null) . Since error is falsy, code assumes socket is valid and crashes.

Code is changed to always return a error string.

Fixes #3455

- Code is changed to always return a error string on a socket error

Checklist:

- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Unreleased
 
+- fix(outbound): in outbound, fix a crash when socket connection errors #3456
+
 ### [3.1.0] - 2025-01-30
 
 #### Changes
--- a/outbound/client_pool.js
+++ b/outbound/client_pool.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const util = require('node:util');
+
 const utils = require('haraka-utils');
 const net_utils = require('haraka-net-utils')
 
@@ -32,7 +34,12 @@ exports.get_client = function (mx, callback) {
         socket.end();
         socket.removeAllListeners();
         socket.destroy();
-        callback(err.message, null);
+        const errMsg = err.message ?
+            err.message :
+            err instanceof AggregateError ?
+                err.map(e => e.message).join(', ') :
+                util.inspect(err, { depth: 3 });
+        callback(errMsg, null);
     })
 
     socket.once('timeout', () => {


msimerson:
Hmmm, this *might* be related to a new [[ buggy / suboptimal ] connection startup](https://github.com/nodejs/node/issues/54359) in node.js.

My first thought with this change is that there's a multitude of reasons why a connection might fail, and a static error message will tend to mask the specifics. Perhaps testing err with `err instanceof AggregateError` would be better, and if it's an aggregate error, return the first error message (`err.errors[0].message`)?

msimerson:
Or perhaps something more like this...

```js
const errMsg = err.message ? err.message :
	err instanceof AggregateError ? err.map(e => e.message).join(', ') :
	util.inspect(err { depth: 3 });
```

Testing this code with some of the actual errors see In The Wild[TM] would likely suggest a refinement or two.

msimerson (approved):
Looks great to me!

Issue #3460, opened 2025-04-10
Release 3.1.1


msimerson:
- [x] Start a N-week quiet period
- [x] Update [haraka-tld](https://github.com/haraka/haraka-tld), haraka/haraka-tld#88
- [x] Draft a release on the [releases page](https://github.com/haraka/Haraka/releases) (Save draft)
- Create a PR with:
    - [x] Updated [Changes](https://github.com/haraka/Haraka/blob/master/Changes.md)
    - [x] Bumped version in [package.json](https://github.com/haraka/Haraka/blob/master/package.json)
- [x] Update release date in [Changes](https://github.com/haraka/Haraka/blob/master/Changes.md)
- [x] `npm publish`
- [x] Post `Changes` file on the draft release and publish the release
- [ ] Email the mailing list
- [ ] Publish a blog post

PR #3458, opened 2025-04-10
Fix base configuration on install

fatalbanana:
Fixes #3453

Changes proposed in this pull request:

- Install `connection.ini` as part of base configuration

Checklist:

- [ ] docs updated
- [ ] tests updated
- [ ] [Changes](https://github.com/haraka/Haraka/blob/master/Changes.md) updated

Issue #3466, opened 2025-05-16
Error while setting outbound


Satej:
### Describe the bug

While trying to follow the documentation at https://haraka.github.io/tutorials/settingupoutbound, I get an issue

### Expected behavior

Haraka installation should start the server.

### Observed behavior

Error as below

haraka -c .
loaded TLD files:
  1=1443
  2=7571
  3=3661
loaded 9753 Public Suffixes
/home/user/.nvm/versions/node/v22.15.1/lib/node_modules/Haraka/connection.js:40
for (const ip of cfg.haproxy.hosts) {
                             ^

TypeError: Cannot read properties of undefined (reading 'hosts')
    at Object.<anonymous> (/home/user/.nvm/versions/node/v22.15.1/lib/node_modules/Haraka/connection.js:40:30)
    at Module._compile (node:internal/modules/cjs/loader:1730:14)
    at Object..js (node:internal/modules/cjs/loader:1895:10)
    at Module.load (node:internal/modules/cjs/loader:1465:32)
    at Function._load (node:internal/modules/cjs/loader:1282:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:235:24)
    at Module.require (node:internal/modules/cjs/loader:1487:12)
    at require (node:internal/modules/helpers:135:16)
    at Object.<anonymous> (/home/user/.nvm/versions/node/v22.15.1/lib/node_modules/Haraka/server.js:14:21)

Node.js v22.15.1

### Steps To Reproduce

npm i Haraka -g
haraka -i /tmp/haraka_demo
cd /tmp/haraka_demo

update port in config/smtp.ini file
port=587

create a file config/auth_flat_file.ini with below contents

[core]
methods=PLAIN,LOGIN,CRAM-MD5
constrain_sender=true

[users]
user1=password1
user@domain.com=password2

Start haraka

haraka -c .

### System Info

Please report your OS, Node version, and Haraka version by running this shell script on your Haraka server and replacing this section with the output.

echo "Haraka | $(haraka -v)"; echo " --- | :--- "; echo "Node | $(node -v)"; echo "OS | $(uname -a)"; echo "openssl | $(openssl version)"

Haraka | Haraka.js — Version: 3.1.0
 --- | :--- 
Node | v22.15.1
[...]

PR #3467, opened 2025-05-19
Release v3.1.1

msimerson:
- Fix: install connection.ini with base configuration on install #3458
- Fix crash when outbound connection fails #3456


msimerson:
fixes #3460, #3466

PR #3498, opened 2025-11-12
Make Haraka more configurable and secure

AprilGrimoire:

- Add config option in smtp.ini for ports to enable STARTTLS
- Make authentication required on ports with a whitelist for authentication-exemption; add config option for this exemption

Checklist:

- [y] docs updated
- [n] tests updated (Needs help)
- [y] [Changes](https://github.com/haraka/Haraka/blob/master/Changes.md) updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 - feat(rabbitmq_amqplib): configurable optional exchange arguments #3472
 - feat(rabbitmq_amqplib): configurable message priority #3472
 - add save-sent to Plugins.md
+- Make ports to allow no authentication a whitelist configurable in smtp.ini
+- Make ports to allow no tls a whitelist configurable in smtp.ini
 
 ### [3.1.1] - 2025-05-19
 
@@ -37,7 +39,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 | max_line_length | [max] line_length |
 | max_data_line_length | [max] data_line_length |
 
-AND 
+AND
 
 - moves the following settings from smtp.ini to connection.ini:
   - headers.*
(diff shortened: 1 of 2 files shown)


AprilGrimoire:
It breaks server/nodemailer/accpets SMTP message, but that is by design. Should I add more testing code with a new server instance for this?

msimerson:
> It breaks server/nodemailer/accpets SMTP message, but that is by design.

Presumably because you believe that makes it more secure? How, exactly? Please explain why you believe this.

> Should I add more testing code with a new server instance for this?

Why? The existing testing code is doing its job of preventing well intentioned PRs from breaking existing and useful functionality.

> (btw why are there many certificate errors? Are those related to me?

There errors are spat out because there are broken (in unique ways) TLS certificates in the test configs. Broken TLS certs can make it into production and we prefer to handle the inability to load a TLS certificate gracefully.

config/smtp.ini
```diff
@@ -11,6 +11,12 @@
 ; try to figure it out. If that doesn't work, set it here.
 ;public_ip=N.N.N.N
 
+; Ports not requiring authentication (default: 25)
+;auth_exempt_ports[]=25
```
msimerson: Avoid using negative / inverse names. They tend towards double negatives which are more complicated to reason about. Eg, it's much easier to reason about `if (auth_required) ...` than if (!auth_exempt)...` When choosing names, use positive and imperatives.

AprilGrimoire: `auth_exempt` would be more fool-proof than `auth_required`. If some user opens another port but fails to notice that there is such an option, it would be left in a secured state by default.

AprilGrimoire:
Sorry, I lost interest in Haraka. I don't like JavaScript at the first place, and knowing that Stalwart will be released soon, I lost motivation to finish this pull request.

PR #3502, opened 2025-11-19
configurable outbound IPv4/IPv6 preference using `inet_prefer`

gramakri:
- Add `inet_protocols` to outbound.ini to configure the outbound IP address types.

Checklist:
- [X] docs updated
- [ ] tests updated

Background:

Many servers in the wild are flagging mail sent via IPv6 as "spammy". This is despite setting valid IPv6 PTR records and correct SPF. Primary issue comes from various RBL lists blocking IPv6 addresses wholesale. Spamhaus uses [/64 blocks](https://www.spamhaus.org/blocklists/combined-spam-sources/) for IPv6 (waaay too big) and this results in IPv6 never delivering properly with public cloud VPS IPv6 addresses.

With this option, users can configure Haraka to send with IPv4 only. They can still receive emails via IPv6. A similar option is [also in postfix](https://www.postfix.org/postconf.5.html#smtp_address_preference) and has similar notes on its purpose as above.

Maybe someday we can have other options like `preferIPv4` (which node.js dns.resolve() has) etc. but I don't want to implement them without having a use case. The semantics of such flags are also not straightforward when mixed with MX priority.

### DIFF
--- a/docs/Outbound.md
+++ b/docs/Outbound.md
@@ -53,6 +53,18 @@ Default: false. By default, outbound to a local IP is disabled, to avoid creatin
 
 Set this to specify the delay intervals to use between trying to re-send an email that has a temporary failure condition. The setting is a comma separated list of time spans and multipliers. The time span is a number followed by `s`, `m`, `h`, or `d` to represent seconds, minutes, hours, and days, respectively. The multiplier is an asterisk followed by an integer representing the number of times to repeat the interval. For example, the entry `1m, 5m*2, 1h*3` results in an array of delay times of `[60,300,300,3600,3600,3600]` in seconds. The email will be bounced when the array runs out of intervals (the 7th failure in this case). Set this to `none` to bounce the email on the first temporary failure.
 
+* `inet_prefer`
+
+Default: default. Selects the preferred address family (IP version) to deliver messages.
+
+| Value              | Description |
+|------------------------|-------------|
+| `default` | Prefer IPv6 when IPv4 and IPv6 IPs exist at the same MX priority |
+| `v4`    | Try IPv4 addresses first, then IPv6 |
+| `v6`    | Try IPv6 addresses first, then IPv4 |
+
+Note: Delivery attempts follow MX priority order. Socket-based deliveries ignore this setting.
+
 ### outbound.bounce\_message
 
 See "Bounce Messages" below for details.
--- a/Changes.md
+++ b/Changes.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 - feat(rabbitmq_amqplib): configurable message priority #3472
 - add save-sent to Plugins.md
 - deferred hook is now passed the failed recips list and mx info
+- feat(outbound): configurable outbound IPv4/IPv6 preference using `inet_prefer`
 
 ### [3.1.1] - 2025-05-19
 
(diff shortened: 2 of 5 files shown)


msimerson:
> Haraka will deliver messages using IPv4 and IPv6 (no preference)

I'm curious, have you tested this? I'd expect Haraka to use IPv6 first, when it's available.

gramakri:
@msimerson whoops, the docs I wrote are wrong. It definitely prefers IPv6 first. That is part of the reason I wrote this PR in the first place. Many VPS are auto-assigning IPv6 (good). Our product's email server (which uses Haraka internally) ends up using that IPv6 for outbound and most times ends up in Spam despite good IPv4 reputation.

Initially, I planned to just fix this in our plugin which already handles mx hook. But I thought maybe this config is generally useful.

Also, apologies, I am still testing this change, didn't expect you to see it this early :-)

gramakri:
Recently, what's also happened is that DigitalOcean has started blocking outbound IPv6 port 25 wholesale at the network level. Our mail server is on DO since a decade and basically all IPv6 outbound has stopped working and (currently) there is no way to tell Haraka to just skip IPv6.

gramakri:
> This is good right? In order to remove the block, you need to contact them and have them open it for your MTAs, same as other providers like AWS have been doing on IPv4 for .... ages now.

Yes, that's good. But DO does not allow removing the block on IPv6 (only IPv4). It's at the network level apparently and not VPS/account level. I had talked to their Support earlier today and they just gave me a link to https://www.digitalocean.com/community/questions/outgoing-connections-on-port-25-587-143-blocked-over-ipv6

msimerson:
This PR looks great. Notice that in outbound, when we filter the MX, we emit a message saying so. That's because folks have been surprised by the MX filters.

That leads me to a thought partially expressed in #3405. What happens when someone has IPv4 preferred and doesn't have a v4 address?

I wouldn't expect someone to purposefully set it up that way, but it's easy to see someone having V4 preferred today and then removing the V4 address in the future.

Assuring a v(4|6) address exists before filtering seems like a good safeguard against accidentally breaking outbound. Thoughts?

docs/Outbound.md
```diff
@@ -53,6 +53,16 @@ Default: false. By default, outbound to a local IP is disabled, to avoid creatin
...
 
+* `inet_prefer`
+
+Default: default. Selects the preferred connection protocol to deliver messages. Deliveries over a socket are unaffected by this option.
+
+| Value              | Description |
+|------------------------|-------------|
+| `default` | Use either IPv4 or IPv6. When both address families exist at the same MX priority level, IPv6 is preferred. |
+| `v4`    | Prefer IPv4. Tries all IPv4 addresses first, in MX priority order. IPv6 addresses are attempted thereafter. |
+| `v4`    | Prefer IPv6. Tries all IPv6 addresses first, in MX priority order. IPv4 addresses are attempted thereafter. |
```
msimerson: There's a typo (second v4 in Value column) and a lot of redundancy in this description. Can we wordsmith this down to something more succinct like this?

Default: default. Selects the preferred address family (IP version) to deliver messages.

| Value              | Description |
|------------------------|-------------|
| `default` | Prefer IPv6 when IPv4 and IPv6 IPs exist at the same MX priority |
| `v4`    | Try IPv4 addresses first, then IPv6 |
| `v6`    | Try IPv6 addresses first, then IPv4 |

Note: Delivery attempts follow MX priority order. Socket-based deliveries ignore this setting.

gramakri: Done, good rewrite.

msimerson (approved):

PR #3503, opened 2025-11-20
outbound: skip mxes for which no local address family is bound

### LINKED ISSUE #3395: EAFNOSUPPORT errors when IPv6 is disabled on server
Starting with Haraka 3.0.4, when an IPv4 only server tries to send an email to an MX listening on both IPv4 and IPv6, it gets an error:
`[outbound] Failed to get socket: connect EAFNOSUPPORT 2607:f8b0:4023:1002::1b:25 - Local (undefined:undefined)
`

Also, if this IPv4 only server tries to start up with listen=[::0]:25 in config/smtp.ini then it will get these errors:
```
[server] Failed to setup listeners: bind EAFNOSUPPORT ::0:25
[server] Error: bind EAFNOSUPPORT ::0:25
    at listenOnPrimaryHandle (node:net:1970:18)
    at rr (node:internal/cluster/child:163:12)
    at Worker.<anonymous> (node:internal/cluster/child:113:7)
    at process.onInternalMessage (node:internal/cluster/utils:49:5)
    at process.emit (node:events:531:35)
    at emit (node:internal/child_process:951:14)
    at process.processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -97,
  code: 'EAFNOSUPPORT',
  syscall: 'bind',
  address: '::0',
  port: 25
}

```

I expect that Haraka, when starting, would throw an error if smtp.ini has it listening on an IPv6 address when IPv6 is not supported.
I also expect that Haraka would test if IPv6 is supported when trying to connect to remote MX that is listening on IPv6 and ignore that MX in favor of one listening on IPv4.

This is easy to reproduce. This server has IPv6 disabled and has Haraka listening on a single IPv4 address, e.g. 1.2.3.4:25. 
Postfix is listening on 127.0.0.1:25.

Haraka | Haraka.js — Version: 3.0.4
 --- | :--- 
Node | v20.15.0
OS | Linux daisy.bluestreak.net 6.1.0-20-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.85-1 (2024-04-11) x86_64 GNU/Linux
openssl | OpenSSL 3.0.13 30 Jan 2024 (Library: OpenSSL 3.0.13 30 Jan 2024)

msimerson:
Fixes #3395 

Checklist:

- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 - feat(rabbitmq_amqplib): configurable message priority #3472
 - add save-sent to Plugins.md
 - deferred hook is now passed the failed recips list and mx info
+- outbound: skip mxes for which no local address family is bound #3503
 
 ### [3.1.1] - 2025-05-19
 


gramakri:
Won't `os.networkInterfaces()` always return lo and thus IPv4 (127.0.0.1) and IPv6 (::1) always be there?

I mostly use Ubuntu and atleast on that debian/ubuntu and afaik you cannot compile out ipv4 and ipv6 stack. I would be really suspicious about running software on servers without loopback interfaces.

Also, not 100% sure but I think #3395 is because maybe Haraka tries to listen on IPv6 and not about sending over IPv6.

msimerson:
> Won't os.networkInterfaces() always return lo and thus IPv4 (127.0.0.1) and IPv6 (::1) always be there?

That is typically true on a host, but when the host has IPv6 (or IPv4) disabled, it won't have those default addresses bound. Also, it's not necessarily the case in containers:

```sh
jexec mailtest
ifconfig
em0: flags=1008843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST,LOWER_UP> metric 0 mtu 1500
	ether 00:0c:29:f3:85:6a
	media: Ethernet autoselect (1000baseT <full-duplex>)
	status: active
lo0: flags=1008049<UP,LOOPBACK,RUNNING,MULTICAST,LOWER_UP> metric 0 mtu 16384
	groups: lo
lo1: flags=1008049<UP,LOOPBACK,RUNNING,MULTICAST,LOWER_UP> metric 0 mtu 16384
	inet 172.16.15.19 netmask 0xffffffff
	groups: lo
# node
Welcome to Node.js v22.19.0.
Type ".help" for more information.
> function has_addr_family (version) {
...       if (!['IPv4', 'IPv6'].includes(version)) {
...         this.logerror(`invalid arg to has_addr_family`)
...       }
... 
...       for (const [ifName, ifInfo] of Object.entries(os.networkInterfaces())) {
...         for (const info of ifInfo) {
...           if (info.family === version) return true;
...         }
...       }
... 
...       return false;
...     }
undefined
> has_addr_family('IPv6')
false
> has_addr_family('IPv4')
true
> 
```

msimerson:
> I mostly use Ubuntu and atleast on that debian/ubuntu and afaik you cannot compile out ipv4 and ipv6 stack.

Ubuntu, you say?

```
$ uname -a
Linux ubuntu-vmware 6.8.0-44-generic #44-Ubuntu SMP PREEMPT_DYNAMIC Tue Aug 13 14:20:20 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux
$ cat /etc/debian_version 
trixie/sid

$ tail -n4 /etc/sysctl.conf 
net.ipv6.conf.all.disable_ipv6=1
net.ipv6.conf.default.disable_ipv6=1
net.ipv6.conf.lo.disable_ipv6=1

$ ifconfig lo
lo: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
        inet 127.0.0.1  netmask 255.0.0.0
        loop  txqueuelen 1000  (Local Loopback)

$ node
Welcome to Node.js v18.19.1.
Type ".help" for more information.
> function has_addr_family (version) {
...       if (!['IPv4', 'IPv6'].includes(version)) {
...         this.logerror(`invalid arg to has_addr_family`)
...       }
... 
...       for (const [ifName, ifInfo] of Object.entries(os.networkInterfaces())) {
...         for (const info of ifInfo) {
...           if (info.family === version) return true;
...         }
...       }
... 
...       return false;
...     }
undefined
> has_addr_family('IPv6')
false
> has_addr_family('IPv4')
true

```

gramakri:
@msimerson (not an expert) If the intent here is to check if we can send out IPv6, then we should just let OS return an error. After all, we don't need to keep calling os.networkInterfaces(), not sure how expensive this is or what node does. Haraka will anyway log and continue to next IP when connect fails. If the intent is to have better logs for this specific situation, maybe we should catch that connect error like you seem to do in https://github.com/haraka/Haraka/blob/master/server.js#L421 ?

Also, AFAIK The EAFNOSUPPORT reported in https://github.com/haraka/Haraka/issues/3395 is only returned when IPv6 is compiled out or the module is not loaded and then disabled in kernel altogether. Disabling using sysctl flags won't return that error since it's only disabling intf configuration. The kernel will still support ::0 . Admittedly, I am guessing here. This is a setup which I have never dealt with myself since practically all cloud VPS has IPv6 compiled in.

PR #3560, opened 2026-05-14
fix(outbound): fix reading from null this.todo in temp fail bounce

DoobleD:
Fixes this error I encountered in Haraka@3.1.5:

```
[CRIT] [-] [core] TypeError: Cannot read properties of null (reading 'rcpt_to')
[CRIT] [-] [core]     at HMailItem.convert_temp_failed_to_bounce (/path/to/Haraka/outbound/hmail.js:1324:19)
[CRIT] [-] [core]     at HMailItem.temp_fail (/path/to/Haraka/outbound/hmail.js:1339:25)
[CRIT] [-] [core]     at HMailItem.size_file (/path/to/Haraka/outbound/hmail.js:92:18)
[CRIT] [-] [core] TypeError: Cannot read properties of null (reading 'rcpt_to')
[CRIT] [-] [core]     at HMailItem.convert_temp_failed_to_bounce (/path/to/Haraka/outbound/hmail.js:1324:19)
[CRIT] [-] [core]     at HMailItem.temp_fail (/path/to/Haraka/outbound/hmail.js:1339:25)
[CRIT] [-] [core]     at HMailItem.size_file (/path/to/Haraka/outbound/hmail.js:92:18)
[CRIT] [-] [core] TypeError: Cannot read properties of null (reading 'rcpt_to')
[CRIT] [-] [core]     at HMailItem.convert_temp_failed_to_bounce (/path/to/Haraka/outbound/hmail.js:1324:19)
[CRIT] [-] [core]     at HMailItem.temp_fail (/path/to/Haraka/outbound/hmail.js:1339:25)
[CRIT] [-] [core]     at HMailItem.size_file (/path/to/Haraka/outbound/hmail.js:92:18)
```

I didn't spend much time trying to understand the root cause more deeply, but Claude says:

image

Checklist:

- [ ] docs updated
- [ ] tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Unreleased
 
+- fix(outbound): fix reading from null this.todo in temp fail bounce
 - fix(outbound): guard against error emit after listeners removed #3554
 - fix(outbound): yield before delivery attempts #3552
 - test(outbound,conn,endpoint,server,tls_socket): added tests #3552
--- a/outbound/hmail.js
+++ b/outbound/hmail.js
@@ -1321,12 +1321,14 @@ class HMailItem extends events.EventEmitter {
     }
 
     convert_temp_failed_to_bounce(err, extra) {
-        this.todo.rcpt_to.forEach((rcpt_to) => {
-            rcpt_to.dsn_action = 'failed'
-            if (rcpt_to.dsn_status) {
-                rcpt_to.dsn_status = `${rcpt_to.dsn_status}`.replace(/^4/, '5')
-            }
-        })
+        if (this.todo) {
+            this.todo.rcpt_to.forEach((rcpt_to) => {
+                rcpt_to.dsn_action = 'failed'
+                if (rcpt_to.dsn_status) {
+                    rcpt_to.dsn_status = `${rcpt_to.dsn_status}`.replace(/^4/, '5')
+                }
+            })
+        }
         return this.bounce(err, extra)
     }
 


msimerson:
Hmmm, this isn't right, it trades a crash for a silent queue slot leak. Looking into it... While the `if (this.todo)` guard would have prevented the crash, it introduced a silent deadlock — bounce() would wait for the 'ready' event that can never fire since read_todo() was never called.

msimerson:
Just for sport, I'm asking Copilot to weigh in.

msimerson:
Take a look at #3561 

AI can be helpful, but it's also prone to hallucinating and gaslighting. Caveat emptor.

DoobleD:
> Take a look at #3561
> 
> AI can be helpful, but it's also prone to hallucinating and gaslighting. Caveat emptor.

Thanks for the responsiveness @msimerson! #3561 looks good to me, with my limited understanding of the hmail code though.

Do you think the 2 other occurrences mentioned by Copilot are hallucinations, or are they addressed by the same fix?

DoobleD:
Closing in favor of #3561.

msimerson:
> Do you think the 2 other occurrences mentioned by Copilot are hallucinations, or are they addressed by the same fix?

They're "similar" code which is what Copilot was flagging on. But neither one was getting at the root issue, which is that you've got an unreadable queue file entry. An admin could have deleted it, another process could have renamed it, it could have been delivered, etc. Regardless, when we can't read the TODO, the only appropriate thing to do stop trying.

DoobleD:
> An admin could have deleted it, another process could have renamed it, it could have been delivered, etc.

Yes, in my case it was probably some kind of NFS sync or access issue, the queue directory being on an NFS server.

PR #3565, opened 2026-05-15
deps(all): switch from ^ to ~ version ranges

msimerson:
Haraka has historically favored using ^ versions for our dependencies. Caret (^) ranges permit new installs to get any *major* version greater than the version listed in package.json. The primary argument for using caret or tilde (~) ranges is **user** safety. We want Haraka installs to automatically pick up security and bug-fix updates.

Using caret ranges also means new installs sometimes pick up changes they aren't ready for. This is mostly because Haraka's tradition is to be stingy with major version bumps. We prioritize backwards compatibility and avoiding breaking changes whenever possible.

I think switching to tilde ranges is a better choice that still allows new installs to pick up semver patch (eg 1.0.3 -> 1.0.4) updates, while narrowing down the potential behavior changes a caret version might bring in.

### Open Questions

- should Haraka pin versions? (Maximum predictability, slower security/bugfix uptake)


DoobleD:
> * should Haraka pin versions? (Maximum predictability, slower security/bugfix uptake)

The tilde versions is a good user safety improvement and I'd say is probably good enough. Why not revisit that with pinned versions later on if it turns out tilde wasn't enough.

PR #3577, opened 2026-05-26
Implicit TLS with Proxy Protocol

### LINKED ISSUE #3105: Nginx proxy protocol to Haraka on port 465 w/ proxy protocol enabled = SSL handshake fail
Is proxy protocol supposed to work also on port 465 (implicit SSL/TLS)?

I've set up an Nginx stream proxy in front of Haraka, that works fine. But when I enable the proxy protocol (`proxy_protocol` directive on Nginx and `haproxy_hosts` config file on Haraka), the connection breaks before Haraka logs even a single thing, with the following error from Nginx: 

`peer closed connection in SSL handshake while SSL handshaking to upstream
`

I tried the same thing without SSL (targeting port 587 STARTTLS instead), and it worked fine. Hence I'm wondering if there could be some issue when using proxy protocol with SSL/TLS?

Interestingly Dovecot works fine with the same setup (proxy protocol to implicit SSL/TLS IMAP and POP ports).

cdgco:
Previously, SMTPS used `tls.createServer`, so OpenSSL expected the first bytes on the connection to be a TLS ClientHello. When proxies sent a plaintext `PROXY` header first, OpenSSL interpreted that header as TLS and failed with errors like `wrong version number` or `packet length too long`, as mentioned in #3105.

This PR adds support for SMTPS behind proxies by pre-reading the `PROXY` line before starting the TLS handshake. That lets Haraka consume and validate the cleartext `PROXY` header before handing the remaining bytes to Node’s TLS layer.

Proxy parsing logic in `connection.js` was refactored through some reusable functions extracted from `Connection.cmd_proxy` so `Server.create_smtps_server` can parse and validate the `PROXY` line before starting the TLS handshake.

SMTPS now starts with a raw `net.createServer` for the initial TCP connection. If the peer is trusted and the first bytes begin with `PROXY` then the socket is wrapped in TLS with `start_tls_with_buffer`. If the connection started with a TLS handshake rather than `PROXY` or the IP is not a trusted peer, the socket is wrapped with TLS via `start_tls`, or `start_tls_with_buffer` (to unshift any bytes we inspected before starting TLS).

Fixes #3105

Verified:
- `node --test test/server.js test/connection.js`
- Manually tested SMTPS behind Traefik / Nginx / HAProxy with PROXY Protocol v1

Checklist:
- [ ] docs updated
- [x] tests updated
- [ ] [CHANGELOG](https://github.com/haraka/Haraka/blob/master/CHANGELOG.md) updated

### DIFF
--- a/config/connection.ini
+++ b/config/connection.ini
@@ -19,6 +19,9 @@
 
 
 [haproxy]
+; Bool: enable HAProxy PROXY protocol support and SMTPS pre-parsing.
+; enabled=true
+
 ; Array: hosts or CIDRs that Haraka should enable the PROXY protocol from. See docs/HAProxy for format
 hosts[] =
 ; hosts[] = 192.0.2.4
--- a/connection.js
+++ b/connection.js
@@ -31,31 +31,38 @@ const cfg = config.get('connection.ini', {
         '+headers.add_received',
         '+headers.show_version',
         '+headers.clean_auth_results',
+        '+haproxy.enabled',
     ],
 })
 
+const haproxy_enabled = cfg.haproxy.enabled !== false
 const haproxy_hosts_ipv4 = []
 const haproxy_hosts_ipv6 = []
 
-for (const ip of cfg.haproxy.hosts) {
-    if (!ip) continue
-    if (net.isIPv6(ip.split('/')[0])) {
-        haproxy_hosts_ipv6.push([ipaddr.IPv6.parse(ip.split('/')[0]), parseInt(ip.split('/')[1] || 64)])
-    } else {
-        haproxy_hosts_ipv4.push([ipaddr.IPv4.parse(ip.split('/')[0]), parseInt(ip.split('/')[1] || 32)])
+if (haproxy_enabled) {
+    for (const ip of cfg.haproxy.hosts) {
+        if (!ip) continue
+        if (net.isIPv6(ip.split('/')[0])) {
+            haproxy_hosts_ipv6.push([ipaddr.IPv6.parse(ip.split('/')[0]), parseInt(ip.split('/')[1] || 64)])
+        } else {
+            haproxy_hosts_ipv4.push([ipaddr.IPv4.parse(ip.split('/')[0]), parseInt(ip.split('/')[1] || 32)])
+        }
     }
 }
 
+// TODO: Move these three functions to net_utils
 function normalize_ip(ip) {
     if (!net.isIP(ip)) return null
     return ipaddr.process(ip).toString()
 }
 
-function is_haproxy_allowed(ip) {
+function is_haproxy_allowed(ip, enabled, ipv6_hosts, ipv4_hosts) {
+    if (!enabled) return false
+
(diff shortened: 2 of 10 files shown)


msimerson:
random question: do you wish Haraka had a ./lib/ folder?

If yes, wouldn't that make a great spot for a `./lib/proxy.js` to exist with all the proxy logic encapsulated and only minimal additions being made in server.js?  I sure do... In completely unrelated news, see lsongdev/node-dns#122 for what I'm suggesting.

Besides proxy 2 support, the other thing that's sorely missing from Haraka's proxy implementation is a `[haproxy] enabled = true | false` gate so the proxy parsing is only hit when it's configured.

cdgco:
Yeah, it felt like some of this didn't have a clear place to live or there wasn't a clear pattern for helpers besides going in something like `haraka-net-utils`.

msimerson:
> besides going in something like haraka-net-utils.

Yep, net-utils is exactly the right spot for those functions.

1. Possible regression(s) on resumeSession and OCSP. The original code does, on a real tls.Server:  `tls_socket.addOCSP(server); server.on('resumeSession', …)` In this PR, server is now a net.Server. Per-connection tls.TLSSockets don't bubble resumeSession/OCSPRequest up to the parent (and `opts.server = <netServer>` doesn't reattach them either), so TLS session resumption and OCSP stapling on the SMTPS port are probably silently broken and not covered by the new tests. Have you verified with a session-resuming client? (I'm less worried about OCSP since Let's Encrypt stopped supporting stapling.)
2. I really want to see this new code path gated behind a `[haproxy]enabled=true` (default true) setting. The TLS resumption interaction with net.Server-hosted TLSSockets is the kind of thing that may not surface until production traffic hits it. Being able to turn off this code path is a safety switch.
3. Refactoring is_haproxy_allowed / normalize_ip / parse_proxy_line in connection.js is a win 🙏🏻, but they belong in haraka-net-utils. Then the tests don't need to monkey-patch the export to stub allowlist behavior.
4. Two ad-hoc flags on the socket (haraka_proxy, haraka_proxy_checked) drive three branches in setupClient. Consider collapsing them to a single object (socket.haraka_smtps = { proxy, peer_allowed }) with a short comment explaining the contract; right now the flag semantics aren't obvious.
5. 512 appears twice as a magic number
6. the cleanup/unshift/resume dance is unannotated.
7. Two new integration tests cover the happy paths (PROXY-then-TLS, direct-TLS from allowed peer). No tests for: untrusted peer (the start_tls(socket) branch), oversized PROXY line, malformed PROXY, proxy timeout, byte-by-byte arrival (the prefix-mismatch fallback).
8. That windows test failure is curious.

cdgco:
Session resumption was not working with the net server, I changed that to start with a net server then hand the socket to a tls server so tls can manage lifecycle again. I verified that it is resuming properly now.

msimerson:
At first glance, this looks much nicer. I anticipate giving it a closer look later today.

msimerson (approved):

Issue #3564, opened 2026-05-15
Announce: @haraka/email-address to replace address-rfc282{1,2}


msimerson:
Haraka is deprecating [address-rfc2821](https://github.com/haraka/node-address-rfc2821) and [address-rfc2822](https://github.com/haraka/node-address-rfc2822) in favor of [@haraka/email-address](https://github.com/haraka/email-address).

### Why?

- security: haraka/node-address-rfc2821#58, haraka/node-address-rfc2821#59, and unpublished issues raised by internal audits
- [Performance](https://github.com/haraka/email-address/blob/master/PERFORMANCE.md): it’s ~ 43x faster than address-rfc2821 and ~ 19x faster than address-rfc2822 
- @haraka/email-address is a modern, single-package replacement with zero-dependencies.
- ESM + CJS support
- A more consistent, explicit, and simpler API

### Compatibility

The breaking surface is:

| Area | address-rfc2821 | address-rfc2822 | @haraka/<br>email-address | Example migration |
 |---|---|---|---|---|
 | Envelope address | `.address()` | `.address` | `.address` | `txn.rcpt_to[0].address()` → `txn.rcpt_to[0].address` |
 | Envelope host | `.host` | `.host()` | `.host` | `parse(sender)[0].host()` → `parseHeader(sender)[0].host` |
 | Header parsing | n/a | `parse(sender)` | `parseHeader(sender)` | `parse(sender)` → `parseHeader(sender)` |
 | Haraka runtime | n/a | n/a | `asLegacy()` wrapper applied | Both `.address()` and `.address` work on conn/txn objects |

The compatibility shims ~will be removed in 2027~ are gone as of 3.3.0.

### Required Actions

Plugins need to be updated to use the property methods before 2027. This will be completed for all Haraka bundled plugins in May 2026. Third party plugins that use those 3 interfaces must be updated.

- fixes haraka/node-address-rfc2821#58
- fixes haraka/node-address-rfc2821#57

msimerson:
With the benefit of hindsight, what I should have done in every one of the changes linked in this issue, is instead of normalizing `.host()` and `.address()` calls to `.host` and `.address`, I should have replaced them with `String(host | address)` to coerce the results to strings. Then, after the sunset period for the legacy wrapper, the String could be removed. #missedOpportunity

PR #3581, opened 2026-06-02
Add logging to address validation errors

TaaviE:
This PR adds two log lines for address validation errors so that they can be properly debugged.

I would also propose to expose this error (or a subset of errors) to the SMTP client in a subsequent PR, what do you think?

Checklist:

- [X] docs updated (N/A)
- [X] tests updated (N/A)
- [ ] [CHANGELOG](https://github.com/haraka/Haraka/blob/master/CHANGELOG.md) updated


msimerson:
Thoughts: 

- lognotice? spam bots will make this noisy in production. Logdebug is a better default.
- JSON.stringify escapes control chars, so this doesn't open the door to log injection attacks
- you'll get an odd result when from_raw is undefined.
- relaying attacker controlled data is a classic open door for injection attacks.

> This PR adds two log lines for address validation errors so that they can be properly debugged.

If you're going to watch the log files for this, that's fine but changing the response to clients is not welcome. It can break response parsing clients, it imposes the need to sanitize / escape the replies, and the remote already KNOWS what they sent. I fail to see how relaying it to them adds value.

TaaviE:
> relaying attacker controlled data is a classic open door for injection attacks.

Plenty of other log lines log the MAIL FROM and RCPT TO addresses.

>  and the remote already KNOWS what they sent. I fail to see who relaying it to them adds value.

They do not know which validation failed.

> lognotice? spam bots will make this noisy in production. Logdebug is a better default.

Considering how this stops a delivery for any number of reasons `logdebug` does not seem like a better default.

msimerson:
> Plenty of other log lines log the MAIL FROM and RCPT TO addresses.

The difference is those are logging data that has parsed successfully. In theory, once we've parsed that data, it is safe for later uses.

> They do not know which validation failed.

Sure they do, because we tell them: `this.respond(501, 'Invalid MAIL FROM address')`

TaaviE:
> Sure they do, because we tell them: this.respond(501, 'Invalid MAIL FROM address')

There's a big difference between an excessively long local part and something that uses invalid codepoints for example. The fact is that the sender doesn't know which part of the specification they've supposedly violated. They don't even know if it's a recipient's local policy or not unless they check Haraka's code.

msimerson:
> There's a big difference between an excessively long local part and something that uses invalid codepoints for example. The fact is that the sender doesn't know which part of the specification they've supposedly violated. They don't even know if it's a recipient's local policy or not unless they check Haraka's code.

Fair points.

For casual readers, `cmd_mail` and `cmd_rcpt` address-validation currently swallow the error. This PR adds server-side observability at the lognotice level. Until my recent upgrade to _@haraka/email-address_  (in #3564), this PR wouldn't have been useful at all. (See below for TL;DR)

> I would also propose to expose this error (or a subset of errors) to the SMTP client in a subsequent PR, what do you think?

I think you should amend this PR and do it now, eg:

```js
    const msg = `Invalid MAIL FROM address ${JSON.stringify(from_raw)}: ${err.message}`
    this.lognotice(msg)
    return this.respond(501, msg)
```

### TL;DR

Today an operator debugging "why did this sender get a 501?" has nothing in the logs to go on. The PR gives them the offending token and the precise parse failure. All messages below are the real output captured by replaying the inputs through the `Address` constructor.

| `MAIL FROM:` input    | TODAY          | PR #3581                                                                 |
| --------------------- | ------------------------------- | ------------------------------------------------------------------------------------------ |
| `<foo>`               |  | `Invalid MAIL FROM address "<foo>": Invalid RFC-5321 address at position 4: expected "@"` |
| `<user@>`             |  | `Invalid MAIL FROM address "<user@>": Invalid RFC-5321 address at position 6: expected sub-domain`                          |
| `<user@@example.com>` |  | `Invalid MAIL FROM address "<user@@example.com>": Invalid RFC-5321 address at position 6: expected sub-domain`              |

Boundary note: inputs like `<user example.com>` (space, no `@`) fail earlier in `rfc1869.parse`, returning `501 Command parsing failed` — that path already logs (`err.stack` first line). See also haraka/haraka-utils#37, which defers that behavior.

### Why `err.message` is log-friendly: new vs previous parser

The PR's `err.message` is a clean single line **only because the parser was already migrated** to `@haraka/email-address`. The same set of failing inputs through the previous parser (`address-rfc2821`, a nearley grammar) produces huge multi-line error dumps.

| Input                 | `@haraka/email-address` (current, what the PR logs)        | `address-rfc2821` (previous parser)                       |
| --------------------- | ---------------------------------------------------------- | --------------------------------------------------------- |
| `<foo>`               | `Invalid RFC-5321 address at position 4: expected "@"`     | `Syntax error at line 1 col 5: … Unexpected ">"` + ~20 lines of grammar productions |
| `<@example.com>`      | `… at position 13: expected ":" after source route`        | Syntax error + ~45 lines of nearley state dump            |
| `<user@>`             | `… at position 6: expected sub-domain`                     | Syntax error + ~18 lines                                  |
| `<user@@example.com>` | `… at position 6: expected sub-domain`                     | Syntax error + ~18 lines                                  |
| `<user example.com>`  | `… at position 5: expected "@"`                            | Syntax error + ~22 lines                                  |
| `<user@exa mple.com>` | `… at position 9: expected ">"`                            | Syntax error + ~40 lines                                  |
| `<user@.com>`         | `… at position 6: expected sub-domain`                     | Syntax error + ~18 lines                                  |
| `<user@-bad.com>`     | `… at position 6: expected sub-domain`                     | Syntax error + ~18 lines                                  |

A representative full legacy message, for `<foo>`:

```
Invalid RFC-5321 address: Syntax error at line 1 col 5:

1 <foo>
      ^

Unexpected ">". Instead, I was expecting to see one of the following:

A character matching /[0-9A-Za-z!#$%&'*+\-/=?^_`{|}~-￿/]/ based on:
    Atom$ebnf$1 → Atom$ebnf$1 ● /[0-9A-Za-z...]/
    Atom →  ● Atom$ebnf$1
    Dot_string →  ● Atom Dot_string$ebnf$1
    ... (many more production lines) ...
```

Had the same `this.lognotice(... ${err.message})` been written against the old parser, every rejected address would have dumped a 20–45 line nearley grammar trace into the log (with embedded blank lines and `^` caret art) — unusable operationally, and would bury the actual log.

TaaviE:
> I think you should amend this PR and do it now, eg: [...]

I've applied the suggestion.

The comment that was left (and deleted) brings up a fair point. The reason behind improving logging is somewhat related to it. Some configurability around local part validation would be nice. Though for starters, logging and more detailed responses will help operators.

msimerson:
> The reason behind improving logging is somewhat related to it.

I agree. That's why we can use special syntax to link issues, like: related to haraka/email-address#8

msimerson (approved):

msimerson:
copilot is rightly concerned with multi-line, and JSON.stringify doesn't help us there. Copilot lacks awareness that the parser is OSS with, "exposes internal parsing details to the client."

Instead of using JSON.stringify, use `utils.santize()` and bump the dep haraka-utils to version `~2.2.0` Then I'll land this PR.

msimerson:
and if you don't mind, also apply this diff:

```diff
diff --git a/test/connection.js b/test/connection.js
index 2c090e9e..bbc281dd 100644
--- a/test/connection.js
+++ b/test/connection.js
@@ -654,6 +654,44 @@ describe('connection', () => {
             assert.equal(responses[2][0], 503)
         })
 
+        it('cmd_mail strips control chars from invalid address logs', () => {
+            const notices = []
+            const responses = []
+            this.connection.hello.host = 'example.test'
+            this.connection.lognotice = (msg) => notices.push(msg)
+            this.connection.respond = (code, msg) => {
+                responses.push({ code, msg })
+            }
+
+            this.connection.cmd_mail('FROM:<mail\x00@example.com>')
+
+            assert.equal(responses[0].code, 501)
+            assert.match(responses[0].msg, /^Invalid MAIL FROM address /)
+            assert.equal(responses[0].msg.includes('\r'), false)
+            assert.equal(responses[0].msg.includes('\n'), false)
+            assert.equal(responses[0].msg.includes('\\u0000'), false)
+            assert.equal(notices[0], responses[0].msg)
+        })
+
+        it('cmd_rcpt strips control chars from invalid address logs', () => {
+            const notices = []
+            const responses = []
+            this.connection.transaction = { mail_from: new Address('<from@example.com>'), rcpt_to: [] }
+            this.connection.lognotice = (msg) => notices.push(msg)
+            this.connection.respond = (code, msg) => {
+                responses.push({ code, msg })
+            }
+
+            this.connection.cmd_rcpt('TO:<rcpt\x00@example.com>')
+
+            assert.equal(responses[0].code, 501)
+            assert.match(responses[0].msg, /^Invalid RCPT TO address /)
+            assert.equal(responses[0].msg.includes('\r'), false)
+            assert.equal(responses[0].msg.includes('\n'), false)
+            assert.equal(responses[0].msg.includes('\\u0000'), false)
+            assert.equal(notices[0], responses[0].msg)
+        })
+
```

